PR #86, opened 2019-10-11
Add Grpc client for dotnet and example.

brendandburns:
cc @yaron2 @amanbha


loic-sharma:
I'm a little late to the game, but you can find the requirements for publishing Microsoft packages to nuget.org here: https://devdiv.visualstudio.com/DevDiv/_wiki/wikis/DevDiv.wiki/972/NuGet?anchor=publishing-microsoft-packages-on-nuget.org

> @davidfowl Yes, we are having conversation about whether to sign or not. We have reserved Dapr.*.

All packages published to NuGet.org should be author signed to ensure that our artifacts are malware scanned, audited, and protected against tampering.

Issue #112, opened 2019-10-18
Unable to reference Dapr.Client or Dapr.AspNetCore due to use of netstandard3.0


philliphoff:
## Expected Behavior

The reference is added, the package is restored, and the library can be built.

## Actual Behavior

The reference is not added, failing with the error:

```
error NU1202: Package Dapr.Client 0.1.0-preview01 is not compatible with netcoreapp3.0 (.NETCoreApp,Version=v3.0). Package Dapr.Client 0.1.0-preview01 supports: netstandard3.0 (.NETStandard,Version=v3.0)
```

The issue appears to be due to the fact that `Dapr.Client` assemblies are added to the NuGet package in the `netstandard3.0` folder instead of `netcoreapp3.0` folder (despite being built as `netcoreapp3.0` assemblies).

## Steps to Reproduce the Problem

  1. Create a `netcoreapp3.0` library (e.g. `dotnet new console`).
  1. `dotnet add package Dapr.Client --version 0.1.0-preview01`

youngbupark:
cc/ @amanbha @shalabhms

amanbha:
@philliphoff  Thansk for reporting the issue.
I have uploaded new packages https://www.nuget.org/packages/Dapr.AspNetCore/0.1.0-preview02
These are being indexed now and should be available soon

davidfowl:
@amanbha Why are we using exact version ranges? We shouldn't be. I'll open an issue for that.

amanbha:
> @amanbha Why are we using exact version ranges? We shouldn't be. I'll open an issue for that.

Sure, makes sense.

philliphoff:
Thanks for the quick turnaround!

PR #113, opened 2019-10-18
Fixing nuspec for nuget generation.

amanbha:
# Description

Fixing nuspec for nuget generation.

Fixes https://github.com/dapr/dotnet-sdk/issues/112

### DIFF
--- a/src/Dapr.Client.Grpc/nuspec/Dapr.Client.Grpc.nuspec
+++ b/src/Dapr.Client.Grpc/nuspec/Dapr.Client.Grpc.nuspec
@@ -21,7 +21,7 @@
   </metadata>
   <files>
     <file src="..\..\..\properties\logo-transparent.png" target="images\" />
-    <file src="..\..\..\bin\$Configuration$\Dapr.Client.Grpc.dll" target="lib\netstandard3.0" exclude="" />
-    <file src="..\..\..\bin\$Configuration$\Dapr.Client.Grpc.xml" target="lib\netstandard3.0" exclude="" />
+    <file src="..\..\..\bin\$Configuration$\Dapr.Client.Grpc.dll" target="lib\netcoreapp3.0" exclude="" />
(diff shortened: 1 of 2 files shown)


loic-sharma:
Out of curiosity @amanbha, why is Dapr using nuspec to generate .nupkg files? Is there something preventing you from packing  packages using the .csproj instead? For more information, please refer to:

* https://docs.microsoft.com/en-us/nuget/create-packages/creating-a-package-dotnet-cli
* https://docs.microsoft.com/en-us/dotnet/core/tools/csproj#nuget-metadata-properties

amanbha:
@loic-sharma  Nuspec allows more control and flexibility on how and where you want to put files, (we are putting icon files in the pkg itself) and provides more control specially if you decide to put multiple dlls form different projects into one nuget pkg.

loic-sharma:
> @loic-sharma Nuspec allows more control and flexibility on how and where you want to put files, (we are putting icon files in the pkg itself) 

Packing from a .nuspec is more flexible, but it is also significantly harder. NuGet has conventions for things like asset compatibility, you will continuously run into issues if you choose to manually pack with a .nuspec. Also, it is possible to [pack arbitrary files](https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#including-content-in-a-package) and or to [embed icons](https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#packing-an-icon-image-file) using .csproj.

> and provides more control specially if you decide to put multiple dlls form different projects into one nuget pkg.

Out of curiousity, why do you need this? I believe you can do this too by [including the assets from your project references](https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#project-to-project-references). This seems like a weird scenario though...

amanbha:
@loic-sharma yes, simplicity is there with .csproj, we picked flexible over simple. We will definatley change it if we run into further issues.

PR #121, opened 2019-10-21
Clean ActorId

shaggygi:
Random code cleaning for ActorId.cs.

### DIFF
--- a/src/Dapr.Actors/ActorId.cs
+++ b/src/Dapr.Actors/ActorId.cs
@@ -30,34 +30,34 @@ namespace Dapr.Actors
         /// <summary>
         /// Determines whether two specified actorIds have the same id.
         /// </summary>
-        /// <param name="x">The first actorId to compare, or null. </param>
-        /// <param name="y">The second actorId to compare, or null. </param>
+        /// <param name="id1">The first actorId to compare, or null. </param>
+        /// <param name="id2">The second actorId to compare, or null. </param>
         /// <returns>true if the id is same for both objects; otherwise, false.</returns>
-        public static bool operator ==(ActorId x, ActorId y)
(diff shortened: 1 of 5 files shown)


src/Dapr.Actors/ActorId.cs
```diff
@@ -31,39 +30,38 @@ public ActorId(string id)
...
         /// </summary>
-        /// <param name="x">The first actorId to compare, or null. </param>
-        /// <param name="y">The second actorId to compare, or null. </param>
+        /// <param name="id1">The first actorId to compare, or null.</param>
+        /// <param name="id2">The second actorId to compare, or null.</param>
         /// <returns>true if the id is same for both objects; otherwise, false.</returns>
-        public static bool operator ==(ActorId x, ActorId y)
+        public static bool operator ==(ActorId id1, ActorId id2)
         {
-            if (ReferenceEquals(x, null) && ReferenceEquals(y, null))
```
amanbha: Could you please add unit tests for the changes being made for all these ActorId comparisions.

shaggygi: Sure.  Reviewing now.

shaggygi: @amanbha You were correct (referencing #123) and appears you could not have both adapters in project at once.  I only removed MSTest and added xUnit packages for Dapr.Actors.Test project.  The other projects appear to show up in Test Explorer and all test pass.

Since it might take a little longer to complete for other projects, you might go ahead and accept this PR (if all looks good 😄).  I can do another PR to complete the other test projects with xUnit referencing issue #123.

Thoughts?

amanbha: Thanks for the changes, that should be fine, but please do that as having 2 different test framework s for long would be weird.

Issue #134, opened 2019-10-24
Remove usage of string.Substring and use Span API instead


gmanvel:
[JsonReaderExtensions.ReadValueAsTimeSpanDaprFormat](https://github.com/dapr/dotnet-sdk/blob/df58a98be9cfb5c7ea3b633be3ac2e28494c93cb/src/Dapr.Actors/JsonReaderExtensions.cs#L387) uses substrings to parse Dapr format TimeSpan string.
Change the method to make use of Span API.

PR #135, opened 2019-10-24
Move to using span slicing instead of string substrings

gmanvel:
# Description

Replaced string.Substring usage with Span API.

## Issue reference

https://github.com/dapr/dotnet-sdk/issues/134

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/src/Dapr.Actors/JsonReaderExtensions.cs
+++ b/src/Dapr.Actors/JsonReaderExtensions.cs
@@ -394,25 +394,32 @@ namespace Dapr.Actors
             {
                 case JsonToken.String:
                     var valueString = (string)reader.Value;
+                    var spanOfValue = valueString.AsSpan();
                     try
                     {
                         // Change the value returned by Dapr runtime, so that it can be parsed with TimeSpan.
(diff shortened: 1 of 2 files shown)


src/Dapr.Actors/JsonReaderExtensions.cs
```diff
@@ -394,25 +394,32 @@ public static byte ReadValueAsByte(this JsonReader reader)
             {
                 case JsonToken.String:
                     var valueString = (string)reader.Value;
+                    var spanOfValue = valueString.AsSpan();
```
amanbha: @gmanvel Thanks for making the changes, could you please add unit test around this change as well.

amanbha: Yes, the internals are exposed to Test project 
https://github.com/dapr/dotnet-sdk/blob/master/src/Dapr.Actors/properties/AssemblyInfo.cs#L6

gmanvel: @amanbha added a unit test

Issue #144, opened 2019-11-03
Async main in console


JamesNK:
Is there a reason why the Main method here is not async?

https://github.com/dapr/dotnet-sdk/blob/07fa2c615dd4c5c9df35947f7fc7262d8205ae94/samples/Grpc/GrpcClient/Program.cs

Making this method async will allow you to replace `.GetAwaiter().GetResult();` with `await`.

Details: https://blogs.msdn.microsoft.com/mazhou/2017/05/30/c-7-series-part-2-async-main/

amanbha:
No specific reason, yes, It can be made async.

PR #145, opened 2019-11-05
Making Main async and using Grpc.Net.Client

### LINKED ISSUE #143: Use Grpc.Net.Client
This sample - https://github.com/dapr/dotnet-sdk/blob/07fa2c615dd4c5c9df35947f7fc7262d8205ae94/samples/Grpc/GrpcClient/Program.cs#L33 - is using Grpc.Core as the client.

Consider removing `Grpc` and `Grpc.Core` package references and use `Grpc.Net.Client` as the client. `Grpc.Net.Client`, unlike `Grpc.Core`, doesn't have any native dependencies.

To call unsecured services, set an AppContext flag once when the app starts up - https://docs.microsoft.com/en-us/aspnet/core/grpc/troubleshoot?view=aspnetcore-3.0#call-insecure-grpc-services-with-net-core-client (note: this will be improved in .NET 5.0)

amanbha:
fixes #143 & #144

### DIFF
--- a/samples/Actor/ActorClient/Program.cs
+++ b/samples/Actor/ActorClient/Program.cs
@@ -20,16 +20,8 @@ namespace ActorClient
         /// Entry point.
         /// </summary>
         /// <param name="args">Arguments.</param>
-        public static void Main(string[] args)
-        {
-            MakeActorCalls().GetAwaiter().GetResult();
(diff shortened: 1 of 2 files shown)


amanbha:
cc @JamesNK

PR #165, opened 2019-11-20
Allow explicit actor type

### LINKED ISSUE #164: Allow actor type be specified during registration
Currently actors are registered via: 

```csharp
Dapr.Actors.Runtime.ActorRuntime.RegisterActor<TActor>(Func<ActorTypeInformation, ActorService> actorServiceFactory = null)
```
The "type" of the registered actor is inferred from the actor implementation class name.  However, clients of the actor must specify the same type string when invoking actor methods.  This makes the implementation class name an implicit component of the actor contract, subject to the risk of breaking the contract due to an innocent refactoring of the implementation.

Proposal:

Add an overload that allows passing an explicit actor type string, which can then be a shared contract between implementation and client (alongside the actor interface definition):

```csharp
Dapr.Actors.Runtime.ActorRuntime.RegisterActor<TActor>(string actorType, Func<ActorTypeInformation, ActorService> actorServiceFactory = null)
```

philliphoff:
# Description

Allows developers to make the actor type name of an actor implementation explicit (i.e. not derived from the actor implementation class name).

Changes include:

 - Adds an optional `ActorAttribute` attribute that can be attached to actor implementation types (e.g. those derived from `Actor`)
   - The attribute provides a property `TypeName` that can be used to explicitly set the actor type name
 - Adds an `ActorTypeName` property to `ActorTypeInformation`
   - This property is set to either `ActorAttribute.TypeName` if attached/set or the implementation type name
 - Updates all uses of the actor type's name to use `ActorTypeInformation.ActorTypeName`
 - Some refactoring of `ActorRuntime` to allow for isolated unit testing
 - Adds unit tests for `ActorTypeInformation` and `ActorRuntime` related features

Note that the original proposal was a simple `RegisterActor()` overload that accepted the explicit type name.  However, as the implementation started, the type name ended up becoming something more than just a simple registration key and more of an intrinsic property of the actor implementation (and used by components with only access to the actor, not the registration).  Hence, `ActorAttribute` was created to allow overrides, like the type name, to be attached to the implementation in an optional, non-intrusive way, and then surfaced through `ActorTypeInformation`.

## Issue reference

Resolves #164.

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


src/Dapr.Actors/Runtime/ActorTypeInformation.cs
```diff
@@ -23,6 +23,13 @@ public ActorTypeInformation()
         {
         }
 
+        /// <summary>
+        /// Gets the name of the actor type represented by the actor.
+        /// </summary>
+        /// <value>The <see cref="string"/> name of the actor type represented by the actor.</value>
+        /// <remarks>Defaults to the name of the class implementing the actor.</remarks>
```
amanbha: Could you also add the override behavior of type name using attribute here as well in remarks.

src/Dapr.Actors/Runtime/ActorRuntime.cs
```diff
@@ -24,16 +24,16 @@ public class ActorRuntime
         private const string TraceType = "ActorRuntime";
 
         // Map of ActorType --> ActorManager.
-        private static Dictionary<string, ActorManager> actorManagers = new Dictionary<string, ActorManager>();
+        private readonly Dictionary<string, ActorManager> actorManagers = new Dictionary<string, ActorManager>();
 
-        private ActorRuntime()
+        internal ActorRuntime()
```
amanbha: Seems like its made internal for UnitTests? Can UnitTest access using ActorRuntime.Instance? My concern is that in future someone can new up another instance of ActorRuntime in product code by mistake

src/Dapr.Actors/Runtime/ActorStateManager.cs
```diff
@@ -21,7 +21,7 @@ internal sealed class ActorStateManager : IActorStateManager
         internal ActorStateManager(Actor actor)
         {
             this.actor = actor;
-            this.actorType = actor.GetType().Name;
```
amanbha: I understand its not part of this PR, but could you please rename actorType member variable to actorTypeName also with this PR as it seems more logical.

philliphoff: Updated.

src/Dapr.Actors/Runtime/ActorTypeInformation.cs
philliphoff: Updated.

src/Dapr.Actors/Runtime/ActorRuntime.cs
amanbha: This risk is low and can be enforced by adding Warning Xml doc and through review.
Looking at the unit tests being added, could you help me understand how does accumulating state impact the tests?

philliphoff: Once you register an actor type, you can't *unregister* it, so using a common `ActorRuntime` instance means that tests cannot run in complete isolation; they'll always be a risk that a previous test impacts the result of a following one.  (It also tends to prevent tests from being able to run in parallel.)

amanbha: Sure, sounds good. Add a warning in the XML doc of the internal constructor.

philliphoff: Warning added.

PR #175, opened 2019-12-03
Update error codes usage in dotnet-sdk

shaggygi:
Updated the error codes to match changes with https://github.com/dapr/dapr/pull/860

Also added missing error enums and attempted to reorder/group.

There could be some of these that could go through another round of name changes or even consolidation, but I need to review/understand more first.


amanbha:
Thanks @shaggygi I will merge it after https://github.com/dapr/dapr/pull/860 is merged in dapr runtime.

Issue #177, opened 2019-12-03
DemoActor calling RegisterRemind start to error on each remind


amanbha:
Cloning from dapr repo to dotnet-sdk repo
https://github.com/dapr/dapr/issues/864

What version of Dapr?
0.2.x

Expected Behavior
After Registering a reminder, the ReceiveReminder executes without error

Actual Behavior
Solution works great, however, as soon as a reminder is added, the ReceiverReminder fails each time it's invoked. - see attached screen shots of console log

Steps to Reproduce the Problem
Add the line "this.RegisterReminder()" to SaveData, and invoke SaveData once.
RemindFailure
CodeChangeDemoActor

![image](https://user-images.githubusercontent.com/11221966/70095367-6b451f00-15d9-11ea-8590-a5679fae630d.png)
![image](https://user-images.githubusercontent.com/11221966/70095379-71d39680-15d9-11ea-8f72-9b7fe3ce887f.png)

amanbha:
cc @codeputer Thanks for reporting the issue,
I looked at it, the issue is happening because ReminderInfo is serialized synchronously and aspnetcore 3.0 disallows it by default.
I will update the code to async deserializtion.

To unblock, you can change the code where AspNetCore WebHost is created to allow Synchronous IO

Add to `.ConfigureKestrel(options => options.AllowSynchronousIO = true)` to `CreateWebHostBuilder` method so that method will look like:
 

` public static IWebHostBuilder CreateWebHostBuilder(string[] args) =>
            WebHost.CreateDefaultBuilder(args)
                .UseStartup<Startup>()
            .ConfigureKestrel(options => options.AllowSynchronousIO = true)
            .UseActors(actorRuntime =>
            {
                actorRuntime.RegisterActor<DemoActor>();
            });`

codeputer:
I tried this in the ConfigureServices....


   public void ConfigureServices(IServiceCollection services)
    {
      services.AddRouting();
      services.Configure<IISServerOptions>(options =>
      {
        options.AllowSynchronousIO = true;
      });
    }

but as far as I can tell, it didn't have any effect

amanbha:
@codeputer Could you please try as I suggested above?

codeputer:
Working on it....

codeputer:
I'm not sure what is going on... but I put in the changes you suggested, and got a 502 back
With the as the payload (I'm using fiddler to post )
HTTP/1.1 502 Fiddler - Connection Failed
[...]

PR #178, opened 2019-12-04
Making Serialization, Deserialization of ReminderInfo, Timer async and using System.Text.Json for Reminder, Timer and State,

### LINKED ISSUE #179: Make Serialization, Deserialization of ReminderInfo, Timer async
Making Serialization, Deserialization of ReminderInfo, Timer async and use System.Text.Json for them instead of Newtonsoft

amanbha:
Fixes #179
and also handles issue reported in #177 

Summary of Changes:
1. Making Serialization, Deserialization of ReminderInfo, Timer async 
2. using System.Text.Json for Reminder, Timer and State instead of Newtonsoft


src/Dapr.Actors/Runtime/ReminderInfo.cs
```diff
@@ -36,82 +36,62 @@ internal class ReminderInfo
...
-        internal static ReminderInfo Deserialize(Stream stream)
+        internal static async Task<ReminderInfo> DeserializeAsync(Stream stream)
         {
-            // Deserialize using JsonReader as we know the property names in response. Deserializing using JsonReader is most performant.
+            var serializeOptions = new JsonSerializerOptions();
             using var streamReader = new StreamReader(stream);
-            using var reader = new JsonTextReader(streamReader);
-            return reader.Deserialize(GetFromJsonProperties);
-        }
+            var text = await streamReader.ReadToEndAsync();
```
davidfowl: if It’s always utf8 (which you’re defaulting to right now) you can just pass the stream directly to the serializer. You can fallback to the string for non-utf8 encodings

Issue #184, opened 2019-12-11
Add an InvokeClient and InvokeHttpClient implementation


jeastham1993:
Add an InvokeClient to the Dapr.Client library to allow the use of the Invoke API Endpoints in the same way the StateClient currently works.

amanbha:
Closed by PR https://github.com/dapr/dotnet-sdk/pull/185

PR #185, opened 2019-12-11
Issue 184

jeastham1993:
# Description

Added a HttpClient implementation of the Dapr invoke endpoints following the same pattern as the StateClient.

## Issue reference

This issue is related to issue #184 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


src/Dapr.Client/InvokeHttpClient.cs
```diff
@@ -0,0 +1,89 @@
...
+                throw new HttpRequestException($"Failed to invoke method with status code '{response.StatusCode}': {error}.");
+            }
+            else if (!response.IsSuccessStatusCode)
+            {
+                throw new HttpRequestException($"Failed to invoke method with status code '{response.StatusCode}'.");
+            }
+
+            if (response.Content == null || response.Content.Headers?.ContentLength == 0)
+            {
+                // The state store will return empty application/json instead of 204/404.
```
amanbha: Please update the comment, its talking about state.

src/Dapr.Client/InvokeHttpClient.cs
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        /// <summary>
+        /// Invokes a method using the Dapr invoke endpoints using the properties supplied in the  <paramref name="envelope" /> variable.
+        /// </summary>
+        /// <param name="envelope">The envelope containing the invoke request parameters.</param>
+        /// <param name="cancellationToken">A <see cref="CancellationToken" /> that can be used to cancel the operation.</param>
+        /// <typeparam name="TValue">The data type.</typeparam>
+        /// <returns>A <see cref="Task" /> that will return a deserialized object of the reponse from the Invoke request.</returns>
+        public override async Task<TValue> InvokeMethodAsync<TValue>(InvokeEnvelope envelope, CancellationToken cancellationToken = default)
```
amanbha: I think, passing three arguments here, without an InvokeEnvelope should be fine.
The api should handle all 3 types of invokes:
1. No Data and no return type. 
2. No Data but a return type.
3. Data but no return type.

jeastham1993: Hey Aman, I've removed the InvokeEnvelope and added the three separate methods for the different variations as you describe above. I wanted to be explicit with the three different options, hence the three separate methods with broadly the same implementation

src/Dapr.Client/InvokeHttpClient.cs
jeastham1993: Thanks! The downfall of trying to get ahead of myself with a bit of copying and pasting. I debated moving that whole section out into a separate HttpHandler that the State/Invoke/Topic clients could all share.

I may do that if I add a separate issue for a topic client

src/Dapr.Client/InvokeClient.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// <typeparam name="TValue">The data type.</typeparam>
+        /// <returns>A <see cref="Task" /> that will return a deserialized object of the reponse from the Invoke request. If the Dapr response content is null the return type will be null.</returns>
+        public abstract Task<TValue> InvokeMethodAsync<TValue>(string serviceName, string methodName, CancellationToken cancellationToken = default);
+
+        /// <summary>
+        /// Invokes a method using the Dapr invoke endpoints.
+        /// </summary>
+        /// <param name="serviceName">The name of the service to be called in the Dapr invoke request.</param>
+        /// <param name="methodName">THe name of the method to be called in the Dapr invoke request.</param>
+        /// <param name="data">The data to be sent within the body of the Dapr invoke request.</param>
```
amanbha: I am assuming the data to be the serialized json here.
I was thinking if we can make the apis look something like following, what do you think?
`
/// <summary>
        /// Invokes the specified method on target service, Request and Response data is serialized/deserialized using the default System.Text.Json Serialization.
        /// Request & Response can be serialized/deserialized by providing an optional delegate also if user wants custom serialization but not needed for this impl. can be added later if needed.
        /// </summary>
        public Task<TResponse> InvokeService<TRequest, TResponse>(string serviceId, string methodName, TRequest requestData, CancellationToken cancellationToken = default)
        {
            return default;
        }
`

jeastham1993: Yeah, I like that. passing in a strongly typed variable instead of just a random string that could really be anything feels a lot more robust.

amanbha:
Thanks @jeastham1993  for the contribution, this looks good. I see that you have added unit tests which is great. Have you validated it with invoking another service using Dapr.

jeastham1993:
Happy to help @amanbha . I can go through and add clients for the other endpoints as well if you like?

Yep, I built an application locally against my service and invoked a method using the API calls. All seems good to me 👍

amanbha:
> Happy to help @amanbha . I can go through and add clients for the other endpoints as well if you like?
> 
> Yep, I built an application locally against my service and invoked a method using the API calls. All seems good to me 👍

Thats great, I will merge this. Yes, your other contributions to the project are welcome too.

PR #205, opened 2020-01-17
Updating new proto files from dapr repo and fixing the gRPC sample 

### LINKED ISSUE #204: Update Proto files from Dapr runtime for gRPC client
There are new Proto files after multi state stores support , which causes breaking changes in state API.

So need to update Proto files from Dapr runtime for gRPC client in dotnets-sdk.

shalabhms:
# Description

Dapr runtime has now feature to support multi state stores support. Due to that state APIs are changed. We need to update the proto files in dotnet SDK for GRPC client from dapr repo.

## Issue reference

Closes #204 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/src/Dapr.Client.Grpc/Protos/dapr.proto
+++ b/src/Dapr.Client.Grpc/Protos/dapr.proto
@@ -28,18 +28,21 @@ message InvokeServiceResponseEnvelope {
 }
 
 message DeleteStateEnvelope {
-  string key = 1;
-  string etag = 2;
-  StateOptions options = 3;
+  string storeName = 1;
+  string key = 2;
+  string etag = 3;
+  StateOptions options = 4;
 }
 
 message SaveStateEnvelope {
-  repeated StateRequest requests = 1;
+  string storeName = 1;
+  repeated StateRequest requests = 2;
 }
 
 message GetStateEnvelope {
-    string key = 1;
-    string consistency = 2;   
+    string storeName = 1;
+    string key = 2;
+    string consistency = 3;
 }
 
 message GetStateResponseEnvelope {
--- a/samples/Grpc/GrpcClient/Program.cs
+++ b/samples/Grpc/GrpcClient/Program.cs
@@ -18,6 +18,7 @@ namespace GrpcClient
     public class Program
     {
         private static string stateKeyName = "mykey";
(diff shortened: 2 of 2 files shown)


src/Dapr.Client.Grpc/Protos/dapr.proto
```diff
@@ -1,106 +1,109 @@
-syntax = "proto3";
```
amanbha: Could you please update the gRPC sample code with this change with this PR as well so that they are not broken?

codeputer:
Is the DAPR team familiar with the work that @mgravell is doing over at https://github.com/protobuf-net/protobuf-net.Grpc 

I think it would greatly reduce the effort of creating and maintaining the ProtoBuf files.

shalabhms:
> Is the DAPR team familiar with the work that @mgravell is doing over at https://github.com/protobuf-net/protobuf-net.Grpc
> 
> I think it would greatly reduce the effort of creating and maintaining the ProtoBuf files.

I am not aware of that work, will look at the details. thanks for the reference !

src/Dapr.Client.Grpc/Protos/dapr.proto
shalabhms: > Could you please update the gRPC sample code with this change with this PR as well so that they are not broken?

Sure.

amanbha:
> > Is the DAPR team familiar with the work that @mgravell is doing over at https://github.com/protobuf-net/protobuf-net.Grpc
> > I think it would greatly reduce the effort of creating and maintaining the ProtoBuf files.
> 
> I am not aware of that work, will look at the details. thanks for the reference !

Thanks for pointing to it, its an interestign project, I believe that its is for generating proto from a C# code (thus taking a service first approach) Here we are generating clients for the Dapr server which is already written in Go, the cDApr clients would be generated in multiple languages.

src/Dapr.Client.Grpc/Protos/dapr.proto
shalabhms: > Could you please update the gRPC sample code with this change with this PR as well so that they are not broken?

Thanks Aman for pointing this out ! . I have now updated the sample and tested it.

PR #207, opened 2020-01-25
Add support in dotnet client sdk to support multi state store support

### LINKED ISSUE #206: Add support in dotnet client sdk to support multi state store support
There is now multi state stores support feature in Dapr runtime.  
This issue is to add support in dotnet client sdk to support multi state store support.

shalabhms:
# Description

Add support in dotnet client sdk to support multi state store support

## Issue reference

Closes #206 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


samples/AspNetCore/RoutingSample/Startup.cs
```diff
@@ -20,6 +20,11 @@ namespace RoutingSample
     /// </summary>
     public class Startup
     {
+        /// <summary>
+        /// State store name.
+        /// </summary>
+        public const string Storename = "testStore";
```
rynowak: nit: should be `StoreName`

src/Dapr.AspNetCore/FromStateAttribute.cs
```diff
@@ -18,19 +18,28 @@ public class FromStateAttribute : Attribute, IBindingSourceMetadata
...
+        public FromStateAttribute(string stateStoreName, string key)
         {
+            this.StateStoreName = stateStoreName;
             this.Key = key;
         }
 
+        /// <summary>
+        /// Gets the state store name.
+        /// </summary>
+        public string StateStoreName { get; }
```
rynowak: nit: you used `StoreName` elsewhere, I think `StoreName` is better, a little more brief.

src/Dapr.AspNetCore/FromStateAttribute.cs
```diff
@@ -18,19 +18,28 @@ public class FromStateAttribute : Attribute, IBindingSourceMetadata
         /// <summary>
         /// Initializes a new instance of the <see cref="FromStateAttribute"/> class.
         /// </summary>
-        public FromStateAttribute()
+        /// <param name="stateStoreName">The state store name.</param>
+        public FromStateAttribute(string stateStoreName)
         {
+            this.StateStoreName = stateStoreName;
```
rynowak: This should throw if the store name is null/empty, since that won't work.

src/Dapr.AspNetCore/StateEntryApplicationModelProvider.cs
```diff
@@ -33,7 +33,7 @@ public void OnProvidersExecuted(ApplicationModelProviderContext context)
                     }
                     else if (IsStateEntryType(property.ParameterType))
                     {
-                        property.BindingInfo.BindingSource = new FromStateBindingSource(null);
+                        property.BindingInfo.BindingSource = new FromStateBindingSource(null, null);
```
rynowak: This is the piece of code that has to change... We can't use a state store without providing a store name. You should change this if block to throw an exception explaining that `[FromState(...)]` is required.

test/Dapr.AspNetCore.IntegrationTest.App/DaprController.cs
```diff
@@ -25,22 +25,22 @@ public ActionResult<UserInfo> RegisterUser(UserInfo user)
             return user; // echo back the user for testing
         }
 
-        [HttpPost("/controllerwithoutstateentry/{widget}")]
-        public async Task AddOneWithoutStateEntry([FromServices]StateClient state, [FromState] Widget widget)
+        [HttpPost("/controllerwithoutstateentry/{testStore}/{widget}")]
```
rynowak: You don't need to add a parameter to the route, it's not actually doing anything because you're not parsing the state store name from the URL. You should be able to revert just this part of the test.

test/Dapr.AspNetCore.IntegrationTest.App/Startup.cs
```diff
@@ -46,10 +46,10 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
 
                 endpoints.MapPost("/topic-a", context => Task.CompletedTask).WithTopic("A");
 
-                endpoints.MapPost("/routingwithstateentry/{widget}", async context =>
+                endpoints.MapPost("/routingwithstateentry/{testStore}/{widget}", async context =>
```
rynowak: Same comment as the controller, you don't need to add `{testStore}` to the route, it's not doing anything meaningful in the test.

src/Dapr.AspNetCore/StateEntryApplicationModelProvider.cs
amanbha: @ryanvolum Does that mean that following cannot be written without a FromState attribute?
```C#
[HttpGet("{account}")]
public ActionResult<Account> Get(StateEntry<Account> account)
```

It must always be written as:
```C#
[HttpGet("{account}")]
public ActionResult<Account> Get([FromState("storeName")]StateEntry<Account> account)
```

samples/AspNetCore/ControllerSample/Controllers/SampleController.cs
```diff
@@ -15,6 +15,11 @@ namespace ControllerSample.Controllers
     [ApiController]
     public class SampleController : ControllerBase
     {
+        /// <summary>
+        /// State store name.
+        /// </summary>
+        public const string Storename = "testStore";
```
rynowak: nit: StoreName

src/Dapr.AspNetCore/StateEntryApplicationModelProvider.cs
rynowak: correct. The store name is required.

PR #209, opened 2020-01-25
Close #208

### LINKED ISSUE #208: Actor Reminders with `period=TimeSpan.FromMilliseconds(-1)` corrupt state passed to `IRemindable.ReceiveReminderAsync()`
## Expected Behavior

Actors can register reminders. The .NET API indicates that passing a TimeInterval value of -1ms means that the timer is never triggered. My actor code has a `Tick` function, and after receiving `Tick` message, my actor should independently do some long-running stuff. So I decided to call `RegisterReminderAsync` with a DueTime of +1ms (so that the registered core is executed directly after `Tick()` finishes. And I chose to `Period` to be `TimeSpan.FromMilliseconds(-1)`, so that my function is not called on a recurring basis. Also, I'm passing a serialized JSON as state.

## Actual Behavior

When `IRemindable.ReceiveReminderAsync()` is invoked by the runtime, I noticed that the supplied state does not contain my JSON string, but the value `"0h0m0s1ms"`. So somehow, the DueTime slips over into the State. 

## Repro

```csharp
namespace Core.Application.Common.ActorImplementations
{
    using System;
    using System.Text;
    using System.Threading.Tasks;
    using Newtonsoft.Json;
    using Dapr.Actors;
    using Dapr.Actors.Runtime;
    using Core.Domain;

    public class FlightServiceQueryActorState
    {
        public string CoordinatorId { get; set; }

        public byte[] ToBytes() => Encoding.UTF8.GetBytes(JsonConvert.SerializeObject(this));

        public static FlightServiceQueryActorState FromBytes(byte[] bytes) => JsonConvert.DeserializeObject<FlightServiceQueryActorState>(Encoding.UTF8.GetString(bytes));
    }

    [Actor(TypeName = ActorNames.FlightServiceQueryActor)]
    public class FlightServiceQueryActor : Actor, IRemindable, IFlightServiceQueryActor
    {
        public FlightServiceQueryActor(ActorService actorService, ActorId actorId) : base(actorService, actorId) { }

        const string AfterTickReminder = "AfterTick";
        static readonly TimeSpan Never = TimeSpan.FromMilliseconds(-1);

        async Task IFlightServiceQueryActor.Tick(string senderId, int duration)
        {
            var state = new FlightServiceQueryActorState { CoordinatorId = senderId };

            await base.RegisterReminderAsync(
                reminderName: AfterTickReminder, 
                state: state.ToBytes(),
                dueTime: TimeSpan.FromMilliseconds(1),
                period: Never);

            await Console.Out.WriteLineAsync($"Tick tock {this.Id.GetId()}");
        }

        async Task IRemindable.ReceiveReminderAsync(string reminderName, byte[] stateBytes, TimeSpan dueTime, TimeSpan period)
        {
            var state = FlightServiceQueryActorState.FromBytes(stateBytes);

            var coordinator = ActorProxies.CreateFlightTravelCoordinationFor(state.CoordinatorId);

            await Console.Out.WriteLineAsync($"Reminder {this.Id.GetId()} -- {state.CoordinatorId}");
        }
    }
}
```

Result exception is this:

```text
time="2020-01-25T11:37:35+01:00" level=error msg="error executing reminder: error from actor service: "
fail: Microsoft.AspNetCore.Server.Kestrel[13]
      Connection id "0HLT1GA8OGV6N", Request id "0HLT1GA8OGV6N:00000006": An unhandled exception was thrown by the application.
System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values.
   at Dapr.Actors.Runtime.ConverterUtils.ConvertTimeSpanFromDaprFormat(String valueString) in C:\github\dapr\dotnet-sdk\src\Dapr.Actors\Runtime\ConverterUtils.cs:line 28
   at Dapr.Actors.Runtime.ReminderInfo.DeserializeAsync(Stream stream) in C:\github\dapr\dotnet-sdk\src\Dapr.Actors\Runtime\ReminderInfo.cs:line 56
   at Dapr.Actors.Runtime.ActorManager.FireReminderAsync(ActorId actorId, String reminderName, Stream requestBodyStream, CancellationToken cancellationToken) in C:\github\dapr\dotnet-sdk\src\Dapr.Actors\Runtime\ActorManager.cs:line 162
   at Dapr.Actors.AspNetCore.RouterBuilderExtensions.<>c.<<AddReminderRoute>b__4_0>d.MoveNext() in C:\github\dapr\dotnet-sdk\src\Dapr.Actors.AspNetCore\RouterBuilderExtensions.cs:line 101
```
## Root cause

The current code serializes a negative `TimeSpan` into an empty string, which results subsequently in an `ArgumentOutOfRangeException` within `ConverterUtils.ConvertTimeSpanFromDaprFormat`. I changed that implementation to return a `default(TimeSpan)` when an empty or null string is passed in. However, looking through the code, I'm not sure whether it should instead be the -1ms.

chgeuer:
# Description

## Issue reference

This will close #208 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


amanbha:
Thanks @chgeuer  for the PR,

src/Dapr.Actors/Runtime/ActorTimer.cs
```diff
@@ -48,12 +48,12 @@ internal async Task<string> SerializeAsync()
 
             writer.WriteStartObject();
 
-            if (this.DueTime != null)
+            if (this.DueTime != null && this.DueTime >= TimeSpan.Zero)
```
amanbha: dueTime will always be >=0, Reminders do a check,(https://github.com/dapr/dotnet-sdk/blob/master/src/Dapr.Actors/Runtime/ReminderInfo.cs#L93). There should be a similar check for Timer dueTime also.

src/Dapr.Actors/Runtime/ConverterUtils.cs
```diff
@@ -11,6 +11,11 @@ internal class ConverterUtils
     {
         public static TimeSpan ConvertTimeSpanFromDaprFormat(string valueString)
         {
+            if (string.IsNullOrEmpty(valueString))
+            {
+                return default; // TODO: Should that be TimeSpan.FromMilliseconds(-1); DAPR spec says -1ms == never
```
amanbha: yes, it could be -1, it would happen for period, for dueTime , it would never be empty as SDK will pass a value and dapr will return it.

PR #218, opened 2020-02-12
Updating readme for Actor Sample to clarify client invocation.

### LINKED ISSUE #216: Clarify client behavior/way to run in sample
Clarify actor + client sample here:  https://github.com/dapr/dotnet-sdk/tree/master/samples/Actor.

The "Running the sample" and "Making client calls" sections have a number of places where assumptions are made and things could be clearer.
1) If you run a client containing an ActorProxy without dapr run, there are some things the user needs to do first to specify the port.  It happens to work in the sample because the actor app uses 3500.
2) Indicate how to run this client app could be run using dapr run.

See https://github.com/dapr/dapr/issues/1009

amanbha:
Fixes #216

### DIFF
--- a/samples/Actor/Readme.md
+++ b/samples/Actor/Readme.md
@@ -47,6 +47,12 @@ Run the client project from `ActorClient` directory as:
  dotnet run
  ```
 
+ *Note: If you  started the actor service with dapr port other than 3500, then set the environment variable DAPR_HTTP_PORT to the value of --port specified while starting the actor service before running the client in terminal.*
+ ```
+ On Windows: set DAPR_HTTP_PORT=<port>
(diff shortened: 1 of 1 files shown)


samples/Actor/Readme.md
```diff
@@ -47,6 +47,12 @@ Run the client project from `ActorClient` directory as:
  dotnet run
  ```
 
+ *Note: If you  started the actor service with dapr port other than 3500, then set the environment variable DAPR_HTTP_PORT to the value of --port specified while starting the actor service before running the client in terminal.*
```
LMWF: It looks like there's an extra space between "you" and "started".

PR #220, opened 2020-02-13
Malotho/making actorstestable

malotho:
# Description

I have changed the constructor on the Actor class to take an optional IActorStateManager. 
The constructor will use the provided IActorStateManager if passed, otherwise a new actorstatemanager instance will be created. 

Please reference the issue this PR will close: #173

Issue #173, opened 2019-11-26
Make it possible to create unit tests for Actors


Xorpio:
Right now i can't create a actor inside a unit test. Take the following Actor code for example

    public class TestActor : Actor
    {
        public TestActor(ActorService actorService, ActorId actorId) : base(actorService, actorId)
        {
        }

        public async Task<string> GetString()
        {
            return await StateManager.GetStateAsync<string>("Test");
        }
    }
And using the following unit test:

        [Test]
        public async Task CanCreateActor()
        {
            var actor = new TestActor(
                new ActorService(
                    new ActorTypeInformation()
                ),
                new ActorId("Test")
            );

            Assert.IsInstanceOf<Actor>(actor);
        }

When i run the unit test it throws a NullRefference exception

```
Message: 
    System.NullReferenceException : Object reference not set to an instance of an object.
  Stack Trace: 
    Actor.ctor(ActorService actorService, ActorId actorId)
    TestActor.ctor(ActorService actorService, ActorId actorId) line 12
    IAMEmployeeTest.CanCreateActor() line 19
    GenericAdapter`1.GetResult()
    AsyncToSyncAdapter.Await(Func`1 invoke)
    TestMethodCommand.RunTestMethod(TestExecutionContext context)
    TestMethodCommand.Execute(TestExecutionContext context)
    SimpleWorkItem.PerformWork()
```

amanbha:
@Xorpio Thanks for reporting. Will using Moq and setting them correctly for ActorService help in this case? Or do you have other suggestions in mind?

Xorpio:
The problem is that there are some hardcoded dependencies that cannot be faked. (as far as i know). Like on this line:
https://github.com/dapr/dotnet-sdk/blob/master/src/Dapr.Actors/Runtime/Actor.cs#L42

malotho:
Any traction on this?

malotho:
@Xorpio please check out this work. 
https://github.com/Ignition-Group-Open-Source-Contrib/DaprDotnetActorUnitests. Code coverage, and 100% testebility on the actor.

Xorpio:
Oh nice work. it looks like a working solution. I can look further in depth in the coming days.

malotho:
@amanbha should this issue be closed?

amanbha:
@malotho I would suggest to keep it open to track more improvements needed in SDK itself to make it easier to write unit tests.

malotho:
@amanbha that makes sense.

[...]

PR #229, opened 2020-02-19
Malotho/making actorstestable

malotho:
# Description

https://github.com/dapr/dotnet-sdk/pull/220#issuecomment-586623718
This is a fix to the null reference exception which was caused by an incorrect actor dependency assignment order.

## Issue reference

This is in relation to issue #173


yaron2:
cc @amanbha

amanbha:
@malotho Thanks for fixing the initialization order. Could you please update your branch as well before the merge.

malotho:
@amanbha I will merge the branches and let you know when done. Thanks

malotho:
@amanbha @yaron2

yaron2:
Thanks!

PR #250, opened 2020-03-13
Adding ref. to System.IO.Pipelines explicitly to work around build issue with .netcore sdk 3.1.102

amanbha:
Adding ref. to System.IO.Pipelines explicitly to work around build issue with .netcore sdk 3.1.102
https://github.com/dotnet/core/blob/master/release-notes/3.1/3.1-known-issues.md#net-core-312

This will be reverted when the fix will be available in future .net core sdk.


src/Dapr.Actors.AspNetCore/Dapr.Actors.AspNetCore.csproj
```diff
@@ -11,6 +11,9 @@
     <Description>This package contains the reference assemblies for developing Actor services using Dapr and AspNetCore.</Description>
     <PackageTags>$(PackageTags);Actors</PackageTags>
   </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="System.IO.Pipelines" Version="4.7.0" />
```
LMWF: Can you add a comment here saying it's a workaround and when it can be removed?

Issue #249, opened 2020-03-13
There should be exception if actor method has more than one parameter


wanton7:
## Expected Behavior

Exception if actor method has more than one parameter. Example exception for `IUserActor.TryCreate(string email, string name)`

## Actual Behavior

There is JSON deserializing exception if you call method with more than one parameter like this
>{"errorCode":"ERR_ACTOR_INVOKE_METHOD","message":"error from actor service: System.Text.Json.JsonException: The JSON value could not be converted to System.String. Path: $ | LineNumber: 0 | BytePositionInLine: 1.\

## Steps to Reproduce the Problem

Create actor method with more than one parameter and call it.
`curl -H "Content-Type: application/json" -d "{\"email\":\"foo@bar.com\",\"name\":\"user name\"}" -X POST http://localhost:3500/v1.0/actors/user/0123456789abcdef0123456789abcdef/method/TryCreate`

amanbha:
@wanton7 Actor methods with 2 ards can be called using remoting apis. The restriction is only for methods called using the non-remoting apis in proxy.

For my understanding: Are you saying throw a more meaningful exception when its invoked over http and it has multiple args.

wanton7:
Yes exactly meaningful exception. Now it's just crashing in JSON deserializer and I didn't understand what was happening and what did I do wrong. Should that single parameter be checked that it's a class or struct as well?

amanbha:
@wanton7 Yes, sure, it makes sense, I will look into it. Feel free to create a PR as well if you want to contribute

wanton7:
It seems that checking for struct is not that easy because example DateTime is a struct. I haven't done pull requests to github before but it could be nice to try. If I will just do pull request to check for max one parameter.

PR #252, opened 2020-03-13
Throw exception if actor method has more than one parameter

wanton7:
# Description
Throw exception if actor method has more than one parameter when invoked through http

## Issue reference

#249 

## Checklist

* [X] Code compiles correctly
* [ ] Created/updated tests


amanbha:
@wanton7  Thanks for the contribution.

Issue #255, opened 2020-03-15
Standard Dependency Injection for Actors


chleit:
## Describe the proposal
The ActorService should instantiate actors through the ASP.Net Core DI container and inject the ActorId through a property or method.

Yes I know that you can use ```actorRuntime.RegisterActor<DemoActor>( (type) => new ActorService(type, (actorService, actorId) => new DemoActor(actorService, actorId)));```

But the problem IMHO is that at the place you register the actor there is AFAIK no easy way to access the ASP.Net Core DI container.

Maybe there is another way to make the actor instantiation play better with dependency injection. I'm open for discussion.

amanbha:
@chleit Thats a good idea. The curent approach was chosena s the Actors library itself is not dependent on Aspnet core.

pohorely:
Hello,
accessing DI is not so difficult. We can register actors by a factory method.

Here is working example (Dapr v0.6.0):

**Usage**
```csharp
// Program.cs
    public class Program
    {
        public static async Task Main(string[] args)
        {
            CreateHostBuilder(args).Build().Run();
        }

        public static IHostBuilder CreateHostBuilder(string[] args) =>
            Host.CreateDefaultBuilder(args)
                .ConfigureWebHostDefaults(webBuilder =>
                {
                    webBuilder.UseStartup<Startup>()
                        .UseActors(actorRuntime => { });
                });
    }

// Startup.cs
    public class Startup
    {
        public Startup(IConfiguration configuration)
        {
            Configuration = configuration;
        }

        public IConfiguration Configuration { get; }

        public void ConfigureServices(IServiceCollection services)
        {
            services.AddActor(MyActorFactory);

            // other registrations
        }

        public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
        {
            // without change
        }

        private static MyActor MyActorFactory(ActorService service, ActorId id, IServiceProvider provider) =>
            new MyActor(
                service,
                id,
                provider.GetRequiredService<IMyService>());
    }

    public class MyActor : Dapr.Actors.Runtime.Actor, IMyActor
    {
        private readonly IMyService _myService;

        public MyActor(ActorService actorService, ActorId actorId, IMyService myService) : base(actorService, actorId)
        {
            _myService = myService;
        }

        // business logic methods
    }
```
**Implementation**
`AddActor` extension method implementation:
```csharp
// ReSharper disable once CheckNamespace
namespace Microsoft.Extensions.DependencyInjection
{
[...]

PR #260, opened 2020-03-19
[WIP] 255 standard dependency injection

chleit:
# Description

Implemented changes to support standard .Net Core Dependency Injection

@amanbha: First draft for the changes we discussed yesterday. Tests are partially commented out. Manual test was  #successful.

## Issue reference

#255

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/src/Dapr.Actors/Runtime/Actor.cs
+++ b/src/Dapr.Actors/Runtime/Actor.cs
@@ -20,8 +20,8 @@ namespace Dapr.Actors.Runtime
     public abstract class Actor
     {
         private const string TraceType = "Actor";
-        private readonly string traceId;
-        private readonly string actorImplementaionTypeName;
+        private string traceId;
+        private string actorImplementationTypeName;
 
         /// <summary>
         /// Contains timers to be invoked.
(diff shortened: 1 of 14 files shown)


amanbha:
@chleit Thanks for the contribution and first draft, I will review these.

src/Dapr.Actors/Runtime/Actor.cs
```diff
@@ -20,41 +20,25 @@ namespace Dapr.Actors.Runtime
...
-            this.StateManager = actorStateManager ?? new ActorStateManager(this);
-            this.actorImplementaionTypeName = this.ActorService.ActorTypeInfo.ImplementationType.Name;
-        }
-
         /// <summary>
         /// Gets the identity of this actor.
         /// </summary>
         /// <value>The <see cref="ActorId"/> for the actor.</value>
-        public ActorId Id { get; }
+        public ActorId Id { get; private set; }
```
amanbha: How would user provide its own constructor for Actor? In its constructor, ActorId will not be available>

chleit: I think there are several ways for that:

- Make the setter for ActorId public - I wouldn't do that
- ActorId is currently set in OnActivateInternalAsync() - make that method public
- Extract the code that was originally in the Actor() constructor from OnActivateInternalAsync() to a public method ConfigureActor(actorId, actorService, ...)

I personally would go for option 3.

chleit: I know this is kind of "not object oriented programming" because the Actor would not be fully instantiated by the constructor but would need an extra ConfigureActor() call.
But the problem with this basic kind of IoC container IServiceProvider implements, that AFAIK you can't easily inject instance runtime information (like an ActorId).

amanbha: User does not know the actor id, it needs to be provided in the constructor when a call comes for it, OnActivateAsync seems too late in the actor lifecyle.

chleit: OK, I have to differentiate:

- In normal (DAPR) operation, the Actor is instantiated and activated in the ActorManager:
```
        internal async Task ActivateActor(ActorId actorId)
        {
            // An actor is activated by "Dapr" runtime when a call is to be made for an actor.
            var actor = this.actorService.CreateActor();
            await actor.OnActivateInternalAsync(this.actorService, actorId);
        ...
        }
```
So there's practically no difference between providing the ActorService and ActorId in the constructor or the call to OnActivateInternalAsync(). So we also could easily have the instantiation and activation like this:
```
        internal async Task ActivateActor(ActorId actorId)
        {
            // An actor is activated by "Dapr" runtime when a call is to be made for an actor.
            var actor = this.actorService.CreateActor();
            actor.Configure(this.actorService, actorId);
            await actor.OnActivateInternalAsync();
        ...
        }
```

- For unit-testing the user has to provide the ActorService (or a mock thereof) and an ActorId. So for that we need same public methode like actor.Configure(...)

amanbha: Let me think more about it, when the user actor constructor is called, they can choose to load  things based on the actor id.
IMO, public Configure method for mocking doesnt seem like the right thing to do.

amanbha:
@chleit just fyi, sorry for the delay, I am still looking at it.

amanbha:
cc @malotho  as well.

Issue #261, opened 2020-03-19
New gRPC DaprClient fails for Secret API requests if mTLS is enabled 


cmendible:
## Expected Behavior

DaprClient should work with mTLS enabled.

## Actual Behavior

clients created through the builder:

```chsharp
var client = new DaprClientBuilder().Build();
```

Fail to get secrets through calling `client.GetSecretAsync` when Dapr mTLS is enabled.

## Steps to Reproduce the Problem

Deploy the following yaml to K8s:

```yaml
---
# Create a Kubernetes Secret
apiVersion: v1
kind: Secret
metadata:
  name: super-secret
  namespace: default
type: Opaque
data:
  super-secret: eW91ciBzdXBlciBzZWNyZXQK

---
# If RBAC is enabled in K8s, give the default SA access to secrets in the default namespace
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: dapr-secret-reader
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: secret-reader
subjects:
  - kind: ServiceAccount
    name: default
    namespace: default
```

Create and deploy an ASP.NET Core application which implements the following call:

```chshap
var secrets = await client.GetSecretAsync("kubernetes", "super-secret", new Dictionary<string, string>() { { "namespace", "default" } });
```

The call will fail with mTLS enabled.

amanbha:
@cmendible I believe, it would not be specific to dotnet sdk as the calls from user service code to Dapr is not secured.

cmendible:
@amanbha should I post a bug or question on the Dapr repo?

amanbha:
> @amanbha should I post a bug or question on the Dapr repo?

Yes, it would be in dapr/dapr repo as the calls to dapr sidecar from user service code is not secured. Is the issue only with secret api or with any api?

cmendible:
Roger that. 

Only tested the secret api.

yaron2:
I closed the issue on `dapr/dapr` *for now* as this is likely an SDK usage issue with gRPC.

Notice the client config used here to talk to Dapr from a client when gRPC is enabled:
https://github.com/dapr/docs/blob/master/howto/create-grpc-app/README.md#invoking-dapr---go-example

Look at the mTLS enabled part. You need to provide a TLS config with `InsecureSkipVerify`.

yaron2:
[...]

PR #263, opened 2020-03-21
Workaround for mTLS enabled Dapr endpoint.

amanbha:
Allows the DaprClient to work with mTLS enabled grpc endpoint when https endpoint is provided with DaprClientBuilder.UseEndpoint
It will be cherry-picked into 0.5.0 branch to realse a new nuget package.
The change will be reverted from master branch for 0.6.0 when the issue is fixed in Dapr runtime.

Fixes https://github.com/dapr/dotnet-sdk/issues/261


cmendible:
@amanbha Just tested this and it's working!!!

Issue #219, opened 2020-02-13
Support the granular Actor config such as actor idle time


youngbupark:
## Describe the proposal

dotnet-sdk doesn't provide the option to configure actor configurations such as idle timeout, scaninterval, etc. So, in the scenarios which need longer actor idle time, users cannot configure idle time.

https://github.com/dapr/docs/blob/master/reference/api/actors.md#get-registered-actors

```
{
  "actorIdleTimeout": "1h",
  "actorScanInterval": "30s",
  "drainOngoingCallTimeout": "30s",
  "drainRebalancedActors": true
}
```

PR #269, opened 2020-03-28
Support more actor configuration

LMWF:
# Description

This PR gives the user the option to specify these actor configuration values:
actorIdleTimeout
actorScanInterval
drainOngoingCallTimeout
drainRebalancedActors

## Issue reference

Please reference the issue this PR will close: #219

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/src/Dapr.Actors/Runtime/ActorConfig.cs
+++ b/src/Dapr.Actors/Runtime/ActorConfig.cs
@@ -1,0 +1,116 @@
+﻿// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+// ------------------------------------------------------------
+
+namespace Dapr.Actors.Runtime
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Text.Json;
+    using System.Threading.Tasks;
+
+    /// <summary>
+    /// Represents Dapr actor configuration for this app.  See
+    /// https://github.com/dapr/docs/blob/master/reference/api/actors_api.md
+    /// </summary>
+    public sealed class ActorConfig
+    {
+        /// <summary>
+        /// Default constructor.  This is only here for the unit test.
+        /// </summary>
+        private ActorConfig()
+            : this(null, null, null, false)
+        { }
+
+        /// <summary>
+        /// Users should use this contstructor to override default values.
+        /// </summary>
+        /// <param name="actorIdleTimeout"></param>
+        /// <param name="actorScanInterval"></param>
+        /// <param name="drainOngoingCallTimeout"></param>
+        /// <param name="drainBalancedActors"></param>
+        public ActorConfig(
+            TimeSpan? actorIdleTimeout = null,
+            TimeSpan? actorScanInterval = null,
+            TimeSpan? drainOngoingCallTimeout = default,
+            bool drainBalancedActors = default)
+        {
+            this.RegisteredActorTypes = new List<string>();
+            this.ActorIdleTimeout = actorIdleTimeout;
+            this.ActorScanInterval = actorScanInterval;
+            this.DrainOngoingCallTimeout = drainOngoingCallTimeout;
+            this.DrainBalancedActors = drainBalancedActors;
+        }
+
+        /// <summary>
+        /// The actor types to register with Dapr.
+        /// </summary>
+        public List<string> RegisteredActorTypes { get; private set; }
+
+        /// <summary>
+        /// Specifies how long to wait before deactivating an idle actor. An actor is idle 
+        /// if no actor method calls and no reminders have fired on it.
+        /// </summary>
+        public TimeSpan? ActorIdleTimeout { get; set; }
+
+        /// <summary>
+        /// A duration which specifies how often to scan for actors to deactivate idle actors. 
+        /// Actors that have been idle longer than the actorIdleTimeout will be deactivated.
+        /// </summary>
+        public TimeSpan? ActorScanInterval { get; set; }
+
+        /// <summary>
+        /// A duration used when in the process of draining rebalanced actors. This specifies 
+        /// how long to wait for the current active actor method to finish. If there is no current actor method call, this is ignored.
+        /// </summary>
+        public TimeSpan? DrainOngoingCallTimeout { get; set; }
+
+        /// <summary>
+        /// A bool. If true, Dapr will wait for drainOngoingCallTimeout to allow a current 
+        /// actor call to complete before trying to deactivate an actor. If false, do not wait.
+        /// </summary>
+        public bool DrainBalancedActors { get; set; }
+
+        internal async Task SerializeAsync(System.Buffers.IBufferWriter<byte> output)
+        {
+            using Utf8JsonWriter writer = new Utf8JsonWriter(output);
+            writer.WriteStartObject();
+
+            // array 
+            writer.WritePropertyName("entities");
+            writer.WriteStartArray();
+
+            foreach (var actorType in this.RegisteredActorTypes)
+            {
+                writer.WriteStringValue(actorType);
+            }
+
+            writer.WriteEndArray();
+
+            if (this.ActorIdleTimeout != null)
+            {
+                writer.WriteString("actorIdleTimeout", ConverterUtils.ConvertTimeSpanValueInDaprFormat(this.ActorIdleTimeout));
+            }
+
+            if (this.ActorScanInterval != null)
+            {
+                writer.WriteString("actorScanInterval", ConverterUtils.ConvertTimeSpanValueInDaprFormat(this.ActorScanInterval));
+            }
+
+            if (this.DrainOngoingCallTimeout != null)
+            {
+                writer.WriteString("drainOngoingCallTimeout", ConverterUtils.ConvertTimeSpanValueInDaprFormat(this.DrainOngoingCallTimeout));
+            }
+
+            // default is true, don't write it if default
+            if (this.DrainBalancedActors != true)
+            {
+                writer.WriteBoolean("drainBalancedActors", (this.DrainBalancedActors));
+            }
+
+            writer.WriteEndObject();
+            await writer.FlushAsync();
+        }
+    }
+}
--- a/src/Dapr.Actors/Runtime/ActorConfig.cs
+++ b/src/Dapr.Actors/Runtime/ActorConfig.cs
@@ -24,7 +24,7 @@ namespace Dapr.Actors.Runtime
         { }
 
         /// <summary>
-        /// Users should use this contstructor to override default values.
+        /// Users should use this constructor to override default values.
         /// </summary>
         /// <param name="actorIdleTimeout"></param>
         /// <param name="actorScanInterval"></param>
--- a/src/Dapr.Actors/Runtime/ActorRuntime.cs
+++ b/src/Dapr.Actors/Runtime/ActorRuntime.cs
@@ -8,6 +8,7 @@ namespace Dapr.Actors.Runtime
     using System;
     using System.Collections.Generic;
     using System.IO;
+    using System.Text.Json;
     using System.Threading;
     using System.Threading.Tasks;
 
@@ -26,7 +27,7 @@ namespace Dapr.Actors.Runtime
         // Map of ActorType --> ActorManager.
         private readonly Dictionary<string, ActorManager> actorManagers = new Dictionary<string, ActorManager>();
 
-        private ActorConfig actorConfig;
+        private ActorSettings actorSettings;
 
         /// <remarks>
         /// WARNING: This type is expected to be accessed via the <see cref="Instance" /> singleton instance.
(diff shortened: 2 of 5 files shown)


src/Dapr.Actors/Runtime/ActorConfig.cs
```diff
@@ -0,0 +1,116 @@
...
+        /// <param name="actorScanInterval"></param>
+        /// <param name="drainOngoingCallTimeout"></param>
+        /// <param name="drainBalancedActors"></param>
+        public ActorConfig(
+            TimeSpan? actorIdleTimeout = null,
+            TimeSpan? actorScanInterval = null,
+            TimeSpan? drainOngoingCallTimeout = default,
+            bool drainBalancedActors = default)
+        {
+            this.RegisteredActorTypes = new List<string>();
```
amanbha: Registered ActorTypes should stay in ActorRuntime and should be exposed in ActorConfig

src/Dapr.Actors/Runtime/ActorConfig.cs
```diff
@@ -0,0 +1,116 @@
...
+
+            if (this.DrainOngoingCallTimeout != null)
+            {
+                writer.WriteString("drainOngoingCallTimeout", ConverterUtils.ConvertTimeSpanValueInDaprFormat(this.DrainOngoingCallTimeout));
+            }
+
+            // default is true, don't write it if default
+            if (this.DrainBalancedActors != true)
+            {
+                writer.WriteBoolean("drainBalancedActors", (this.DrainBalancedActors));
```
amanbha: its "drainRebalancedActors" in dapr runtime.

src/Dapr.Actors/Runtime/ActorConfig.cs
```diff
@@ -0,0 +1,116 @@
...
+            {
+                writer.WriteString("actorScanInterval", ConverterUtils.ConvertTimeSpanValueInDaprFormat(this.ActorScanInterval));
+            }
+
+            if (this.DrainOngoingCallTimeout != null)
+            {
+                writer.WriteString("drainOngoingCallTimeout", ConverterUtils.ConvertTimeSpanValueInDaprFormat(this.DrainOngoingCallTimeout));
+            }
+
+            // default is true, don't write it if default
```
amanbha: I believe its default value is false in runtime, could you please confirm..

src/Dapr.Actors/Runtime/ActorConfig.cs
```diff
@@ -0,0 +1,116 @@
...
+    using System;
+    using System.Collections.Generic;
+    using System.Text.Json;
+    using System.Threading.Tasks;
+
+    /// <summary>
+    /// Represents Dapr actor configuration for this app.  See
+    /// https://github.com/dapr/docs/blob/master/reference/api/actors_api.md
+    /// </summary>
+    public sealed class ActorConfig
```
amanbha: rename to ActorSettings

src/Dapr.Actors/Runtime/ActorRuntime.cs
```diff
@@ -63,6 +66,26 @@ public void RegisterActor<TActor>(Func<ActorTypeInformation, ActorService> actor
 
             // Create ActorManagers, override existing entry if registered again.
             this.actorManagers[actorTypeInfo.ActorTypeName] = new ActorManager(actorService);
+            this.actorConfig.RegisteredActorTypes.Add(actorTypeInfo.ActorTypeName);
+        }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        public void UseActorConfig(ActorConfig config)
```
amanbha: rename to UseActorSettings.
It would be better if  we can create ActorSettings and then pass it to user, user can then configure it by doing:
runtime.ConfigureActorSettings(Action<ActorSettings> configureActorSettings)

That way they dont need to worry about creating ActorSettings and just change what they want to change on ActorSettings object given to them.

src/Dapr.Actors/Runtime/ActorConfig.cs
```diff
@@ -0,0 +1,116 @@
...
+        /// Specifies how long to wait before deactivating an idle actor. An actor is idle 
+        /// if no actor method calls and no reminders have fired on it.
+        /// </summary>
+        public TimeSpan? ActorIdleTimeout { get; set; }
+
+        /// <summary>
+        /// A duration which specifies how often to scan for actors to deactivate idle actors. 
+        /// Actors that have been idle longer than the actorIdleTimeout will be deactivated.
+        /// </summary>
+        public TimeSpan? ActorScanInterval { get; set; }
```
amanbha: Please add validation on setters for these.

src/Dapr.Actors/Runtime/ActorConfig.cs
LMWF: Yes good catch.  I got this name from a test in the runtime repo which his it misnamed too.  Opened an issue in the other repo.

src/Dapr.Actors/Runtime/ActorConfig.cs
LMWF: Yes, default is false.  Fixed.

src/Dapr.Actors/Runtime/ActorConfig.cs
LMWF: Is this a convention?  "Config" conveys the meaning of this object and it more closely resembles the name of the endpoint (which is not a goal, but it's good when it can).

amanbha: Not a convention, keeping the naming in line with SF actors.

src/Dapr.Actors/Runtime/ActorRuntime.cs
LMWF: I'm not sure I understand the benefit here.  If the user doesn't specify anything, all properties are default.  If we were to pass them a prepopulated object, it would have all default (null, etc) values too.

amanbha: They don't need to new up the settings class.

LMWF: It seems pretty natural and normal to create an object like this.  Unless this is a convention I'd like to keep it this way.

amanbha: There is no convention, but when users create an object, they need to think about setting values for all of these but when an object is given, they just modify what they need to and assume default is already configured for it.

src/Dapr.Actors/Runtime/ActorRuntime.cs
```diff
@@ -65,6 +69,41 @@ public void RegisterActor<TActor>(Func<ActorTypeInformation, ActorService> actor
...
         }
 
+        /// <summary>
+        /// 
+        /// </summary>
+        public void UseActorSettings(ActorSettings settings)
+        {
+            ArgumentVerifier.ThrowIfNull(settings, nameof(settings));
+
+            this.actorSettings = new ActorSettings(
```
amanbha: why new it again, its already given.

src/Dapr.Actors/Runtime/ActorRuntime.cs
```diff
@@ -65,6 +69,41 @@ public void RegisterActor<TActor>(Func<ActorTypeInformation, ActorService> actor
...
+                settings.ActorIdleTimeout,
+                settings.ActorScanInterval,
+                settings.DrainOngoingCallTimeout,
+                settings.DrainRebalancedActors);
+        }
+
+        internal Task SerializeActorSettingsAsync(System.Buffers.IBufferWriter<byte> output)
```
amanbha: Consider renaming to SerializeSettingsAndRegisteredTypes?

src/Dapr.Actors/Runtime/ActorRuntime.cs
```diff
@@ -26,12 +27,15 @@ public class ActorRuntime
...
 
+        private ActorSettings actorSettings;
+
         /// <remarks>
         /// WARNING: This type is expected to be accessed via the <see cref="Instance" /> singleton instance.
         /// This constructor is exposed only for unit testing purposes.
         /// </remarks>
         internal ActorRuntime()
         {
+            this.actorSettings = new ActorSettings();
```
amanbha: Ideally this should be created, only when user hasn't provided one with UseActorSettings. If we are creating a default instance then exposing it through ConfigureACtorSettings as discussed [here](https://github.com/dapr/dotnet-sdk/pull/269#discussion_r399728003) would make more sense

src/Dapr.Actors/Runtime/ActorSettings.cs
```diff
@@ -0,0 +1,109 @@
...
+    /// <summary>
+    /// Represents Dapr actor configuration for this app.  See
+    /// https://github.com/dapr/docs/blob/master/reference/api/actors_api.md
+    /// </summary>
+    public sealed class ActorSettings
+    {
+        /// <summary>
+        /// Default constructor.  This is only here for the unit test.
+        /// </summary>
+        private ActorSettings()
```
amanbha: This is not used by unit tests so can be removed.

src/Dapr.Actors/Runtime/ActorRuntime.cs
```diff
@@ -65,6 +68,44 @@ public void RegisterActor<TActor>(Func<ActorTypeInformation, ActorService> actor
...
+            writer.WriteStartArray();
+
+            foreach (var actorType in this.RegisteredActorTypes)
+            {
+                writer.WriteStringValue(actorType);
+            }
+
+            writer.WriteEndArray();
+
+            if (this.actorSettings != null)
```
amanbha: Could you please add a comment that actor settings is only set when provided by user using the UseActorSettings method, would have to look at the code to understand this.

src/Dapr.Actors/Runtime/ActorRuntime.cs
```diff
@@ -65,6 +68,44 @@ public void RegisterActor<TActor>(Func<ActorTypeInformation, ActorService> actor
             this.actorManagers[actorTypeInfo.ActorTypeName] = new ActorManager(actorService);
         }
 
+        /// <summary>
+        /// 
+        /// </summary>
+        public void UseActorSettings(ActorSettings settings)
```
amanbha: Thinking more about it, I am more inclined towards ConfigureSettings method as it allows us(in future) to use ActorSettings with non null defaults and users can modify whatever they want on the object.

src/Dapr.Actors/Runtime/ActorSettings.cs
```diff
@@ -0,0 +1,109 @@
...
+        /// </summary>
+        public TimeSpan? DrainOngoingCallTimeout { get; private set; }
+
+        /// <summary>
+        /// A bool. If true, Dapr will wait for drainOngoingCallTimeout to allow a current 
+        /// actor call to complete before trying to deactivate an actor. If false, do not wait.
+        /// </summary>
+        public bool DrainRebalancedActors { get; private set; }
+
+        internal void Serialize(System.Buffers.IBufferWriter<byte> output, Utf8JsonWriter writer)
```
amanbha: Usually serialize method will add start and end object tags so it's not serializing the complete object. Please merge this code directly into SerializeSettingsAndRegisteredTypes(System.Buffers.IBufferWriter

src/Dapr.Actors/Runtime/ActorSettings.cs
```diff
@@ -0,0 +1,109 @@
...
+        /// </summary>
+        /// <param name="actorIdleTimeout">Specifies how long to wait before deactivating an idle actor. An actor is idle if no actor method calls and no reminders have fired on it.</param>
+        /// <param name="actorScanInterval">A duration which specifies how often to scan for actors to deactivate idle actors. 
+        /// Actors that have been idle longer than the actorIdleTimeout will be deactivated.</param>
+        /// <param name="drainOngoingCallTimeout">A duration used when in the process of draining rebalanced actors. This specifies 
+        /// how long to wait for the current active actor method to finish. If there is no current actor method call, this is ignored.</param>
+        /// <param name="drainRebalancedActors">A bool. If true, Dapr will wait for drainOngoingCallTimeout to allow a current 
+        /// actor call to complete before trying to deactivate an actor. If false, do not wait.</param>
+        public ActorSettings(
+            TimeSpan? actorIdleTimeout = null,
```
amanbha: Don't add this constructor, let users set the values through Properties, move the validations to setter of properties.
If we change the constructor to accept more values or change the order in such a way that its not a compile time error for users, it will cause issues.

src/Dapr.Actors/Runtime/ActorRuntime.cs
LMWF: Example:

```
            actorRuntime.ConfigureActorSettings(a =>
            {
                a.SetActorIdleTimeout(TimeSpan.FromSeconds(33));
                a.SetActorScanInterval(TimeSpan.FromSeconds(44));
                a.SetDrainOngoingCallTimeout(TimeSpan.FromSeconds(55));
                a.SetDrainRebalancedActors(true);
            });
```

PR #270, opened 2020-03-29
Revert the workaround added to work with mTLS enabled grpc endpoint.

amanbha:
Summary of changes:
- This reverts the workaroung which was added to handle mtls on dapr endpoints. With 0.6.0 release, there will be a separate endpoint for app->dapr  which will not use tls. This was the original issue: https://github.com/dapr/dotnet-sdk/issues/261
- Also exposes option to provide GrpcChannelOptions to allow more customizations by users


amanbha:
cc @cmendible

Issue #302, opened 2020-05-06
InvokeMethodAsync API is too permissive for DAPR 0.7


mareklinka:
I just upgraded my small DAPR playground project from 0.6 to 0.7. After that, my remote invocation calls stopped working with the following exception:
```
Grpc.Core.RpcException: Status(StatusCode=InvalidArgument, Detail="missing HTTP extension field")
   at Dapr.Client.DaprClientGrpc.MakeGrpcCallHandleError[TResponse](Func`2 callFunc, CancellationToken cancellationToken)
   at Dapr.Client.DaprClientGrpc.MakeInvokeRequestAsync(String appId, String methodName, Any data, HTTPExtension httpExtension, CancellationToken cancellationToken)
   at Dapr.Client.DaprClientGrpc.InvokeMethodAsync[TRequest,TResponse](String appId, String methodName, TRequest data, HTTPExtension httpExtension, CancellationToken cancellationToken)
   at DotNetBa.Dapr.Main.Controllers.UserController.Login(LoginModel model, DaprClient dapr, CancellationToken cancellationToken) in C:\Source\Personal\dotnetba-dapr\src\DotNetBa.Dapr.Main\Controllers\UserController.cs:line 44
   at lambda_method(Closure , Object )
   at Microsoft.Extensions.Internal.ObjectMethodExecutorAwaitable.Awaiter.GetResult()
   at Microsoft.AspNetCore.Mvc.Infrastructure.ActionMethodExecutor.TaskOfActionResultExecutor.Execute(IActionResultTypeMapper mapper, ObjectMethodExecutor executor, Object controller, Object[] arguments)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.<InvokeActionMethodAsync>g__Awaited|12_0(ControllerActionInvoker invoker, ValueTask`1 actionResultValueTask)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.<InvokeNextActionFilterAsync>g__Awaited|10_0(ControllerActionInvoker invoker, Task lastTask, State next, Scope scope, Object state, Boolean isCompleted)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.Rethrow(ActionExecutedContextSealed context)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.Next(State& next, Scope& scope, Object& state, Boolean& isCompleted)
[...]

PR #303, opened 2020-05-06
httpextension should use post if not specified

LMWF:
# Description

httpextension should use post if not specified.

## Issue reference

Please reference the issue this PR will close: #302

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -191,15 +191,12 @@ private async Task MakePublishRequest<TContent>(string topicName, TContent conte
             Http.HTTPExtension httpExtension,
             CancellationToken cancellationToken = default)
         {
-            Autogenerated.HTTPExtension protoHTTPExtension = null;
+            Autogenerated.HTTPExtension protoHTTPExtension = new Autogenerated.HTTPExtension();
```
amanbha, line 194: nit var for both protoHTTPExtension  and contentType

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -211,6 +208,11 @@ private async Task MakePublishRequest<TContent>(string topicName, TContent conte
 
                 contentType = httpExtension.ContentType == null ? "application/json" : httpExtension.ContentType;
             }
+            else
+            {
+                protoHTTPExtension.Verb = Autogenerated.HTTPExtension.Types.Verb.Post;
+                contentType = "application/json";
```
amanbha, line 214: const for this as its used at 2 places now.

Issue #294, opened 2020-04-28
Actor Invocation over HTTP does NOT invoke middleware


codeputer:
## Expected Behavior

DAPR allows the invocation of an Actor and it's methods over HTTP (default port 3500).  Actors are registered within a web application which listens on an assigned port (examples are 5000).  When Actors are invoked over HTTP, I don't understand why the middleware is not being invoked.

## Actual Behavior

CloudEvents and my own middleware is ignored.

## Steps to Reproduce the Problem

1) create breakpoint in middleware
2) create a post to an actor invoking a method
3) breakpoint in middle ware is never invoked 

Solution: Can we register middleware during the process of an http POST to the --app-port?

amanbha:
@codeputer Could you please point me to the repo where you have this sample code.

codeputer:
I it was in that pull request  - the only test that i didn't put in was a post to an actor as opposed to a post to an endpoint.... can you restore the PR? Or should I do it from my end?

amanbha:
If you can just point me to your repo where you have the sample code with middleware registered with aspnetcore for actors. I can clone it and try it out.

codeputer:
I believe this link is what you are looking  for?
https://github.com/dapr/dotnet-sdk/pull/289/files#r416924384 ... I also added the link in the comment to issue 294.

codeputer:
Updated the Actor project to demo the behavior in what I believe is a bug... I'll add you to the repo so that you can download and test it.... Execute the ActorClient with the DemoActor running, and the ActorClient will fail.

codeputer:
If I'm reading the code right, I'm pretty sure the Routes that are added in the UseActor() will NOT allow a CloudEvent to be routed directly to the Actor using the URL syntax of invoking the Actor from an HTTP Request.  I'm by no means an ASP.NET expert, I'm just trying to trace in the request to port 3500, and why that doesn't invoke the middle ware.

amanbha:
@codeputer Thanks for more info, I will see how configuring the routes is affecting middlewares.

codeputer:
[...]

PR #307, opened 2020-05-12
Using EndpointRouteBuilder and correcting the order of middleware chaining

amanbha:
Fixes https://github.com/dapr/dotnet-sdk/issues/294

### DIFF
--- a/src/Dapr.Actors.AspNetCore/DaprActorSetupFilter.cs
+++ b/src/Dapr.Actors.AspNetCore/DaprActorSetupFilter.cs
@@ -8,7 +8,6 @@ namespace Dapr.Actors.AspNetCore
     using System;
     using Microsoft.AspNetCore.Builder;
     using Microsoft.AspNetCore.Hosting;
-    using Microsoft.AspNetCore.Routing;
 
     internal class DaprActorSetupFilter : IStartupFilter
     {
@@ -17,17 +16,22 @@ namespace Dapr.Actors.AspNetCore
             // Adds routes for Actors interaction.
             return app =>
             {
-                var actorRouteBuilder = new RouteBuilder(app);
-                actorRouteBuilder.AddDaprConfigRoute();
-                actorRouteBuilder.AddDaprHealthzRoute();
-                actorRouteBuilder.AddActorActivationRoute();
-                actorRouteBuilder.AddActorDeactivationRoute();
-                actorRouteBuilder.AddActorMethodRoute();
-                actorRouteBuilder.AddReminderRoute();
-                actorRouteBuilder.AddTimerRoute();
+                app.UseRouting();
 
-                app.UseRouter(actorRouteBuilder.Build());
+                // This allows the middlewares to be configured correctly from Startup class.
                 next(app);
+
+                // Configure endpoints for Actors.
+                app.UseEndpoints(endpoints =>
+                {
+                    endpoints.AddDaprConfigRoute();
+                    endpoints.AddDaprHealthzRoute();
+                    endpoints.AddActorActivationRoute();
+                    endpoints.AddActorDeactivationRoute();
+                    endpoints.AddActorMethodRoute();
+                    endpoints.AddReminderRoute();
+                    endpoints.AddTimerRoute();
+                });
             };
         }
     }
--- a/src/Dapr.Actors.AspNetCore/DaprActorSetupFilter.cs
+++ b/src/Dapr.Actors.AspNetCore/DaprActorSetupFilter.cs
@@ -7,6 +7,7 @@ namespace Dapr.Actors.AspNetCore
 {
     using System;
     using Microsoft.AspNetCore.Builder;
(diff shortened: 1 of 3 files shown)


src/Dapr.Actors.AspNetCore/DaprActorSetupFilter.cs
```diff
@@ -17,17 +16,22 @@ public Action<IApplicationBuilder> Configure(Action<IApplicationBuilder> next)
...
-                actorRouteBuilder.AddReminderRoute();
-                actorRouteBuilder.AddTimerRoute();
+                app.UseRouting();
 
-                app.UseRouter(actorRouteBuilder.Build());
+                // This allows the middlewares to be configured correctly from Startup class.
                 next(app);
+
+                // Configure endpoints for Actors.
+                app.UseEndpoints(endpoints =>
```
rynowak, line 25: this and L19 is the thing I tried to warn you about 😆  - you can't put `next(app)` in between `UseRouting` and `UseEndpoints`. They have to be a pair.

src/Dapr.Actors.AspNetCore/EndpointRouteBuilderExtensions.cs
```diff
@@ -7,32 +7,33 @@ namespace Dapr.Actors.AspNetCore
...
-                await ActorRuntime.Instance.SerializeSettingsAndRegisteredTypes(context.Response.BodyWriter);                
+                await ActorRuntime.Instance.SerializeSettingsAndRegisteredTypes(context.Response.BodyWriter);
             });
         }
 
-        public static void AddDaprHealthzRoute(this IRouteBuilder routeBuilder)
+        public static IEndpointConventionBuilder AddDaprHealthzRoute(this IEndpointRouteBuilder endpoints)
         {
-            routeBuilder.MapGet("healthz", async context =>
+            return endpoints.MapGet("healthz", async context =>
```
rynowak, line 27: If you want to btw you could change this to use the built-in health checks feature. Then the health checks are configurable instead of just always returning 200.

src/Dapr.Actors.AspNetCore/DaprActorSetupFilter.cs
amanbha: Fixed.

src/Dapr.Actors.AspNetCore/EndpointRouteBuilderExtensions.cs
amanbha: Dapr runtime currently only checks for 200 OK

rynowak: it's not related to what's *returned* in the response. 

Health checks allows the healthy/unhealthy check to do something configurable in the app to make that decision. https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/health-checks?view=aspnetcore-3.1

I'm surprised that no one has asked for this yet 😆

src/Dapr.Actors.AspNetCore/DaprActorSetupFilter.cs
```diff
@@ -8,7 +8,6 @@ namespace Dapr.Actors.AspNetCore
     using System;
     using Microsoft.AspNetCore.Builder;
     using Microsoft.AspNetCore.Hosting;
-    using Microsoft.AspNetCore.Routing;
 
     internal class DaprActorSetupFilter : IStartupFilter
```
davidfowl, line 12: Why do we still need a startup filter?

amanbha: To hook everything from the IWebHostBuilder extension.

davidfowl: I mean I understand that but I'm asking why we need it at all? Is it because we want this model to treat ASP.NET Core like an implementation detail?

src/Dapr.Actors.AspNetCore/EndpointRouteBuilderExtensions.cs
amanbha: added endpoints.MapHealthChecks

src/Dapr.Actors.AspNetCore/DaprActorSetupFilter.cs
amanbha: yes

rynowak: I talked about this a bit offline with @amanbha today - right now they treat ASP.NET Core as an implementation detail, but still include `Startup` in examples.

- For someone who wants **just** actors and doesn't care about asp.net core - they should probably use the host builder and drop `Startup` from the example
- For someone who wants to use **actors and asp.net core** it makes sense for them to plug into routing directly, and for the app to have a real middleware pipeline that the user sees. 

Discussion about how to do that second one lead to this PR 😆

Issue #304, opened 2020-05-07
Update pub/sub subscription


yaron2:
We need to update the pub/sub code for subscribers to match the new subscription JSON structure.

Reference: https://github.com/dapr/dapr/issues/1518#issuecomment-624765422

PR #309, opened 2020-05-13
Returning routes for pubsub from aspnetcore integration.

amanbha:
Fixes https://github.com/dapr/dotnet-sdk/issues/304

### DIFF
--- a/src/Dapr.AspNetCore/Dapr.AspNetCore.csproj
+++ b/src/Dapr.AspNetCore/Dapr.AspNetCore.csproj
@@ -11,6 +11,10 @@
 
   <ItemGroup>
     <FrameworkReference Include="Microsoft.AspNetCore.App" />
+    <!-- Adding ref. to System.IO.Pipelines explicitly to work around build issue with .netcore sdk 3.1.102
+    https://github.com/dotnet/core/blob/master/release-notes/3.1/3.1-known-issues.md#net-core-312
+    This should be reverted when the fix will be available in future .net core sdk. -->
+    <PackageReference Include="System.IO.Pipelines" Version="4.7.0" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/Dapr.AspNetCore/DaprEndpointRouteBuilderExtensions.cs
+++ b/src/Dapr.AspNetCore/DaprEndpointRouteBuilderExtensions.cs
@@ -12,6 +12,7 @@ namespace Microsoft.AspNetCore.Builder
     using System.Threading.Tasks;
     using Dapr;
     using Microsoft.AspNetCore.Routing;
+    using Microsoft.AspNetCore.Routing.Patterns;
     using Microsoft.Extensions.DependencyInjection;
 
     /// <summary>
@@ -40,7 +41,7 @@ namespace Microsoft.AspNetCore.Builder
                     .Where(e => e.Metadata.GetMetadata<TopicAttribute>()?.Name != null)   // only endpoints which have  TopicAttribute with not null Name.
                     .Where(e => e.RoutePattern.Parameters.Count == 0) // only endpoints which don't have parameters.
                     .Distinct()
-                    .Select(e => (e.Metadata.GetMetadata<TopicAttribute>().Name, e.RoutePattern.RawText));
+                    .Select(e => (e.Metadata.GetMetadata<TopicAttribute>().Name, e.RoutePattern));
 
                 context.Response.ContentType = "application/json";
                 using Utf8JsonWriter writer = new Utf8JsonWriter(context.Response.BodyWriter);
@@ -50,7 +51,14 @@ namespace Microsoft.AspNetCore.Builder
                 {
                     writer.WriteStartObject();
                     writer.WriteString("topic", entry.Name);
-                    writer.WriteString("route", entry.RawText);
+
+                    var route = string.Join("/",
+                        entry.RoutePattern.PathSegments
+                        .Select(segment => string.Concat(segment.Parts.Cast<RoutePatternLiteralPart>()
+                        .Select(part => part.Content))));
+
+                    writer.WriteString("route", route);
+                    Console.WriteLine(route);
                     writer.WriteEndObject();
                 }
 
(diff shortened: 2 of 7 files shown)


src/Dapr.AspNetCore/Dapr.AspNetCore.csproj
```diff
@@ -11,6 +11,10 @@
 
   <ItemGroup>
     <FrameworkReference Include="Microsoft.AspNetCore.App" />
+    <!-- Adding ref. to System.IO.Pipelines explicitly to work around build issue with .netcore sdk 3.1.102
+    https://github.com/dotnet/core/blob/master/release-notes/3.1/3.1-known-issues.md#net-core-312
+    This should be reverted when the fix will be available in future .net core sdk. -->
+    <PackageReference Include="System.IO.Pipelines" Version="4.7.0" />
```
rynowak, line 17: Yeah, this bug is pretty unfortunate. Hopefully it didn't take too long to figure out.

src/Dapr.AspNetCore/DaprEndpointRouteBuilderExtensions.cs
```diff
@@ -33,14 +36,27 @@ public static IEndpointConventionBuilder MapSubscribeHandler(this IEndpointRoute
...
                 var dataSource = context.RequestServices.GetRequiredService<EndpointDataSource>();
                 var entries = dataSource.Endpoints
-                    .Select(e => e.Metadata.GetMetadata<TopicAttribute>()?.Name)
-                    .Where(n => n != null)
+                    .OfType<RouteEndpoint>()
+                    .Where(e => e.Metadata.GetMetadata<TopicAttribute>()?.Name != null)   // only endpoints which have  TopicAttribute with not null Name.
+                    .Where(e => e.RoutePattern.Parameters.Count == 0) // only endpoints which don't have parameters.
                     .Distinct()
-                    .ToArray();
+                    .Select(e => (e.Metadata.GetMetadata<TopicAttribute>().Name, e.RoutePattern.RawText));
```
rynowak, line 43: You might also think about reporting an error if an endpoint has a parameterful route as well. 

Right now the experience would be that it would silently fail.

src/Dapr.AspNetCore/DaprEndpointRouteBuilderExtensions.cs
```diff
@@ -33,14 +36,27 @@ public static IEndpointConventionBuilder MapSubscribeHandler(this IEndpointRoute
...
                 context.Response.ContentType = "application/json";
-                await JsonSerializer.SerializeAsync(context.Response.Body, entries, context.RequestServices.GetService<JsonSerializerOptions>());
+                using Utf8JsonWriter writer = new Utf8JsonWriter(context.Response.BodyWriter);
+                writer.WriteStartArray();
+
+                foreach (var entry in entries)
+                {
+                    writer.WriteStartObject();
+                    writer.WriteString("topic", entry.Name);
+                    writer.WriteString("route", entry.RawText);
```
rynowak, line 53: Does `route` expect a leading slash? does it tolerate one? (same for trailing). 

`RawText` will be whatever the developer typed in. You might also need to deal with the case where someone wrote `~/foo/` - which is a legacy syntax routing accepts. 

For a non-parameterized route you might want to normalize it with something like:

```C#
string.Join("/", pattern.PathSegments.Select(segment => string.Concat(segment.Parts.Cast<RoutePatternLiteralPart>().Select(part => part.Content))));
```

This will combine all of the segments using `/` as a separator but without a leading or trailing slash.

test/Dapr.AspNetCore.IntegrationTest/SubscribeEndpointTest.cs
```diff
@@ -32,14 +32,20 @@ public async Task SubscribeEndpoint_ReportsTopics()
...
+                        routes.Add(element.GetProperty("route").GetString());
                     }
 
                     topics.Should().Contain("A");
                     topics.Should().Contain("B");
                     topics.Should().Contain("register-user");
+
+                    routes.Should().Contain("B");
+                    routes.Should().Contain("/topic-a");
+                    routes.Should().Contain("register-user");
```
rynowak, line 48: I assume the test code already has a case where topic name is set to something different from the route?

amanbha: Yes, it had different routes.

src/Dapr.AspNetCore/DaprEndpointRouteBuilderExtensions.cs
amanbha: Fixing it.

src/Dapr.AspNetCore/DaprEndpointRouteBuilderExtensions.cs
amanbha: @ryanvolum  Do you mean return error when Dapr calls to get subscribed topics, or just log it as an error and return rest of the topics? How to get logger which has been configured by user, context.RequestServices.GetService<ILogger>() retruns null.

amanbha: Added a warning in this case.
I have a question, I am getting logger as
`context.RequestServices.GetService<ILogger<TopicAttribute>>()`, Is it advisable to use a new class (TopicSubscription) instead of TopicAttribute to get logger so that log is more clear.

src/Dapr.AspNetCore/DaprEndpointRouteBuilderExtensions.cs
```diff
@@ -33,14 +38,44 @@ public static IEndpointConventionBuilder MapSubscribeHandler(this IEndpointRoute
...
                     .Distinct()
-                    .ToArray();
+                    .Select(e => (e.Metadata.GetMetadata<TopicAttribute>().Name, e.RoutePattern));
 
                 context.Response.ContentType = "application/json";
-                await JsonSerializer.SerializeAsync(context.Response.Body, entries, context.RequestServices.GetService<JsonSerializerOptions>());
+                using Utf8JsonWriter writer = new Utf8JsonWriter(context.Response.BodyWriter);
+                writer.WriteStartArray();
+
+                var logger = context.RequestServices.GetService<ILogger<TopicAttribute>>();
```
rynowak, line 50: The type is only used as the category name of the logger - by convention categories look like type names. You can also resolve `ILoggerFactory` and create a logger with an arbitrary name. This seems fine, a small percentage of people will see this.

amanbha: Sounds good, changing it to ILoggerFactory

Issue #312, opened 2020-05-14
SDK changes corresponding to dapr/dapr 1038 - Fix actor activation and method invocation logic to handle app restarts


LMWF:
This corresponds to https://github.com/dapr/dapr/issues/1038 in dapr/dapr.

PR #313, opened 2020-05-14
#312 - make changes corresponding to change in dapr runtime.  The run…

LMWF:
# Description

The runtime will no longer send an activate message, so the SDK should 'activate' an actor anytime it receives an actor method call.  See proposal in https://github.com/dapr/dapr/issues/1038 for longer description.

## Issue reference

Please reference the issue this PR will close: #312

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/src/Dapr.Actors/Runtime/ActorManager.cs
+++ b/src/Dapr.Actors/Runtime/ActorManager.cs
@@ -55,7 +55,7 @@ namespace Dapr.Actors.Runtime
         internal ActorTypeInformation ActorTypeInfo => this.actorService.ActorTypeInfo;
 
         internal async Task<Tuple<string, byte[]>> DispatchWithRemotingAsync(ActorId actorId, string actorMethodName, string daprActorheader, Stream data, CancellationToken cancellationToken)
-        {
+        {          
             var actorMethodContext = ActorMethodContext.CreateForActor(actorMethodName);
 
             // Get the serialized header
@@ -104,7 +104,7 @@ namespace Dapr.Actors.Runtime
         }
 
         internal async Task DispatchWithoutRemotingAsync(ActorId actorId, string actorMethodName, Stream requestBodyStream, Stream responseBodyStream, CancellationToken cancellationToken)
-        {
+        {                       
             var actorMethodContext = ActorMethodContext.CreateForActor(actorMethodName);
 
             // Create a Func to be invoked by common method.
@@ -120,7 +120,7 @@ namespace Dapr.Actors.Runtime
                     awaitable = methodInfo.Invoke(actor, null);
                 }
                 else if (parameters.Length == 1)
-                {
+                {                    
                     // deserialize using stream.
                     var type = parameters[0].ParameterType;
                     var deserializedType = await JsonSerializer.DeserializeAsync(requestBodyStream, type);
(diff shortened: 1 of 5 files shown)


src/Dapr.Actors/Runtime/ActorManager.cs
```diff
@@ -225,7 +225,12 @@ internal async Task DeactivateActor(ActorId actorId)
...
 
         private async Task<T> DispatchInternalAsync<T>(ActorId actorId, ActorMethodContext actorMethodContext, Func<Actor, CancellationToken, Task<T>> actorFunc, CancellationToken cancellationToken)
-        {
+        {            
+            if (!this.activeActors.ContainsKey(actorId))
+            {
+                await this.ActivateActor(actorId);
+            }
+
             if (!this.activeActors.TryGetValue(actorId, out var actor))
```
amanbha, line 234: This if can be removed now as the code above will activate it for sure.

src/Dapr.Actors/Runtime/ActorManager.cs
```diff
@@ -104,7 +104,7 @@ internal ActorManager(ActorService actorService)
         }
 
         internal async Task DispatchWithoutRemotingAsync(ActorId actorId, string actorMethodName, Stream requestBodyStream, Stream responseBodyStream, CancellationToken cancellationToken)
-        {
+        {
```
amanbha, line 107: nit: extra spaces

src/Dapr.Actors/Runtime/ActorManager.cs
```diff
@@ -120,7 +120,7 @@ async Task<object> RequestFunc(Actor actor, CancellationToken ct)
                     awaitable = methodInfo.Invoke(actor, null);
                 }
                 else if (parameters.Length == 1)
-                {
+                {
```
amanbha, line 123: nit: extra spaces

src/Dapr.Actors/Runtime/ActorManager.cs
```diff
@@ -159,7 +159,7 @@ async Task<object> RequestFunc(Actor actor, CancellationToken ct)
         }
 
         internal async Task FireReminderAsync(ActorId actorId, string reminderName, Stream requestBodyStream, CancellationToken cancellationToken = default)
-        {
+        {
```
amanbha, line 162: nit: extra spaces

src/Dapr.Actors/Runtime/ActorManager.cs
LMWF: This call isn't a duplicate - it's calling the one above on line 199.

amanbha: > This call isn't a duplicate - it's calling the one above on line 199.

yes, its fetching the actor instance.  Tt will never go into into the if block. 
You can remove the following line form this if block
"// This should never happen, as "Dapr" runtime activates the actor first. if it ever it would mean a bug in "Dapr" runtime."

LMWF: Oh, it looks like the highlight is pointing at 230-33.  You're pointing at 234.  Ok.

amanbha: > Oh, it looks like the highlight is pointing at 230-33. You're pointing at 234. Ok.

yes, line 234.

amanbha:
LGTM, Thanks @LMWF . Please merge this after dapr runtime changes are merged.

Issue #314, opened 2020-05-15
Update proto files for 0.8.0 release


youngbupark:
## Describe the proposal

Update proto files for 0.8.0 release

LMWF:
Reopening to track an additional proto change for bindings.

PR #320, opened 2020-06-01
Update to sdk for changes in proto for bidirectional binding in runtime

LMWF:
# Description

Update to sdk for changes in proto for bidirectional binding in runtime.

## Issue reference

Please reference the issue this PR will close: #314

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -88,19 +119,20 @@ private async Task MakePublishRequest<TContent>(string topicName, TContent conte
             var envelope = new Autogenerated.InvokeBindingRequest()
```
amanbha, line 119: Could you please add null check and throw for name and operation
ArgumentVerifier.ThrowIfNullOrEmpty(operation, nameof(operation));.

Issue #345, opened 2020-07-21
Secret Store configuration provider doesn't/cannot work


amolenk:
## Expected Behavior

I can use the Dapr secret configuration provider to get secrets from Azure Key Vault.

## Actual Behavior

The configuration provider tries to load the secrets (configured using the the DaprSecretDescriptors) as soon as the application starts. To do this it makes a call to the Dapr sidecar. However, the sidecar does not accept any incoming connections yet because it's still blocking until the app starts listening. So now we're basically in a deadlock situation with the sidecar waiting for the app, and the app cannot continue because the sidecar doesn't respond:

```
== DAPR == time="2020-07-21T16:29:12.794225+02:00" level=info msg="application protocol: http. waiting on port 5000.  This will block until the app is listening on that port." app_id=routing instance=MacBook-Pro.localdomain scope=dapr.runtime type=log ver=0.9.0

== APP == Unhandled exception. Grpc.Core.RpcException: Status(StatusCode=Internal, Detail="Error starting gRPC call: Connection refused")

== APP ==    at Dapr.Client.DaprClientGrpc.MakeGrpcCallHandleError[TResponse](Func`2 callFunc, CancellationToken cancellationToken) in /Users/sander/Git/dapr-dotnet-sdk/src/Dapr.Client/DaprClientGrpc.cs:line 562

== APP ==    at Dapr.Client.DaprClientGrpc.GetSecretAsync(String storeName, String key, Dictionary`2 metadata, CancellationToken cancellationToken) in /Users/sander/Git/dapr-dotnet-sdk/src/Dapr.Client/DaprClientGrpc.cs:line 536

== APP ==    at Dapr.Extensions.Configuration.DaprSecretStore.DaprSecretStoreConfigurationProvider.LoadAsync()

== APP ==    at Dapr.Extensions.Configuration.DaprSecretStore.DaprSecretStoreConfigurationProvider.Load()

== APP ==    at Microsoft.Extensions.Configuration.ConfigurationRoot..ctor(IList`1 providers)

== APP ==    at Microsoft.Extensions.Configuration.ConfigurationBuilder.Build()

== APP ==    at Microsoft.Extensions.Hosting.HostBuilder.BuildAppConfiguration()

== APP ==    at Microsoft.Extensions.Hosting.HostBuilder.Build()

[...]

PR #348, opened 2020-07-21
Remove secret store extension from building

LMWF:
# Description

The secret store provider does not work.  Removing it from build.

## Issue reference

https://github.com/dapr/dotnet-sdk/issues/345

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


amanbha:
@LMWF @cmendible It would be better to remove the code entirely for Extensions and keep it in a separate branch.
Secret store configuration cannot be used with aspnet core currently due to the requirement that app must start a server before secret api is available in runtime.

LMWF:
Removed in latest commit on this pr.

amanbha:
@cmendible we are removing it the ConfigurationExtension for secrets as its not usable currently due to the need for app webserver to be up and running before secret call is made, please see this issue https://github.com/dapr/dotnet-sdk/issues/345

Please let us know if you have any concerns with this, we can add it back once we have a usable behavior with this.

cmendible:
@amanbha since it does not work out of the box I agree that it shouldn't be released. But we definitely must look forward to a change in dapr to allow this behavior which is what many of us, in the .NET ecosystem, would expect.

Did a branch was created to keep the code separated but available?

amanbha:
> @amanbha since it does not work out of the box I agree that it shouldn't be released. But we definitely must look forward to a change in dapr to allow this behavior which is what many of us, in the .NET ecosystem, would expect.
> 
> Did a branch was created to keep the code separated but available?

Thanks for confirming @cmendible the code is on our release branches, release-0.9.0, release-0.8.0

PR #358, opened 2020-07-31
Use protobuf packaging for the Any type

### LINKED ISSUE #268: DaprClient via gRPC: problem with object serialization
## Expected Behavior

InvokeMethodAsync returns deserialized HelloReply object

## Actual Behavior

` Unhandled exception. System.Text.Json.JsonException: '0x13' is an invalid start of a value. Path: $ | LineNumber: 1 | BytePositionInLine: 0.`

1. Create gRPC service with ASP.NET Core 3.1 (in VS 2019 it will be created with Greeter service sample), generate client classes from daprclient.proto (Service only). It also require darp.proto classes generation
2. Create implementation of DaprClientBase where on OnInvoke you receive HelloRequest and return HelloResponce objects.
2. Create ASP.NET Core console application, generate client classes from daprclient.proto (Client only), generate client classes from greet.proto (Client only)
3. Add reference on Dapr.Client package, create dapr client and call InvokeMethodAsync against gRPC service. Something like
`var msg = new HelloRequest() {Name = "Gennadii"};  
 var anyReply = await client.InvokeMethodAsync<HelloRequest, HelloReply>("testGrpcDapr", "SayHello", msg);`
4. Start both apps via Dapr with grpc protocol (I used standalone mode).

Test project attached. (there is a workaround there which is commented out there)
[TestGrpcWithDapr.zip](https://github.com/dapr/dotnet-sdk/files/4393549/TestGrpcWithDapr.zip)

akkie:
# Description

This pull request implements two converters which help with the type conversion between arbitrary types and the protobuf `Any` type. The C# protobuf library provides a mechanism to pack or unpack protobuf messages to `Any`. It provides the methods `Any.Pack` and `Any.Unpack` to serialize/deserialize messages based on `Google.Protobuf.IMessage`. For types that are not based on `Google.Protobuf.IMessage`, the existing JSON serialization/deserialization will be used.

I've also cleaned the existing codebase a little bit.

## Issue reference

Fixes #268

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -460,7 +431,7 @@ public override async ValueTask<(TValue value, string etag)> GetStateAndETagAsyn
         public override async ValueTask<bool> TryDeleteStateAsync(
             string storeName,
             string key,
-            string etag = default,
```
amanbha, line 463: eTag should be optional

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -479,14 +450,14 @@ public override async ValueTask<(TValue value, string etag)> GetStateAndETagAsyn
             return false;
         }
 
-        private async ValueTask MakeDeleteStateCallsync(
+        private async ValueTask MakeDeleteStateCallSync(
```
amanbha, line 453: rename to MakeDeleteStateCallAsync

amanbha:
@akkie thanks for the contribution

src/Dapr.Client/DaprClientGrpc.cs
akkie: It wasn't clear to me when the etag is needed, because the `TrySaveStateAsync` method does defined it as needed and the documentation for `TryDeleteStateAsync` says "Tries to delete the the state associated with the provided key using the etag from the Dapr state." Also the `DaprClient.TryDeleteStateAsync` method defines a default for the etag param.

amanbha: Yes, it should be required for TryDelete.

akkie:
Fixed 👍

PR #359, opened 2020-08-01
Adding support for DAPR_API_TOKEN

### LINKED ISSUE #343: Add support for DAPR_API_TOKEN injection into calls
## Describe the proposal
Add support to inject the DAPR_API_TOKEN from the environment automatically for all calls in the client.
Authentication support for DAPR APIs as seen [here](https://github.com/dapr/docs/tree/master/howto/enable-dapr-api-token-based-authentication)

amanbha:
Fixes #343


src/Dapr.Actors/DaprHttpInteractor.cs
```diff
@@ -478,6 +478,14 @@ HttpRequestMessage FinalRequestFunc()
             {
                 // Get the request using the Func as same request cannot be resent when retries are implemented.
                 var request = requestFunc.Invoke();
+
+                // add token for dapr api token based authentication
+                var daprApiToken = Environment.GetEnvironmentVariable("DAPR_API_TOKEN");
```
LMWF, line 483: Could this value be cached?

amanbha: Fetching it from env every time allows for it to be used on platforms where an env could be refreshed without a process restart.

Issue #368, opened 2020-08-11
Null reference error on BindingSource check in StateEntryModelBinderProvider


altinoren:
Hi,

The following code produces a null reference error when BindingSource is null:
https://github.com/dapr/dotnet-sdk/blob/3b5065f283792e0200524259c1cf347b5483f811/src/Dapr.AspNetCore/StateEntryModelBinderProvider.cs#L32

Should check nulls, e.g:
``` csharp
if (context.BindingInfo.BindingSource?.Id == "state")
```

Thanks.

---
Here's the code I test it on. Dapr is used in the project but not in this controller:

``` csharp
    [ApiController]
    [Route("[controller]")]
    [Produces("application/json")]
    public class TestController : ControllerBase
    {
        private readonly ILogger<TestController> _logger;

        public TestController(ILogger<TestController> logger)
        {
            _logger = logger;
        }

        [HttpGet]
        public ActionResult Get([FromQuery]AttributedVM data)
        {
            return Ok();
        }
    }

    public class AttributedVM
    {
        [Required]
        public string Property1 { get; set; }
        [StringLength(4, MinimumLength = 0)]
        public string Property2 { get; set; }
    }
```
Here's how I register Dapr:
``` csharp
            services.AddControllers()
                .AddFluentValidation(fv => fv.RegisterValidatorsFromAssemblyContaining<Startup>())
                .AddDapr()
                .AddJsonOptions(opt =>
                {
                    opt.JsonSerializerOptions.Converters.Add(new JsonStringEnumConverter());
                });
```

amanbha:
@altinoren Thanks for reporting this, would you like to contribute the fix for it?

altinoren:
OK. I'll check it during the weekend.

amanbha:
> OK. I'll check it during the weekend.

Great, thanks!

PR #371, opened 2020-08-16
Fixing null reference exception when no model binder provided

altinoren:
# Description

Even when there's no Dapr attribute used on an Action, all model binders are checked by ASP.NET from the top against a parameter when needed. StateEntryModelBinderProvider does not handle a null reference when called in this context. This PR fixes that.

## Issue reference

#368 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


rynowak:
LGTM 👍

PR #372, opened 2020-08-16
Add bulk get to client

### LINKED ISSUE #369: Add Bulk Get API
With the addition of the Bulk Get API to the State Building Block, this also needs to be added to the different SDKs.

See API reference here: https://github.com/dapr/docs/blob/master/reference/api/state_api.md#get-bulk-state

yaron2:
Closes #369


src/Dapr.Client/DaprClient.cs
```diff
@@ -143,6 +143,16 @@ public abstract class DaprClient
...
+        /// <summary>
+        /// Gets a list of values associated with the <paramref name="keys" /> from the Dapr state store.
+        /// </summary>
+        /// <param name="storeName">The name of state store to read from.</param>
+        /// <param name="keys">The list of keys to get values for.</param>
+        /// <param name="parallelism">The number of concurrent get operations the Dapr runtime will issue to the state store.</param>
+        /// <param name="cancellationToken">A <see cref="CancellationToken" /> that can be used to cancel the operation.</param>
+        /// <returns>A <see cref="ValueTask{GetBulkItem}" /> that will return the list of values when the operation has completed.</returns>
+        
+        public abstract ValueTask<IEnumerable<GetBulkItem>> GetBulkStateAsync(string storeName, IEnumerable<string> keys, int parallelism = -1, CancellationToken cancellationToken = default);
```
rynowak, line 155: suggestion: use `int?` instead of a random sentinel value

rynowak: suggestion: return `ValueTask<IReadOnlyList<GetBulkItem>>` - this will allow getting the count, indexing, and iterating without an allocation.

Since the results don't truly stream from the server there's no advantage to `IEnumerable<>`

src/Dapr.Client/DaprClient.cs
```diff
@@ -143,6 +143,16 @@ public abstract class DaprClient
...
         public abstract ValueTask<TValue> GetStateAsync<TValue>(string storeName, string key, ConsistencyMode? consistencyMode = default, CancellationToken cancellationToken = default);
 
+        /// <summary>
+        /// Gets a list of values associated with the <paramref name="keys" /> from the Dapr state store.
+        /// </summary>
+        /// <param name="storeName">The name of state store to read from.</param>
+        /// <param name="keys">The list of keys to get values for.</param>
+        /// <param name="parallelism">The number of concurrent get operations the Dapr runtime will issue to the state store.</param>
+        /// <param name="cancellationToken">A <see cref="CancellationToken" /> that can be used to cancel the operation.</param>
+        /// <returns>A <see cref="ValueTask{GetBulkItem}" /> that will return the list of values when the operation has completed.</returns>
```
rynowak, line 153:
```suggestion
        /// <returns>A <see cref="ValueTask{IEnumerable{GetBulkItem}}" /> that will return the list of values when the operation has completed.</returns>
```

src/Dapr.Client/BulkGetItem.cs
```diff
@@ -0,0 +1,44 @@
...
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+// ------------------------------------------------------------
+
+namespace Dapr
+{
+    /// <summary>
+    /// Represents a state object returned from a bulk get state operation.
+    /// </summary>
+    public sealed class GetBulkItem
```
rynowak, line 11: Can this be a readonly struct instead? this is an extra allocation per item that seems like it can be avoided.

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -202,6 +202,36 @@ private static T ConvertFromInvokeBindingResponse<T>(Autogenerated.InvokeBinding
...
+                throw new ArgumentException("keys do not contain any elements");
+
+             var getBulkStateEnvelope = new Autogenerated.GetBulkStateRequest()
+            {
+                StoreName = storeName,
+                Parallelism = parallelism
+            };
+            foreach (var key in keys)
+            {
+                getBulkStateEnvelope.Keys.Add(key);
```
rynowak, line 218: This is generated code so I can't really see it - does the type of `Keys` have an `AddRange` method? It's not just cleaner code - `AddRange` is a lot more efficient in common cases

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -202,6 +202,36 @@ private static T ConvertFromInvokeBindingResponse<T>(Autogenerated.InvokeBinding
             return response.Data.Value.IsEmpty ? default : TypeConverters.FromAny<TResponse>(response.Data, this.jsonSerializerOptions);
         }
 
+        public override async ValueTask<IEnumerable<GetBulkItem>> GetBulkStateAsync(string storeName, IEnumerable<string> keys, int parallelism = -1, CancellationToken cancellationToken = default)
+        {
+            ArgumentVerifier.ThrowIfNullOrEmpty(storeName, nameof(storeName));
+            if (keys.Count() == 0)
```
rynowak, line 208: It's generally not a great idea to walk an IEnumerable twice (here and L216), because you don't know what the code that generated it does 😆 . Do we have any idea how big we expect these lists of keys be to be? 

Some other options:
- Accept a data type that supports random access instead of IEnumerable
- Buffer the items here
- add the items to `envelope.Keys` and then check the count

src/Dapr.Client/DaprClient.cs
yaron2: This throws a compile error: `Type parameter declaration must be an identifier not a type`

src/Dapr.Client/BulkGetItem.cs
yaron2: Change

src/Dapr.Client/DaprClientGrpc.cs
yaron2: Uses `AddRange` now

src/Dapr.Client/DaprClientGrpc.cs
yaron2: Changed to `IList<>`

src/Dapr.Client/DaprClient.cs
yaron2: > suggestion: return `ValueTask<IReadOnlyList<GetBulkItem>>` - this will allow getting the count, indexing, and iterating without an allocation.
> 
> Since the results don't truly stream from the server there's no advantage to `IEnumerable<>`

Changed to `IReadOnlyList<>`

yaron2: > suggestion: use `int?` instead of a random sentinel value

This isn't a random value and the param should be optional. Updated the comment.

src/Dapr.Client/DaprClient.cs
rynowak: hmmm. something is wrong tho - the return type *isn't* `ValueTask{GetBulkItem}`. it should reflect `ValueTask{IReadOnlyList{GetBulkItem}}`

yaron2: I changed it to ValueTask{IReadOnlyList}, it seems like the compiler doesn't support more than one level of generics for cref.

src/Dapr.Client/DaprClient.cs
rynowak: Yeah I get it - it's a little bit non-idiomatic to see sentinel values as part of an API when we have a whole language feature for that in C#. 

There's a very subtle way in which using `null` for "unspecified" is better than using `-1` (which I assume is what Dapr understands in the gRPC interface).  Default argument values get compiled **into** consumers of the library. So if you ever wanted to change the default here, it would not take effect for someone until they recompiled. 

Feel free to ignore if this seems unimportant.

src/Dapr.Client/DaprClient.cs
```diff
@@ -143,6 +143,16 @@ public abstract class DaprClient
...
+        /// <summary>
+        /// Gets a list of values associated with the <paramref name="keys" /> from the Dapr state store.
+        /// </summary>
+        /// <param name="storeName">The name of state store to read from.</param>
+        /// <param name="keys">The list of keys to get values for.</param>
+        /// <param name="parallelism">The number of concurrent get operations the Dapr runtime will issue to the state store. a value equal to or smaller than 0 means max parallelism.</param>
+        /// <param name="cancellationToken">A <see cref="CancellationToken" /> that can be used to cancel the operation.</param>
+        /// <returns>A <see cref="ValueTask{IReadOnlyList}" /> that will return the list of values when the operation has completed.</returns>
+        
+        public abstract ValueTask<IReadOnlyList<GetBulkItem>> GetBulkStateAsync(string storeName, IList<string> keys, int parallelism = -1, CancellationToken cancellationToken = default);
```
rynowak, line 155: Now this should be `IReadOnlyList<>` instead of `IList<>` 😆 - that makes it explicit that we're not going to mess with your parameter.

yaron2: Done

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -202,6 +202,36 @@ private static T ConvertFromInvokeBindingResponse<T>(Autogenerated.InvokeBinding
...
+
+            var response = await this.MakeGrpcCallHandleError(
+                options => client.GetBulkStateAsync(getBulkStateEnvelope, options),
+                cancellationToken);
+
+            var bulkResponse  = new List<GetBulkItem>();
+
+            foreach (var item in response.Items)
+            {
+                bulkResponse.Add(new GetBulkItem(item.Key, item.Data.ToStringUtf8(), item.Etag));
```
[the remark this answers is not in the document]
yaron2: Yeah, I returned it as string as it looks to me like in C# it's a more convenient way to transform values from and to a specific type.
If you look at JSON conversion, for example, it takes strings and not byte[]..

Same thing with conversions to int, bools, and in general `Parse()` methods.

src/Dapr.Client/BulkGetItem.cs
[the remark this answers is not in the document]
yaron2: Done

src/Dapr.Client/BulkGetItem.cs
```diff
@@ -0,0 +1,44 @@
...
+
+         /// <summary>
+        /// Gets the state key.
+        /// </summary>
+        public string Key { get; }
+
+        /// <summary>
+        /// Gets the value.
+        /// </summary>
+        public string Value { get; }
```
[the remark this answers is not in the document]
yaron2: See comment above

Issue #362, opened 2020-08-06
Allow custom serializers for System.Text.Json (for better F# support) 


LukePammant:
## Customizable JSON Serialization

 I am using F# it it appears to not play well with the new System.Text.Json (de)serialization. It would be great to be given an option to set the serializer or deserializer for the Dapr package for communication to the state store and between actors. In the actor runtime configuration maybe something like `actorRuntime.UseNewtonSoftJson()`.

LukePammant:
Is this something that you guys would be interested in supporting or is it a non-starter for performance reasons? I wouldn't mind taking a stab at it but don't want to waste time on it if its not something that is wanted.

amanbha:
@LukePammant We would like to continue to use System.Text.Json for json deserializtion/serialziation so Actor runtime will not take dependency on Newtonsoft.
However I think support for other serializers can be added for state, support can be added to plugin serializer for state with interface like:
```
interface IActorStateSerialzier
    {
        T Deserialize<T>(byte[] buffer);
        internal byte[] Serialize<T>(Type stateType, T state)
    }
```

Similarly it can be extended for calls to Actors, It would be great if you want to contribute the changes for it.

amanbha:
@LukePammant We would like to continue to use System.Text.Json for json deserializtion/serialziation so Actor runtime will not take dependency on Newtonsoft. I would like to understand the issues with System.Text.Json, could you please share the issue and if its a bug in System.Text.Json

However if there is enough community ask for custom serializers, to extend, I think support for other serializers can be added for state, support can be added to plugin serializer for state with interface like:
```
interface IActorStateSerialzier
    {
        T Deserialize<T>(byte[] buffer);
        internal byte[] Serialize<T>(Type stateType, T state)
    }
```

Similarly it can be extended for calls to Actors,

LukePammant:
Understandable @amanbha . One example would be that System.Text.Json doesn't support deserialization of simple discriminated unions which are used quite extensively in F#. 

For example

```
    type ResourceType =
        | Metal
        | Gas

    let test = 
        let discriminatedUnion = Metal

        let json = Newtonsoft.Json.JsonConvert.SerializeObject(discriminatedUnion)
        let obj1 = Newtonsoft.Json.JsonConvert.DeserializeObject<ResourceType>(json) // This works fine

        let json2 = System.Text.Json.JsonSerializer.Serialize(discriminatedUnion)
[...]

PR #379, opened 2020-08-25
Allow customizations of JsonSerializerOptions

LukePammant:
# Description 

- Added ability to customize serialization options for client-side Actor Proxies via static property on ActorProxy class
- Added ability to customize serialization options for Actor Runtime via ConfigureJsonSerializationOptions in the IWebHostBuilder.UseActors extension

## Issue reference

[https://github.com/dapr/dotnet-sdk/issues/362](https://github.com/dapr/dotnet-sdk/issues/362)

## Checklist

* [ x ] Code compiles correctly
* [ x ] Created/updated tests
* [ x ] Extended the documentation


samples/Actor/ActorClient/Program.cs
```diff
@@ -29,6 +29,9 @@ public static async Task Main(string[] args)
                 PropertyB = "ValueB",
             };
 
+            // Register custom serializers for the ActorProxy
+            //ActorProxy.JsonSerializerOptions.Converters.Add(new MyJsonConverter());
+
```
vinayada1, line 34: Remove unused code

samples/Actor/DemoActor/Program.cs
```diff
@@ -31,9 +31,13 @@ public static void Main(string[] args)
...
-            .UseActors(actorRuntime =>
-            {
-                actorRuntime.RegisterActor<DemoActor>();
-            });
+                .UseActors(actorRuntime =>
+                {
+                    actorRuntime.RegisterActor<DemoActor>();
+
+                    // Register custom serializers
+                    // actorRuntime.ConfigureJsonSerializationOptions(options =>
```
vinayada1, line 39: Remove ununsed code

src/Dapr.Actors/Runtime/ActorRuntime.cs
```diff
@@ -29,6 +29,12 @@ public class ActorRuntime
 
         private ActorSettings actorSettings;
 
+        internal static JsonSerializerOptions JsonSerializerOptions { get; } = new JsonSerializerOptions()
+        {
+            PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
+            PropertyNameCaseInsensitive = true,
```
vinayada1, line 35: Do we want to make this case insensitive?

src/Dapr.Actors/Runtime/ActorRuntime.cs
```diff
@@ -78,6 +84,15 @@ public void ConfigureActorSettings(Action<ActorSettings> actorSettingsDelegate)
             actorSettingsDelegate.Invoke(this.actorSettings);
         }
 
+        /// <summary>
+        /// Allows configuration of this app's actor <see cref="JsonSerializerOptions"/> used for message and state (de)serialization
+        /// </summary>
+        /// <param name="jsonSerializerOptionsDelegate">A delegate to edit the default JsonSerializerOptions object.</param>
+        public void ConfigureJsonSerializationOptions(Action<JsonSerializerOptions> jsonSerializerOptionsDelegate)
```
vinayada1, line 91: This method is not being used. The code is commented

src/Dapr.AspNetCore/DaprApplicationBuilderExtensions.cs
```diff
@@ -18,16 +19,17 @@ public static class DaprApplicationBuilderExtensions
         /// requests that use the cloud events structured format, allowing the event payload to be read directly.
         /// </summary>
         /// <param name="builder">An <see cref="IApplicationBuilder" />.</param>
+        /// <param name="jsonSerializerOptions">The <see cref="JsonSerializerOptions" /> to use when deserializing cloud event structured events.</param>
         /// <returns>The <see cref="IApplicationBuilder" />.</returns>
-        public static IApplicationBuilder UseCloudEvents(this IApplicationBuilder builder)
+        public static IApplicationBuilder UseCloudEvents(this IApplicationBuilder builder, JsonSerializerOptions jsonSerializerOptions = null)
```
vinayada1, line 24: Need to make changes to use the new parameter?

src/Dapr.Actors/Client/ActorProxy.cs
```diff
@@ -21,6 +21,12 @@ namespace Dapr.Actors.Client
     public class ActorProxy : IActorProxy
     {
         internal static readonly ActorProxyFactory DefaultProxyFactory = new ActorProxyFactory();
+
+        /// <summary>
+        /// The <see cref="System.Text.Json.JsonSerializerOptions"/> used for outbound actor message serialization
+        /// </summary>
+        public static readonly JsonSerializerOptions JsonSerializerOptions = new JsonSerializerOptions();
```
rynowak, line 28: We should find another way to accomplish this. Using a static for configuration will result in a lot of problems down the road - it's not testable and it's easy for libraries to use in a way that doesn't compose.

src/Dapr.Actors/Runtime/ActorRuntime.cs
```diff
@@ -29,6 +29,12 @@ public class ActorRuntime
 
         private ActorSettings actorSettings;
 
+        internal static JsonSerializerOptions JsonSerializerOptions { get; } = new JsonSerializerOptions()
```
rynowak, line 32: This should be private and non-static - or public and non-static if it needs to exposed for other consumers. Generally if the user configures it, then the user should be able to access it (public).

src/Dapr.Actors/Runtime/ActorRuntime.cs
rynowak: If we expect actors to only talk to other actors then case-sensitive is the right default. 

We chose case-insensitive in ASP.NET because there's a whole galaxy of other clients out there (especially in .NET) that do `PascalCase` by default - and the default that a lot of people are used to is case-insensitive. It was a pragmatic decision on our part because diagnosing a deserialization problem due to casing can be really hard for users.

src/Dapr.AspNetCore/DaprApplicationBuilderExtensions.cs
rynowak, line 24: I don't think the cloud events middleware should have configurable settings. IIRC it only interacts with the *envelope* of the event, which is part of the specification.

samples/Actor/ActorClient/Program.cs
LukePammant: Hey @vinayada1 - I  added those comments intentionally to act as guide for people that want to add JsonSerialization options. If I un-commented it I would need to implement a custom serializer so the code compiles. Maybe I can change the example to just set camel case to true?

src/Dapr.AspNetCore/DaprApplicationBuilderExtensions.cs
LukePammant: I have removed the option per @rynowak's comment

LukePammant:
@rynowak & @vinayada1 - I have updated the code to the latest master and removed the static instances of the serialization. When you get a chance could you review again please?

vinayada1 (dismissed):
Thanks for your contribution

src/Dapr.Actors/Client/ActorProxy.cs
```diff
@@ -41,6 +43,14 @@ protected internal ActorProxy()
...
         internal IActorMessageBodyFactory ActorMessageBodyFactory { get; set; }
 
+        /// <summary>
+        /// The <see cref="IActorProxyFactory"/> to use when creating new <see cref="ActorProxy"/> instances
+        /// </summary>
+        public static void ConfigureDefaultFactory(Action<ActorProxyFactory> configureDelegate)
+        {
+            configureDelegate.Invoke(ActorProxyFactory);
+        }
+
```
rynowak, line 53: Sorry for just catching up on this, I haven't really played much with the actors code. 

We should not introducing statics that mutate configuration like this - having a static method to configure something is the same as a static property. It's not testable and it doesn't compose with libraries.

src/Dapr.Actors/Client/ActorProxy.cs
```diff
@@ -76,7 +84,7 @@ public static TActorInterface Create<TActorInterface>(ActorId actorId, string ac
         public static object Create(ActorId actorId, Type actorInterfaceType, string actorType)
         {
             _ = actorInterfaceType as IActor ?? throw new ArgumentException("The interface must implement IActor.", nameof(actorInterfaceType));
-            return DefaultProxyFactory.CreateActorProxy(actorId, actorInterfaceType, actorType);
+            return ActorProxyFactory.CreateActorProxy(actorId, actorInterfaceType, actorType);
```
rynowak, line 87: Do we have a test that calls this method? Line 86 will always fail....

src/Dapr.Actors/Client/ActorProxyFactory.cs
```diff
@@ -6,23 +6,29 @@
...
     /// </summary>
-    internal class ActorProxyFactory : IActorProxyFactory
+    public class ActorProxyFactory : IActorProxyFactory
     {
         private readonly IDaprInteractor daprInteractor;
 
+        /// <summary>
+        /// 
+        /// </summary>
+        public readonly JsonSerializerOptions JsonSerializerOptions;
```
rynowak, line 23: This will always be null right? Nothing initializes it and it's readonly.

src/Dapr.Actors/Runtime/DaprStateProvider.cs
```diff
@@ -19,10 +19,12 @@ namespace Dapr.Actors.Runtime
     internal class DaprStateProvider
     {
         private readonly IActorStateSerializer actorStateSerializer;
+        private readonly JsonSerializerOptions jsonSerializerOptions;
 
-        public DaprStateProvider(IActorStateSerializer actorStateSerializer = null)
+        public DaprStateProvider(IActorStateSerializer actorStateSerializer = null, JsonSerializerOptions jsonSerializerOptions = null)
```
rynowak, line 24: It seems like this should really be two separate constructors right?

If you specify a serializer, then the json options have no effect. The bulk of the logic here could be in an implementation of `IActorStateSerializer`- with a constructor here that accepts the options and instantiates it.

src/Dapr.Actors/Client/ActorProxy.cs
LukePammant: @rynowak I agree - the only problem is that the ActorProxy uses the static method `Create` which calls the static property `ActorProxyFactory`. The only way I can think of getting the custom JSON settings to _all_ instantiated `ActorProxy`s would be to change the static `ActorProxyFactory` field a property w/ a setter on it to allow the user to set their own Actor Factory with the custom settings when the service starts up. Either way it's _still_ using a static. Short of rewriting the way that an `ActorProxy` is created I'm not sure there is a good way to get away from the statics.

If you think we shouldn't do this for _all_  ActorProxy instances I could just add an overloaded method for `Create` that takes `JsonSerializerOptions`. That would remove the need to configure the static `ActorProxyFactory`

src/Dapr.Actors/Client/ActorProxy.cs
LukePammant: You are correct - I just noticed this issue too. Once we decide how to do the configuration I'll make sure to get tests made for the new methods.

src/Dapr.Actors/Client/ActorProxyFactory.cs
LukePammant: You are correct - I just noticed this too. Once we decide on how we want to handle getting the `JsonSerializerOptions` in there I'll be sure to get unit tests created for any modifications.

src/Dapr.Actors/Client/ActorProxy.cs
rynowak: @LukePammant - sorry for the delay, I was travelling

Let's try to nail down an API proposal then. There's a few ideas that would work here (can be tweaked):

(sorry my samples will be C# not F# - it's out of ignorance on my part 😎)

**Option 1: create and customize the factory**

```C#
var factory = new ProxyFactory() { SerializerOptions = ... }; // customize
var proxy = factory.Create<IMyActorType>(id: ...);
```

Old API still works and uses a default factory with default options:

```C#
var = ActorProxy.Create<IMyActorType>(id: ...)
```

If you really like statics just make your own.

```C#
class MyFactory
{
    public static readonly ActorProxyFactory Instance = new ActorProxyFactory()
    {
        // customize
    }
}

// elsewhere 
var proxy = MyFactory.Instance.Create<IMyActorType>(id: ...);
```

**Option 2: additional argument**

```C#
var options = new ActorOptions() { new JsonSerializerSettings() {  ... } }; // customize
var proxy = ActorProxy.Create<IMyActorType>(id: ..., options);
```

Common usage would probably be to end up with your static own options instance.

```C#
class MyOptions
{
    public static readonly ActorOptions Instance = new ActorOptions() 
    {
         new JsonSerializerSettings() {  ... }  // customize
     };
}

// elsewhere 
var proxy = ActorProxy.Create<IMyActorType>(id: ..., MyOptions.Instance);
```

The reason for the `ActorOptions` class to future proofing. If we add more options to creating a proxy then we don't have to change the signature. 

**Option 3: Both**

Support both an argument (optional) to create an actor and the factory as a standalone class.

-----

The grow up of this is that we should also probably add `Create` functionality to ActorService. This makes it really easy to create a proxy to another actor (from inside an actor) without having to keep track of statics - you can configure it in a single place.

samples/Actor/DemoActor/Program.cs
```diff
@@ -31,9 +32,17 @@ public static void Main(string[] args)
...
+                .UseActors(actorRuntime =>
+                {
+                    actorRuntime.RegisterActor<DemoActor>();
+
+                    // Optionally register custom serialization options for inbound Actor requests
+                    actorRuntime.ConfigureJsonSerializationOptions(options =>
+                    {
+                        options.PropertyNameCaseInsensitive = true;
+                        options.PropertyNamingPolicy = JsonNamingPolicy.CamelCase;
+                        //options.Converters.Add(new MyCustomConverter();
```
rynowak, line 44: Is there a reason why this needs to be a callback style API? It's clunky compared with just having a property.

src/Dapr.Actors/Client/ActorProxy.cs
LukePammant: This is true - looking at the git comments for this code I believe I fixed what it was intending to do w/: 
`typeof(IActor).IsAssignableFrom(actorInterfaceType)`

I'll get a unit test to hit it to verify.

src/Dapr.Actors/Client/ActorProxy.cs
LukePammant: My latest commit has these changes in it. I reworked a lot of the original code I submitted and think it's a little more inline with your thoughts. I didn't remove the existing statics in order to not break existing code.

LukePammant:
Apologies for the messiness of this PR - this is my first PR to an open source project and I'm used to rearranging my own feature branch commits. I guess I shouldn't have done a force push as it has broken where the comments were pointing to...

Regardless - I have made the requested changes. Please review and let me know if there are any changes you'd like to see.

src/Dapr.Actors/Client/ActorProxyFactory.cs
```diff
@@ -6,58 +6,56 @@
...
     using System;
+    using System.Text.Json;
     using Dapr.Actors.Builder;
     using Dapr.Actors.Communication.Client;
 
     /// <summary>
     /// Represents a factory class to create a proxy to the remote actor objects.
     /// </summary>
-    internal class ActorProxyFactory : IActorProxyFactory
+    public class ActorProxyFactory : IActorProxyFactory
```
rynowak, line 16: What's the value in making this *class* public? vs the interface - do you have a code sample of how the class would be used?

I think what's missing here is that the `JsonSerializerOptions` should be a get/set property on the class. Then you can instantiate your own instance of `ActorProxyFactory` and tweak its settings and reuse that.

If that's the intended design then I think you'll want to:

- Turn `JsonSerializerOptions` into a get/set property
- Make `ActorProxyFactory` sealed
- Change the type of `ActorProxy.DefaultProxyFactory` to be `IActorProxyFactory` so that the default implementation isn't mutable.

src/Dapr.Actors/Client/ActorProxyOptions.cs
```diff
@@ -0,0 +1,20 @@
...
+
+    /// <summary>
+    /// The class containing customizable optins for how the Actor Proxy works
+    /// </summary>
+    public class ActorProxyOptions
+    {
+        /// <summary>
+        /// The <see cref="JsonSerializerOptions"/> used for actor message serialization
+        /// </summary>
+        public JsonSerializerOptions JsonSerializerOptions { get; set; } = new JsonSerializerOptions();
```
rynowak, line 18: Should this be null by default? What will happen here is that this will be non-null by default and so when you pass in the options it will always override the settings in `ActorProxyFactory`.

src/Dapr.Actors/Client/ActorProxyFactory.cs
```diff
@@ -6,58 +6,56 @@
...
     {
         private readonly IDaprInteractor daprInteractor;
 
+        /// <summary>
+        /// The <see cref="JsonSerializerOptions"/> used for actor message serialization for remoting
+        /// </summary>
+        public readonly JsonSerializerOptions JsonSerializerOptions = new JsonSerializerOptions();
```
rynowak, line 23: We should use `JsonSerializerDefaults.Web` or at least the equivalent settings on .NET 5 on .NET 3.1.  https://github.com/dotnet/runtime/blob/1d9e50cb4735df46d3de0cee5791e97295eaf588/src/libraries/System.Text.Json/src/System/Text/Json/Serialization/JsonSerializerOptions.cs#L103

rynowak (changes requested):

src/Dapr.Actors/Client/ActorProxy.cs
```diff
@@ -75,7 +78,9 @@ public static TActorInterface Create<TActorInterface>(ActorId actorId, string ac
         /// <returns>Proxy to the actor object.</returns>
         public static object Create(ActorId actorId, Type actorInterfaceType, string actorType)
         {
-            _ = actorInterfaceType as IActor ?? throw new ArgumentException("The interface must implement IActor.", nameof(actorInterfaceType));
+            if(typeof(IActor).IsAssignableFrom(actorInterfaceType))
```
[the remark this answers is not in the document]
LukePammant: Done (I'm assuming you meant to say "add braces *around* the if"?

src/Dapr.Actors/Client/ActorProxyFactory.cs
LukePammant: Nice, I wasn't aware of this new default. I don't see any equivalent on .NET 3.1 - would you settle for a static class? 
``` C#
namespace Dapr.Actors.Common
{
    /// <summary>
    /// A helper used to stanardize the <see cref="JsonSerializerOptions"/> defaults
    /// </summary>
    public static class JsonSerializerDefaults
    {
        /// <summary>
        /// The <see cref="JsonSerializerOptions"/> used for actor message serialization
        /// </summary>
        public static JsonSerializerOptions Web = new JsonSerializerOptions()
            { 
                PropertyNamingPolicy = JsonNamingPolicy.CamelCase, 
                PropertyNameCaseInsensitive = true
            };
    }
}
```

Used like:
``` C#
public JsonSerializerOptions JsonSerializerOptions { get; set; } = JsonSerializerDefaults.Web;
```

The only issue is there isn't any common shared project across all these Dapr projects so I'd either have to make a new one just for this class or I'd have to duplicate it in each project.

LukePammant: I have moved the `JsonSerializerOptions`  to the `ActorProxyOptions` class which is used by the `ActorProxyFactory` to initialize the `ActorProxy`s. I have also set the default `ActorProxy`'s `JsonSerializerOptions` to the new static instance I made `JsonSerializerDefaults.Web`. This seems to simplify it a bit more than what I had originally.

LukePammant: On second thought - a mutable static instance for `JsonSerializerOptions` isn't a good idea. I've change it to a static method that returns a new `JsonSerializerOptions` instance and will leave it up to the user to reuse it.

src/Dapr.Actors/Client/ActorProxyFactory.cs
LukePammant: My thought was that we would use it in the `Startup.cs` or `Program.cs` via:
```C#
ActorProxy.DefaultProxyFactory.JsonSerializerOptions.Converters.Add(new JsonFSharpConverter())
```
I didn't change it to an interface because I wasn't sure on the original decision to have it be a class. I agree that it should be an interface so the user can just set it to which ever factory implementation they want. 

Changing it to an interface makes it so the example above won't work unless we add `JsonSerializerOptions` to the interface (which might be too prescriptive - ie. what if we want an XML Actor Factory?)

By changing it to an interface the above example becomes much more verbose:
```C#
var fSharpJsonSerializer = new JsonSerializerOptions {
                                   PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
                                   PropertyNameCaseInsensitive = true
                               };
fSharpJsonSerializer.Converters.Add(JsonFSharpConverter());

var proxyOptions = new ActorProxyOptions { JsonSerializerOptions = fSharpJsonSerializer };
var actorProxyFactory = ActorProxyFactory(proxyOptions);

ActorProxy.DefaultProxyFactory = actorProxyFactory;
```

In F#:
```F#
let fSharpJsonSerializer = JsonSerializerOptions()
fSharpJsonSerializer.PropertyNamingPolicy <- JsonNamingPolicy.CamelCase
fSharpJsonSerializer.PropertyNameCaseInsensitive <- true
fSharpJsonSerializer.Converters.Add(JsonFSharpConverter())

let proxyOptions = ActorProxyOptions()
proxyOptions.JsonSerializerOptions <- fSharpJsonSerializer
let actorProxyFactory = ActorProxyFactory(proxyOptions)

ActorProxy.DefaultProxyFactory <- actorProxyFactory
```

By making the `ActorProxy.JsonSerializerOptions` into a public property the user could simply set the serializer options on a per proxy basis via:
```C#
var proxy = ActorProxy.Create<IMyActor>(new ActorId(id), "MyActor");
proxy.JsonSerializerOptions.Converters.Add(JsonFSharpConverter());
```

src/Dapr.Actors/Client/ActorProxyOptions.cs
LukePammant: I have removed the `JsonSerializerOptions` from the `ActorProxyFactory` and added a `ActorProxyOptions` property to the `ActorProxyFactory` instead. It seems appropriate that the defaults would be set in the `ActorProxyOptions` and be overridden there if need be.

src/Dapr.Actors/Client/ActorProxyFactory.cs
LukePammant: I hid my previous comment after thinking about it a little more it seems like it would be acceptable to add an `ActorProxyOptions` property to the `IActorProxyFactory` interface. This would make it so we set the default `JsonSerializerOptions` on new `ActorProxy` instances via the `DefaultActorProxyOptions` in the factory or via the `ActorProxy.Create` method. 

So in order to add a custom serializer to all `ActorProxy`'s by default you would do:
```C#
ActorProxy.DefaultProxyFactory.DefaultActorProxyOptions.JsonSerializerOptions.Converters.Add(new JsonFSharpConverter());
```

src/Dapr.Actors/Client/ActorProxyFactory.cs
rynowak: Thanks @lukepammant - I'll take a look. Agreed, it's not good to expose a static mutable thing.

samples/Actor/DemoActor/Program.cs
```diff
@@ -31,9 +31,9 @@ public static void Main(string[] args)
...
             WebHost.CreateDefaultBuilder(args)
                 .UseStartup<Startup>()
-            .UseActors(actorRuntime =>
-            {
-                actorRuntime.RegisterActor<DemoActor>();
-            });
+                .UseActors(actorRuntime =>
+                {
+                    actorRuntime.RegisterActor<DemoActor>();
+                });
```
rynowak, line 37: You should see some conflicts for this file, the registration of actors has moved to `Startup.cs`.

src/Dapr.Actors/Client/IActorProxyFactory.cs
```diff
@@ -20,18 +22,33 @@ public interface IActorProxyFactory
...
         /// <param name="actorType">Type of actor.</param>
+        /// <param name="options">The optional <see cref="ActorProxyOptions" /> to use when creating the actor proxy</param>
         /// <returns>Actor proxy to interact with remote actor object.</returns>
-        ActorProxy Create(ActorId actorId, string actorType);
+        ActorProxy Create(ActorId actorId, string actorType, ActorProxyOptions options = null);
+
+        /// <summary>
+        /// The <see cref="ActorProxyOptions"/> used for initializing <see cref="ActorProxy"/>s
+        /// </summary>
+        ActorProxyOptions DefaultActorProxyOptions { get; set; }
```
rynowak, line 52: This should be on the implementation but not the interface, exposing this here turns `ActorProxyFactory.Default` into a mutable static.

src/Dapr.Actors/Common/JsonSerializerDefaults.cs
```diff
@@ -0,0 +1,19 @@
+using System.Text.Json;
+
+namespace Dapr.Actors.Common
```
rynowak, line 3: Let's not introduce a new namespace, it's just clutter. If its common code it's fine to set in `Dapr.Actors`, and then we don't have extra usings all over.

src/Dapr.Actors/Common/JsonSerializerDefaults.cs
```diff
@@ -0,0 +1,19 @@
+using System.Text.Json;
```
rynowak, line 1: Needs copyright header

src/Dapr.Actors/Runtime/ActorService.cs
```diff
@@ -21,15 +22,18 @@ public class ActorService : IActorService
         /// <param name="actorTypeInfo">The type information of the Actor.</param>
         /// <param name="loggerFactory">The logger factory.</param>
         /// <param name="actorFactory">The factory method to create Actor objects.</param>
+        /// <param name="jsonSerializerOptions">The <see cref="JsonSerializerOptions"/> to use when actor state persistence and message deserialization</param>
```
rynowak, line 25:
```suggestion
        /// <param name="jsonSerializerOptions">The <see cref="JsonSerializerOptions"/> to use for actor state persistence and message deserialization.</param>
```

rynowak: Where does this get used for message deserialization? I only see it used for state persistence in this PR.

src/Dapr.Actors/Client/ActorProxyFactory.cs
```diff
@@ -16,48 +16,53 @@ internal class ActorProxyFactory : IActorProxyFactory
     {
         private readonly IDaprInteractor daprInteractor;
 
+        /// <inheritdoc/>
+        public ActorProxyOptions DefaultActorProxyOptions { get; set; } = new ActorProxyOptions();
```
rynowak, line 20: Suggestion - this name could be simplified to `DefaultOptions`.

src/Dapr.Actors/Client/ActorProxyFactory.cs
```diff
@@ -16,48 +16,53 @@ internal class ActorProxyFactory : IActorProxyFactory
...
         {
-            var actorProxy = new ActorProxy();
-            var nonRemotingClient = new ActorNonRemotingClient(this.daprInteractor);
-            actorProxy.Initialize(nonRemotingClient, actorId, actorType);
+            var remotingClient = new ActorRemotingClient(this.daprInteractor);
+            var proxyGenerator = ActorCodeBuilder.GetOrCreateProxyGenerator(actorInterfaceType);
+            var actorProxy = proxyGenerator.CreateActorProxy();
+            actorProxy.Initialize(remotingClient, actorId, actorType);
+
+            actorProxy.JsonSerializerOptions ??= options.JsonSerializerOptions ?? this.DefaultActorProxyOptions?.JsonSerializerOptions;
```
rynowak, line 46: This is going to throw when the `options` are null. Needs a `?.`

rynowak, line 46: Will `actorProxy.JsonSerializerOptions` ever be non-null at this point?

rynowak: It would be better to pass this in to initialize

src/Dapr.Actors/Client/ActorProxy.cs
```diff
@@ -33,6 +38,11 @@ protected internal ActorProxy()
         {
         }
 
+        /// <summary>
+        /// The <see cref="JsonSerializerOptions"/> used for Actor messaging
+        /// </summary>
+        public JsonSerializerOptions JsonSerializerOptions { get; set; } = JsonSerializerDefaults.Web;
```
rynowak, line 44: Does this need to be public and get/set? The user already has a way to control this. It's little confusing that defaults are applied at multiple levels. Without enabling some additional scenario I think it would be better to treat this as an implementation detail of the class.

src/Dapr.Actors/Client/ActorProxyOptions.cs
```diff
@@ -0,0 +1,23 @@
...
+
+    /// <summary>
+    /// The class containing customizable options for how the Actor Proxy is initialized
+    /// </summary>
+    public class ActorProxyOptions
+    {
+        // TODO: Add actor retry settings
+
+        /// <summary>
+        /// The <see cref="JsonSerializerOptions"/> used for actor message serialization
```
rynowak, line 19:
```suggestion
        /// The <see cref="JsonSerializerOptions"/> used for actor message serialization in non-remoting invocation.
```

There are actually two different serializers at play. Using the strongly typed interface-based approach results in data contract serializer.

rynowak (changes requested):
This is looking better but it still has a mutable static that needs to go away. 

Also needs some unit tests, at least for the code path that creates actor proxies. It would be really really hard to unit test actor invocation right now, so I don't suggest you try, this is a gap we know about.

rynowak:
@LukePammant - it would be great to see this updated and try to get it merged within the next week. If you don't think that's doable one option is for me to take it over and finish. Let me know how you want to proceed, I want to make sure this is included in the next release.

LukePammant:
Awesome, thanks @rynowak! I just saw your change to master to move the initialization to startup.cs. The routing has been causing me massive headaches w/ trying to use the Spa middleware/default fallback route + Dapr Actors. 

I'll rebase off of the latest master and address the topics you brought up. If we can keep this tight feedback loop going I can definitely try to own it through to completion.

rynowak:
> The routing has been causing me massive headaches w/ trying to use the Spa middleware/default fallback route + Dapr Actors.

It's good to get the confirmation that folks are trying this! I ran into the same problem using actors + blazor server in the same project.

src/Dapr.Actors/Client/ActorProxyFactory.cs
LukePammant: Following your comment on passing the `JsonSerializerOptions` into the `Initialize` method I opted to pass the whole `ActorProxyOptions` instance in.

src/Dapr.Actors/Runtime/ActorService.cs
LukePammant: After rebasing to master I have removed this and moved the JsonSerializerOptions for the run time over to the `ActorRuntimeOptions` class. This is similar to the ActorClient code that has the JsonSerializerOptions in the ActorProxyOptions. I think this helps simplify where the user should be customizing the JsonSerializerOptions.

src/Dapr.Actors/Client/IActorProxyFactory.cs
LukePammant: I see what you're saying. I added it as a convenience but I can see how it can cause strange behavior. By removing it from the interface it just changes the serialization customization code from:
```C#
ActorProxy.DefaultProxyFactory.DefaultActorProxyOptions.JsonSerializerOptions.Converters.Add(new JsonFSharpConverter());
```
to either cast it and modify it:
```c#
((ActorProxyFactory)ActorProxy.DefaultProxyFactory).DefaultActorProxyOptions.JsonSerializerOptions.Converters.Add(new JsonFSharpConverter());
```
or make a new instance w/ the custom options:
```c#
var proxyOptions = new ActorProxyOptions();
proxyOptions.JsonSerializerOptions.Converters.Add(new JsonFSharpConverter());
ActorProxy.DefaultProxyFactory = new ActorProxyFactory(proxyOptions);
```

src/Dapr.Actors/Client/ActorProxy.cs
LukePammant: Currently the only way a user would be able to control the serialization options of the proxy is by setting them in the ActorProxyOptions on the DefaultProxyFactory. Once they get an instance of the ActorProxy the won't be able to change the serialization options on a per instance scale.

That being said - I added the `ActorProxyOptions` as an option parameter to the static `Create` methods to give the users the ability to set the options on a per proxy basis. I think this was the intention from the start that I had overlooked. I also made the property JsonSerializerOptions on `ActorProxy` into a private field.

LukePammant:
@rynowak - I'm done with the rebase and refactor. I have updated the code to be a little more clear about where to set the JsonSerializer options for the runtime and the proxies.

For the **_runtime_** you can set it via the `ActorRuntimeOptions.JsonSerializerOptions` via:
```c#
 services.AddActors(options =>
 {
     options.JsonSerializerOptions.AllowTrailingCommas = false;
     options.Actors.RegisterActor<DemoActor>();
 });
```


For the actor **_proxies/client_**:

- You can set it per instance `ActorProxy.Create` via:
```c#
ActorProxy.Create<IMyActor>(actorID, actorType, proxyOptionsWithCustomSerializerOptions);
```

- Or set it for all proxies by default via:
```c#
((ActorProxyFactory)ActorProxy.DefaultProxyFactory).DefaultActorProxyOptions.JsonSerializerOptions.Converters.Add(new JsonFSharpConverter());
```
- Or 
```c#
ActorProxy.DefaultProxyFactory = new ActorProxyFactory(proxyOptionsWithCustomSerializerOptions);
```

I haven't started on the unit tests yet. Once we are in agreement on the code I'll start on that. I will need a little time to wrap my head around the unit testing but I should be able to get it done :)

src/Dapr.Actors/Client/IActorProxyFactory.cs
LukePammant: For more context - I realized we were talking about the decision to make `ActorProxyFactory` public in this comment too: https://github.com/dapr/dotnet-sdk/pull/379#discussion_r550822085

LukePammant:
I added a simple set of unit tests. I didn't see a standard naming convention for the unit tests so I just copied from an arbitrary set of tests.

src/Dapr.Actors/Client/ActorProxyFactory.cs
rynowak: > So in order to add a custom serializer to all ActorProxy's by default you

This is a case I want to avoid. No mutable statics allowed.

This has been addressed by removing the options property from `IActorProxyFactory`.

src/Dapr.Actors/Client/ActorProxyFactory.cs
```diff
@@ -12,52 +12,59 @@ namespace Dapr.Actors.Client
...
             this.daprInteractor = new DaprHttpInteractor();
         }
 
-        /// <inheritdoc/>
-        public TActorInterface CreateActorProxy<TActorInterface>(ActorId actorId, string actorType)
-            where TActorInterface : IActor
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ActorProxyFactory"/> class.
+        /// </summary>
+        public ActorProxyFactory(ActorProxyOptions options) : this()
```
rynowak, line 39: Note that this constructor isn't needed anymore since the property is get/set. I'd kinda prefer a single paradigm - either a constructor and a get-only property or a no-arg constructor and a get/set property.

src/Dapr.Actors/Common/JsonSerializerDefaults.cs
```diff
@@ -0,0 +1,24 @@
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+// ------------------------------------------------------------
```
rynowak, line 4: Move this out of the Common folder

src/Dapr.Actors/Runtime/ActorHost.cs
```diff
@@ -40,11 +42,12 @@ public sealed class ActorHost
...
 
-        internal DaprStateProvider StateProvider { get; }
-
         /// <summary>
         /// Gets the LoggerFactory for actor service
         /// </summary>
         public ILoggerFactory LoggerFactory { get; }
+
+        internal DaprStateProvider StateProvider { get; }
+        internal JsonSerializerOptions JsonSerializerOptions { get; }
```
rynowak, line 51: This property can be public. Generally if you pass something into a public constructor, it's just good style to make accessible as a readonly property. You're not really abstracting anything by hiding it.

test/Dapr.Actors.AspNetCore.IntegrationTest/HostingTests.cs
```diff
@@ -80,7 +80,7 @@ public async Task MapActorsHandlers_WorksWithFallbackRoute()
             await Assert2XXStatusAsync(response);
         }
 
-        private static IHost CreateHost<TStartup>()
+        private static IHost CreateHost<TStartup>() where TStartup : class
```
rynowak, line 83: Thanks 👍

test/Dapr.Actors.Test/ActorProxyTests.cs
```diff
@@ -40,8 +41,29 @@ public void TestCreateActorProxyThatImplementsInterface_Success()
...
+        [Fact]
+        public void TestCustomSerializerOptionsOnDefaultActorProxyFactory_Success()
+        {
+            var factory = new ActorProxyFactory();
+            factory.DefaultOptions.JsonSerializerOptions = new JsonSerializerOptions();
+
+            var actorId = new ActorId("abc");
+            var proxy = ActorProxy.Create(actorId, typeof(ITestActor), "TestActor");
+
+            Assert.NotNull(proxy);
```
rynowak, line 57: Can we verify that the options actually got to the proxy? Should the serialization options be an internal property on the proxy so we can check it here?

test/Dapr.Actors.Test/ActorProxyTests.cs
```diff
@@ -40,8 +41,29 @@ public void TestCreateActorProxyThatImplementsInterface_Success()
...
+        }
+
```
rynowak, line 59: Would like to see tests for the other overloads of ActorProxy.Create

rynowak: As well as cases where the options is passed in to ActorProxy.Create

test/Dapr.Actors.Test/ActorProxyTests.cs
LukePammant: Yeah I was trying to think of how we can go about testing that. Making the options into an internal property seems like it will fit the bill so I'll go ahead with that.

rynowak (approved):
Great! Thanks @LukePammant

Issue #373, opened 2020-08-17
Add state transaction operation api 


youngbupark:
## Describe the proposal

Add state transaction operation api  - dapr/dapr#1892

PR #385, opened 2020-09-01
State Transaction API changes

vinayada1:
# Description

State Transaction operation API changes for .net sdk

#373 

Please reference the issue this PR will close: #373 
## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


src/Dapr.Client/StateTransactionRequest.cs
```diff
@@ -0,0 +1,49 @@
...
+        /// </summary>
+        public string Key { get; }
+
+        /// <summary>
+        /// Gets or sets the value locally.
+        /// </summary>
+        public TValue Value { get; set; }
+
+        /// <summary>
+        /// The Operation type.
```
vinayada1, line 45: @youngbupark - Is etag and metadata required here?

src/Dapr.Client/DaprClient.cs
```diff
@@ -228,6 +228,37 @@ public async ValueTask<StateEntry<TValue>> GetStateEntryAsync<TValue>(string sto
...
+        /// Saves the provided <paramref name="stateTransactionRequests" /> to the Dapr state
+        /// store.
+        /// </summary>
+        /// <param name="storeName">The name of the state store.</param>
+        /// <param name="stateTransactionRequests">A list of StateTransactionRequests.</param>
+        /// <param name="metadata">An key/value pair that may be consumed by the state store.  This depends on the state store used.</param>
+        /// <param name="cancellationToken">A <see cref="CancellationToken" /> that can be used to cancel the operation.</param>
+        /// <typeparam name="TValue">The data type of the value to save.</typeparam>
+        /// <returns>A <see cref="ValueTask" /> that will complete when the operation has completed.</returns>
+        public abstract Task ExecuteStateTransactionAsync<TValue>(
```
amanbha, line 241: This method will not take TValue as different states can have different types, its better to take state values as a byte[] and users will serialize the types and pass the values as byte[]

rynowak:
Is there a good description somewhere of what this is/how it works? I tried following the links but none of these me to any useful information :shrug:

LMWF:
This is the api spec: https://github.com/dapr/docs/blob/master/reference/api/state_api.md#state-transactions

src/Dapr.Client/DaprClient.cs
vinayada1: fixed

src/Dapr.Client/DaprClient.cs
```diff
@@ -229,6 +229,35 @@ public async ValueTask<StateEntry<TValue>> GetStateEntryAsync<TValue>(string sto
...
+        /// store.
+        /// </summary>
+        /// <param name="storeName">The name of the state store.</param>
+        /// <param name="stateTransactionRequests">A list of StateTransactionRequests.</param>
+        /// <param name="metadata">An key/value pair that may be consumed by the state store.  This depends on the state store used.</param>
+        /// <param name="cancellationToken">A <see cref="CancellationToken" /> that can be used to cancel the operation.</param>
+        /// <returns>A <see cref="ValueTask" /> that will complete when the operation has completed.</returns>
+        public abstract Task ExecuteStateTransactionAsync(
+            string storeName,
+            IReadOnlyList<StateTransactionRequest> stateTransactionRequests,
```
LMWF, line 243: I think this name is fine, but throwing an idea out there - maybe it should be called operations like in the api spec?

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -450,6 +451,114 @@ public override async ValueTask<(TValue value, string etag)> GetStateAndETagAsyn
...
+        public override async Task ExecuteStateTransactionAsync(
+            string storeName,
+            IReadOnlyList<StateTransactionRequest> stateTransactionRequests,
+            Dictionary<string, string> metadata = default,
+            CancellationToken cancellationToken = default)
+        {
+            ArgumentVerifier.ThrowIfNullOrEmpty(storeName, nameof(storeName));
+            ArgumentVerifier.ThrowIfNull(stateTransactionRequests, nameof(stateTransactionRequests));
+            if (stateTransactionRequests.Count == 0)
+                throw new ArgumentException("states do not contain any elements");
```
LMWF, line 465: "states" should be renamed to the name of the arg

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -450,6 +451,114 @@ public override async ValueTask<(TValue value, string etag)> GetStateAndETagAsyn
...
+
+            try
+            {
+                await this.MakeExecuteStateTransactionCallAsync(
+                storeName,
```
LMWF, line 489: indentation

src/Dapr.Client/StateTransactionRequest.cs
```diff
@@ -0,0 +1,70 @@
...
+
+        /// <summary>
+        /// Gets or sets the value locally.
+        /// </summary>
+        public byte[] Value { get; set; }
+
+        /// <summary>
+        /// The Operation type.
+        /// </summary>
+        public string OperationType { get; set; }
```
LMWF, line 53: OperationType should be an enum from the user's view even though in the lower layers it's really a string with 2 possible values.

vinayada1: What should be the enum values? Insert, Delete, Upsert

LMWF: upsert and delete only.

src/Dapr.Client/StateTransactionRequest.cs
youngbupark: no state components use etag/metadata/options per operation. but I would keep these properties as optional.

vinayada1: Yes, they are optional. Added them as per the api spec

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -450,6 +451,114 @@ public override async ValueTask<(TValue value, string etag)> GetStateAndETagAsyn
...
+        public override async Task ExecuteStateTransactionAsync(
+            string storeName,
+            IReadOnlyList<StateTransactionRequest> operations,
+            Dictionary<string, string> metadata = default,
+            CancellationToken cancellationToken = default)
+        {
+            ArgumentVerifier.ThrowIfNullOrEmpty(storeName, nameof(storeName));
+            ArgumentVerifier.ThrowIfNull(operations, nameof(operations));
+            if (operations.Count == 0)
+                throw new ArgumentException($"{nameof(operations)} does not contain any elements");
```
amanbha, line 465: add {} around if

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -450,6 +451,114 @@ public override async ValueTask<(TValue value, string etag)> GetStateAndETagAsyn
...
+            return false;
+        }
+
+        private async ValueTask MakeExecuteStateTransactionCallAsync(
+            string storeName,
+            IReadOnlyList<StateTransactionRequest> states,
+            Dictionary<string, string> metadata = default,
+            CancellationToken cancellationToken = default)
+        {
+            Autogenerated.ExecuteStateTransactionRequest executeStateTransactionRequestEnvelope = new Autogenerated.ExecuteStateTransactionRequest()
```
amanbha, line 508: var can be used

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -450,6 +451,114 @@ public override async ValueTask<(TValue value, string etag)> GetStateAndETagAsyn
...
+            {
+                executeStateTransactionRequestEnvelope.Metadata.Add(metadata);
+            }
+
+            await this.MakeGrpcCallHandleError(
+                options => client.ExecuteStateTransactionAsync(executeStateTransactionRequestEnvelope, options),
+                cancellationToken);
+        }
+
+        private void FillStateTransactionRequest(Autogenerated.StateItem stateOperation, StateTransactionRequest state)
```
amanbha, line 536: this method can be:
```TransactionalStateOperation ToAutogeneratedTransactionalStateOperation(StateTransactionRequest state)``
which would create a TransactionalStateOperation and return it. easier for code readability

File: src/Dapr.Client/DaprClientGrpc.cs [hunk 10]
```diff
@@ -450,6 +451,114 @@ public override async ValueTask<(TValue value, string etag)> GetStateAndETagAsyn
...
+
+            await this.MakeExecuteStateTransactionCallAsync(
+                storeName,
+                operations,
+                metadata,
+                cancellationToken);
+        }
+
+        /// <inheritdoc/>
+        public override async ValueTask<bool> TryExecuteStateTransactionAsync(
```
amanbha, line 475: why this separate Try method is needed, this is basically ExecuteStateTransactionAsync with catching the RpcException

amanbha (changes requested):

src/Dapr.Client/DaprClientGrpc.cs
vinayada1: Tried to follow the convention for SaveState:-
public override async ValueTask<bool> TrySaveStateAsync<TValue>(

amanbha: TrySaveStateAsync exists because it accepts an etag which SaveState doesn't.

amanbha (approved):

src/Dapr.Client/DaprClientGrpc.cs
amanbha: approved the PR by mistake, we should address this.

vinayada1: ok will remove it

src/Dapr.Client/StateTransactionRequest.cs
```diff
@@ -0,0 +1,70 @@
...
+
+        /// <summary>
+        /// Additional key-value pairs to be passed to the state store (optional).
+        /// </summary>
+        public Dictionary<string, string> Metadata { get; set; }
+
+        /// <summary>
+        /// State Options (optional).
+        /// </summary>
+        public StateOptions Options { get; set; }
```
rynowak, line 68: Is there any reason to have these properties as get/set? They are passed via the constructor, some of the required parameters in the constructor are get/set (the value, and operation). It's a little odd - usually if you're going to have constructors like this, you just make the object immutable.

test/Dapr.Client.Test/StateApiTest.cs
```diff
@@ -13,11 +13,11 @@ namespace Dapr.Client.Test
     using Grpc.Core;
     using Grpc.Net.Client;
     using Xunit;
-    using System;
     using System.Collections.Generic;
     using StateConsistency = Dapr.Client.Autogen.Grpc.v1.StateOptions.Types.StateConsistency;
     using StateConcurrency = Dapr.Client.Autogen.Grpc.v1.StateOptions.Types.StateConcurrency;
     using Google.Protobuf;
+    using System;
```
rynowak, line 20: This sort order is already crazy. Generally it's preferred to do `System.*` stuff first (alphabetical) and then non-system (also alphabetical).

test/Dapr.Client.Test/StateApiTest.cs
```diff
@@ -224,6 +224,135 @@ public async Task SetStateAsync_ThrowsForNonSuccess()
...
+            req1.Request.Key.Should().Be("stateKey1");
+            req1.OperationType.Should().Be(StateOperationType.Upsert.ToString().ToLower());
+            var valueJson1 = req1.Request.Value.ToStringUtf8();
+            var value1 = JsonSerializer.Deserialize<Widget>(valueJson1);
+            value1.Size.Should().Be(stateValue1.Size);
+            value1.Color.Should().Be(stateValue1.Color);
+            req1.Request.Etag.Should().Be("testEtag");
+            req1.Request.Metadata.Count.Should().Be(1);
+            req1.Request.Metadata["a"].Should().Be("b");
+            //optionsFound1.Should().Be(Autogenerated.StateOptions.Concurrency.FirstWrite);
```
rynowak, line 276: do you mean to have this commented out? Is this supposed to be here or not :)

test/Dapr.Client.Test/StateApiTest.cs
```diff
@@ -224,6 +224,135 @@ public async Task SetStateAsync_ThrowsForNonSuccess()
...
+            var metadata1 = new Dictionary<string, string>()
+            {
+                {"a", "b" }
+            };
+            var options1 = new StateOptions();
+            options1.Concurrency = ConcurrencyMode.LastWrite;
+            ;
+            var state1 = new StateTransactionRequest("stateKey1", JsonSerializer.SerializeToUtf8Bytes(stateValue1), StateOperationType.Upsert, "testEtag", metadata1, options1);
+
+            int stateValue2 = 100;
```
rynowak, line 246: nitpick: try to stick to a style. So far I think everything in the repo uses `var`.

test/Dapr.Client.Test/StateApiTest.cs
```diff
@@ -224,6 +224,135 @@ public async Task SetStateAsync_ThrowsForNonSuccess()
...
+            var req3 = transactionRequest.Operations[2];
+            req3.Request.Key.Should().Be("stateKey3");
+            req3.OperationType.Should().Be(StateOperationType.Upsert.ToString().ToLower());
+            var valueJson3 = req3.Request.Value.ToStringUtf8();
+            var value3 = JsonSerializer.Deserialize<string>(valueJson3);
+            value3.Should().Be("teststring");
+        }
+
+        [Fact]
+        public async Task ExecuteStateTransactionAsync_CanClearState()
```
rynowak, line 295: What makes this test distinct from `ExecuteStateTransactionAsync_CanSaveState`? Both them create a transaction with multiple entries and the verify that the data is serialized correctly. Since you're not hitting an actual data store, the distinction between these two tests is kinda meaningless right?

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -450,6 +451,90 @@ public override async ValueTask<(TValue value, string etag)> GetStateAndETagAsyn
...
+
+            if (state.Value != null)
+            {
+                stateOperation.Value = ByteString.CopyFrom(state.Value);
+            }
+
+            if (state.ETag != null)
+            {
+                stateOperation.Etag = state.ETag;
+            }
```
rynowak, line 522: Question for both @vinayada1 and @youngbupark - what are the semantics of including an etag here? Does the transaction ignore it? Does the transaction rollback? 

We designed different APIs for the main state store case with ETags - because the application developer *generally needs to know the difference*.

It might be best to make this look more like `TryDeleteStateAsync` https://github.com/dapr/dotnet-sdk/pull/385/files#diff-ca47ccc1f9e29dbb552689c374bf9023L259

test/Dapr.Client.Test/StateApiTest.cs
```diff
@@ -224,6 +224,135 @@ public async Task SetStateAsync_ThrowsForNonSuccess()
...
+                .Build();
+
+            var stateValue1 = new Widget() { Size = "small", Color = "yellow", };
+            var metadata1 = new Dictionary<string, string>()
+            {
+                {"a", "b" }
+            };
+            var options1 = new StateOptions();
+            options1.Concurrency = ConcurrencyMode.LastWrite;
+            ;
```
rynowak, line 243: `;` 

Is probably the most abstract line of code I've seen in a while.

src/Dapr.Client/DaprClientGrpc.cs
youngbupark: Now, we didn't specify the exact behavior of etag for transaction.  Rather, the behavior of etag depends on state store implementation.

e.g. Redis state store ignores it for transaction operation. postgre and sql state store rollback the transaction because update is failed. some of state store, aws dynamodb state store doesn't support etag even for single operation and transactional operation at all.

cc/ @yaron2

src/Dapr.Client/StateTransactionRequest.cs
vinayada1: not really

test/Dapr.Client.Test/StateApiTest.cs
vinayada1: Fixed. Would be good if vscode could fix this automatically

test/Dapr.Client.Test/StateApiTest.cs
vinayada1: removed. I changed the data type so not needed

test/Dapr.Client.Test/StateApiTest.cs
vinayada1: :)

test/Dapr.Client.Test/StateApiTest.cs
vinayada1: Agree. Will remove

amanbha (approved):

Issue #393, opened 2020-09-14
Widen the pit of success with AddDapr/AddDaprClient


amolenk:
## Describe the proposal

This was triggered while talking to @edwinvw about #390.
The whole `AddDapr`, `AddDaprClient` thing is a bit complicated for ASP.NET devs wanting to use Dapr service invocation.

ASP.NET Web APIs by default use camel case JSON for the responses. The S.T.JsonSerializer used by DaprClient uses Pascal casing and is case sensitive by default.

So, in order to succesfully implement a scenario with multiple .NET microservices where service A calls service B using Dapr service invocation, the developer needs to specify the correct JsonSerializerOptions in service A. For example:

```
services.AddDaprClient(builder => builder.UseJsonSerializationOptions(new JsonSerializerOptions
{
    PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
    PropertyNameCaseInsensitive = true,
}));

services.Controllers().AddDapr();
```

If the `AddDaprClient` call is omitted, the service invocation itself will still succeed (because `services.Controllers.AddDapr()` also registers the `DaprClient`, but deserialization will fail (without any errors logged) because of the case sensitivity. I think this is confusing and hard to debug for developers.

Some ways to improve this:
- Use other default options for the JsonSerializer (`PropertyNameCaseInsensitive = true`)
- Or; allow the user to configure the DaprClient in the AddDapr() call.

amanbha:
@amolenk Thanks for reporting this confusing behavior. I believe both the suggested ways to improve can be implemented. Would you like to contribute a PR for it?

EdwinVW:
We can definitely create a PR. Should we also change the related _MarkerService_ behavior as stated in issue #390 (or do you want to look at this as a separate issue)?

amanbha:
Thanks @EdwinVW lets create separate PRs addressing these 2 issues.

EdwinVW:
PR is created (#394 ).

PR #394, opened 2020-09-14
Add ability to specify default DaprClient options to AddDapr

EdwinVW:
# Description

Added an (optional) argument to the `DaprMvcBuilderExtensions.AddDapr()` method that enables developers to specify a `DaprClientBuilder` in order to configure the default options of the `DaprClient` that is being added registered in DI by `AddDapr`.

## Issue reference

#393 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Updated samples


EdwinVW:
The commit message states this is a fix for issue `390`, but it should be `393`.

src/Dapr.AspNetCore/DaprMvcBuilderExtensions.cs
```diff
@@ -35,7 +38,18 @@ public static IMvcBuilder AddDapr(this IMvcBuilder builder)
                 return builder;
             }
 
-            builder.Services.AddDaprClient();
+            // create default clientbuilder config (if not specified)
+            if (configure == null)
+            {
+                configure = new Action<DaprClientBuilder>(
+                    builder => builder.UseJsonSerializationOptions(
```
amolenk, line 45: Better to move the default JsonSerializationOptions to DaprClient instead. That will ensure that the case insensitive options is always set, even when the user directly calls `AddDaprClient` instead of `AddDapr`. It also enables us to use case insensitive as the default options even when the user does pass a configure action to `AddDaprClient`.

amolenk (changes requested):

samples/AspNetCore/ControllerSample/Readme.md
```diff
@@ -136,13 +136,19 @@ On Windows:
 ```C#
  public void ConfigureServices(IServiceCollection services)
 {
-    services.AddControllers().AddDapr();
+    services.AddControllers().AddDapr(builder => 
+        builder.UseJsonSerializationOptions(
```
amolenk, line 140: We can remove the customized JSOn settings from the startup now that defaults are good enough.

samples/AspNetCore/ControllerSample/Startup.cs
```diff
@@ -37,13 +37,13 @@ public Startup(IConfiguration configuration)
...
         public void ConfigureServices(IServiceCollection services)
         {
-            services.AddControllers().AddDapr();
-
-            services.AddSingleton(new JsonSerializerOptions()
-            {
-                PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
-                PropertyNameCaseInsensitive = true,
-            });
+            services.AddControllers().AddDapr(builder =>
```
amolenk, line 40: We can remove the customized JSOn settings from the startup now that defaults are good enough.

amolenk (changes requested):

samples/AspNetCore/ControllerSample/Readme.md
EdwinVW: Moved JsonSerializer defaults to DaprClientBuilder (in d1c5447).

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -22,8 +22,8 @@ namespace Dapr.Client
     /// </summary>
     internal class DaprClientGrpc : DaprClient
     {
-        private readonly Autogenerated.Dapr.DaprClient client;
-        private readonly JsonSerializerOptions jsonSerializerOptions;
+        internal readonly Autogenerated.Dapr.DaprClient client;
```
amanbha, line 25: As I understand this is for validation in tests. expose it as internal get property.

src/Dapr.Client/DaprClientBuilder.cs
```diff
@@ -14,19 +14,22 @@ namespace Dapr.Client
     /// </summary>
     public sealed class DaprClientBuilder
     {
-        const string defaultDaprGrpcPort = "50001";
-        string daprEndpoint;
-        JsonSerializerOptions jsonSerializerOptions;
-        GrpcChannelOptions gRPCChannelOptions;
-
+        private const string defaultDaprGrpcPort = "50001";
+        internal string daprEndpoint;
```
amanbha, line 18: As I understand this is for validation in tests. expose it as internal get property.

amanbha (changes requested):

amanbha (approved):

amanbha:
Thanks for the contribution @EdwinVW . @vinayada1 could you please review and merge the PR.

Issue #399, opened 2020-09-03
Grpc.Core.RpcException when remote invocation results in BadRequest


denniszielke:
## Expected Behavior

The client sdk should be able to handle HTTP 400 Bad Request and make the request available to the calling application

## Actual Behavior

When the app returns a HTTP400 or BadRequest object the dapr client fails with an exception.

## Steps to Reproduce the Problem
Small modification of https://github.com/dapr/dotnet-sdk/blob/master/samples/AspNetCore/ControllerSample/Controllers/SampleController.cs to add the following method
to return a BadRequest
```
/// Do
[HttpPost("do")]
public IActionResult Do(int value)
{
    Console.WriteLine($"Doing {value}"); 
    return BadRequest(new
    {
        value = value
    });
}
```

When calling the method on the client via 
```
var a = await client.InvokeMethodAsync<object, Account>("routing", "do", data, httpExtension);
```
we are getting an exception
```
== APP == DepositUsingServiceInvocation

== APP == invoking

== APP == Unhandled exception. Grpc.Core.RpcException: Status(StatusCode=Internal, Detail="Bad Request")
== APP ==    at Dapr.Client.DaprClientGrpc.MakeGrpcCallHandleError[TResponse](Func`2 callFunc, CancellationToken cancellationToken) in /dotnet-sdk/src/Dapr.Client/DaprClientGrpc.cs:line 585
```

It seems there is no handling in the client sdk https://github.com/dapr/dotnet-sdk/blob/master/src/Dapr.Client/DaprClientGrpc.cs#L584

youngbupark:
moving back to dotnet-sdk. The fix for grpc-dotnet has been merged - https://github.com/grpc/grpc-dotnet/pull/1046
new version of grpc-dotnet will be released by this week or early next week. then we will use this version to fix this problem.

youngbupark:
Update - We are still waiting for grpc-dotnet 2.32.0 release. Once new nuget pkg is released, we will publish the minor version of dotnet sdk in 0.12.0 milestone.

vinayada1:
Short term - Will be providing code samples to extract the error message from the RpcException
Long term - https://github.com/dapr/dotnet-sdk/issues/400

PR #427, opened 2020-10-13
grpc exception handling sample

vinayada1:
# Description

Modified the sample to demonstrate how to extract error messages from RpcException

#399 

Please reference the issue this PR will close: #399 
## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/samples/AspNetCore/ControllerSample/Controllers/SampleController.cs
+++ b/samples/AspNetCore/ControllerSample/Controllers/SampleController.cs
@@ -80,5 +80,20 @@ namespace ControllerSample.Controllers
             await state.SaveAsync();
             return state.Value;
         }
+
+        /// <summary>
+        /// Method for throwing a BadRequest exception which will cause Dapr sidecar to throw an RpcException
+        /// </summary>
+        /// <param name="transaction">Transaction info.</param>
+        /// <param name="daprClient">State client to interact with Dapr runtime.</param>
+        /// <returns>A <see cref="Task{TResult}"/> representing the result of the asynchronous operation.</returns>
+        [HttpPost("throwException")]
+        public async Task<ActionResult<Account>> ThrowException(Transaction transaction, [FromServices] DaprClient daprClient)
+        {
+            Console.WriteLine("Enter ThrowException");
+            var task = Task.Delay(10);
+            await task;
+            return BadRequest("{statusCode: 404, message: not found...}");
+        }
     }
 }
--- a/samples/AspNetCore/ControllerSample/Controllers/SampleController.cs
+++ b/samples/AspNetCore/ControllerSample/Controllers/SampleController.cs
@@ -89,7 +89,7 @@ namespace ControllerSample.Controllers
             Console.WriteLine("Enter ThrowException");
             var task = Task.Delay(10);
             await task;
-            return BadRequest(@"{""statusCode"": 400, ""message"": ""bad request""}");
+            return BadRequest(new { statusCode = 400, message = "bad request" });
         }
     }
 }
--- a/samples/AspNetCore/ControllerSample/Controllers/SampleController.cs
+++ b/samples/AspNetCore/ControllerSample/Controllers/SampleController.cs
@@ -82,11 +82,7 @@ namespace ControllerSample.Controllers
         }
 
         /// <summary>
-        /// Method for throwing a BadRequest exception which will cause Dapr sidecar to throw an RpcException
-        /// </summary>
-        /// <param name="transaction">Transaction info.</param>
-        /// <param name="daprClient">State client to interact with Dapr runtime.</param>
-        /// <returns>A <see cref="Task{TResult}"/> representing the result of the asynchronous operation.</returns>
+        /// Method for returning a BadRequest result which will cause Dapr sidecar to throw an RpcException
         [HttpPost("throwException")]
         public async Task<ActionResult<Account>> ThrowException(Transaction transaction, [FromServices] DaprClient daprClient)
         {
--- a/samples/AspNetCore/ControllerSample/Controllers/SampleController.cs
+++ b/samples/AspNetCore/ControllerSample/Controllers/SampleController.cs
@@ -89,7 +89,7 @@ namespace ControllerSample.Controllers
             Console.WriteLine("Enter ThrowException");
             var task = Task.Delay(10);
             await task;
-            return BadRequest("{statusCode: 404, message: not found...}");
+            return BadRequest(@"{""statusCode"": 400, ""message"": ""bad request""}");
         }
     }
 }
--- a/samples/Client/DaprClient/Program.cs
+++ b/samples/Client/DaprClient/Program.cs
@@ -21,6 +21,9 @@ namespace DaprClient
         private static readonly string stateKeyName = "mykey";
         private static readonly string storeName = "statestore";
         private static readonly string pubsubName = "pubsub";
+        private static readonly string daprErrorInfoHTTPCodeMetadata  = "http.code";
+        private static readonly string daprErrorInfoHTTPErrorMetadata = "http.error_message";
+        private static readonly string grpcErrorInfoDetail = "google.rpc.ErrorInfo";
 
         /// <summary>
         /// Main entry point.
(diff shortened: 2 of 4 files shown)


samples/AspNetCore/ControllerSample/Controllers/SampleController.cs
```diff
@@ -80,5 +80,20 @@ public async Task<ActionResult<Account>> Withdraw(Transaction transaction, [From
             await state.SaveAsync();
             return state.Value;
         }
+
+        /// <summary>
+        /// Method for throwing a BadRequest exception which will cause Dapr sidecar to throw an RpcException
```
rynowak, line 85: > BadRequest exception which will cause Dapr sidecar to throw an RpcException

minor detail: there's no exception in this code - it's returning an HTTP 400 via the `BadRequestResult` type.

samples/AspNetCore/ControllerSample/Controllers/SampleController.cs
```diff
@@ -80,5 +80,20 @@ public async Task<ActionResult<Account>> Withdraw(Transaction transaction, [From
...
+        /// </summary>
+        /// <param name="transaction">Transaction info.</param>
+        /// <param name="daprClient">State client to interact with Dapr runtime.</param>
+        /// <returns>A <see cref="Task{TResult}"/> representing the result of the asynchronous operation.</returns>
```
rynowak, line 89: I don't think we should continue the practice of placing doc comments on every method in samples. There's a lot of boilerplate here that doesn't matter to anyone.

I'd suggest a normal comment that explains why this is here (what you have in the summary).

samples/Client/DaprClient/Program.cs
```diff
@@ -209,6 +212,34 @@ internal static async Task InvokeBalanceServiceOperationAsync(DaprClient client)
...
+
+            HTTPExtension httpExtension = new HTTPExtension()
+            {
+                Verb = HTTPVerb.Post
+            };
+
+            try
+            {
+                // Invokes a POST method named "Withdraw" that takes input of type "Transaction" as define in the RoutingSample.            
+                await client.InvokeMethodAsync<object>("controller", "throwException", data, httpExtension);
```
rynowak, line 228: minor nitpick - you can omit `<object>` here - someone who's reading this will likely be confused about why the generic type here doesn't match `data`'s declared type.

samples/Client/DaprClient/Program.cs
```diff
@@ -209,6 +212,34 @@ internal static async Task InvokeBalanceServiceOperationAsync(DaprClient client)
...
+            }
+            catch (Grpc.Core.RpcException ex)
```
rynowak, line 230: add a using `using Grcp.Core` to shorten up some of this code.

samples/Client/DaprClient/Program.cs
```diff
@@ -209,6 +212,34 @@ internal static async Task InvokeBalanceServiceOperationAsync(DaprClient client)
...
+            {
+                var entry = ex.Trailers.Get("grpc-status-details-bin");
+                var status = Google.Rpc.Status.Parser.ParseFrom(entry.ValueBytes);
+                Console.WriteLine("Grpc Exception Message: " + status.Message);
+                Console.WriteLine("Grpc Statuscode: " + status.Code);
+                var details = status.Details[0];
+                var rpcError = details.Unpack<Google.Rpc.ErrorInfo>();
+                Console.WriteLine("Grpc Exception: Http Error Message: " + rpcError.Metadata["http.error_message"]);
+                Console.WriteLine("Grpc Exception: Http Status Code: " + rpcError.Metadata["http.code"]);
```
rynowak, line 239: Are there string constants defined for these anywhere? I guess what I'm really asking about is whether `http.code` is a Dapr specific convention or a gRPC convention?

samples/AspNetCore/ControllerSample/Controllers/SampleController.cs
vinayada1: fixed

samples/AspNetCore/ControllerSample/Controllers/SampleController.cs
vinayada1: agree....the other stuff does not really matter in this case

samples/Client/DaprClient/Program.cs
```diff
@@ -209,6 +212,34 @@ internal static async Task InvokeBalanceServiceOperationAsync(DaprClient client)
...
+                // Invokes a POST method named "Withdraw" that takes input of type "Transaction" as define in the RoutingSample.            
+                await client.InvokeMethodAsync<object>("controller", "throwException", data, httpExtension);
+            }
+            catch (Grpc.Core.RpcException ex)
+            {
+                var entry = ex.Trailers.Get("grpc-status-details-bin");
+                var status = Google.Rpc.Status.Parser.ParseFrom(entry.ValueBytes);
+                Console.WriteLine("Grpc Exception Message: " + status.Message);
+                Console.WriteLine("Grpc Statuscode: " + status.Code);
+                var details = status.Details[0];
```
youngbupark, line 240: There is a chance where `status.Details` have no item or multiple items. 
We need to iterate all Details and check if it is ErrorInfo protobuf or not before unpacking data to ErrorInfo. 

If possible I would create new exception class or extend RpcException to have [extension methods](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/extension-methods) to get error method and status code.

For example, we can add the below four methods to RpcException using [extension methods](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/extension-methods).

1. GetMessage -> status.message
2. GetCode -> status.Code
3. GetHTTPErrorMessage -> rpcError.Metadata["http.error_message"]
4. GetHTTPErrorCode -> rpcError.Metadata["http.code"]

So user code can call these methods simply.

samples/AspNetCore/ControllerSample/Controllers/SampleController.cs
rynowak: The repo was set up to enforce full doc comments on all samples and tests - I got rid of that 😎

youngbupark (changes requested):

samples/Client/DaprClient/Program.cs
youngbupark: this is dapr-specific convention. GRPC doc doesn't provide guideline for metadata in ErrorInfo.

samples/Client/DaprClient/Program.cs
vinayada1: removed

samples/Client/DaprClient/Program.cs
youngbupark: http.code and http.error_message are dapr-specific - https://github.com/dapr/dapr/blob/bdd465455c8c6c2b00b3a7e1d32f5f5757b75d53/pkg/messaging/v1/util.go#L56-L57

samples/AspNetCore/ControllerSample/Controllers/SampleController.cs
```diff
@@ -80,5 +80,20 @@ public async Task<ActionResult<Account>> Withdraw(Transaction transaction, [From
...
+        [HttpPost("throwException")]
+        public async Task<ActionResult<Account>> ThrowException(Transaction transaction, [FromServices] DaprClient daprClient)
+        {
+            Console.WriteLine("Enter ThrowException");
+            var task = Task.Delay(10);
+            await task;
+            return BadRequest("{statusCode: 404, message: not found...}");
```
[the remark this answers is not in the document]
rynowak: yeah this is a little confused. If you want to return a 404, then I'd suggest `NotFound()`.

If you want to specify some content use `NotFound(new { message: "not found" });` and it will get JSON serialized. 

ASP.NET Core will return a default JSON blob for errors when you have `[ApiController]` on the class and don't specify your own response body - using problem details.

samples/Client/DaprClient/Program.cs
rynowak: The goal for this PR isn't to design new APIs or add features, just to be an example of what's possible right now.

We're going to write a prososal for that separately.

youngbupark: @rynowak I think we should fix runtime as well. The other gRPC APIs doesn't handle error correctly - https://github.com/dapr/dapr/issues/2067 We need to discuss about this issue as well.

samples/Client/DaprClient/Program.cs
vinayada1: fixed

samples/Client/DaprClient/Program.cs
vinayada1: I fixed the part for iterating over the Details

samples/AspNetCore/ControllerSample/Controllers/SampleController.cs
```diff
@@ -80,5 +80,16 @@ public async Task<ActionResult<Account>> Withdraw(Transaction transaction, [From
...
+
+        /// <summary>
+        /// Method for returning a BadRequest result which will cause Dapr sidecar to throw an RpcException
+        [HttpPost("throwException")]
+        public async Task<ActionResult<Account>> ThrowException(Transaction transaction, [FromServices] DaprClient daprClient)
+        {
+            Console.WriteLine("Enter ThrowException");
+            var task = Task.Delay(10);
+            await task;
+            return BadRequest(@"{""statusCode"": 400, ""message"": ""bad request""}");
```
rynowak, line 92: so rather than putting JSON text in a string, do this. 

`BadRequest(new { statusCode: 400, message: "bad request", });`

vinayada1: fixed

youngbupark (approved):

Issue #400, opened 2020-09-18
Allow return of HTTP response headers from method invocations


philliphoff:
Could there also be a way of extracting HTTP headers sent by an application with the response?  For example, if the response was an opaque set of bytes, the `Content-Type` response header might be needed in order to differentiate between multiple types of data.  Something of this sort:

```csharp

class HttpResponseExtension
{
   IDictionary<string, string> Headers { get; }
}

class MethodResponse<TResponse>
{
    TResponse Content { get; }

    HttpResponseExtension Metadata { get; }
}

class DaprClient
{
    abstract Task<MethodResponse<TResponse>> InvokeMethodWithResponseAsync<TRequest, TResponse>(string appId, string methodName, HTTPExtension requestMetadata, CancellationToken);
}

```

@amolenk suggested a variation on this theme:

The service that's being called may be an HTTP service, but it could also be a gRPC service. So I'd do it slightly different:

```
class MethodResponse<TResponse>
{
    TResponse Content { get; }
}

class HttpResponse<TResponse> : MethodResponse<TResponse>
{
    IDictionary<string, string> Headers { get; }
}

class DaprClient
{
    abstract Task<MethodResponse<TResponse>> InvokeMethodAsync<TRequest, TResponse>(string appId, string methodName, HTTPExtension requestMetadata, CancellationToken);
}

```

Alternatively, at that point it may even be better to make separate invoke methods for Http and gRPC, like InvokeHttpMethod and InvokeGrpcMethod. The abstraction is already a bit leaky due to the need for the HttpExtension object.

rynowak:
@vinayada1 as well

Agree, we didn't include the ability to expose the underlying HTTP semantics as part of the initial design here, but https://github.com/dapr/dotnet-sdk/issues/399 is causing us to revisit this for the error case, and we should probably also expose it as an API for the successful case.

vinayada1:
Adding a proposal here:-
Goal: Enable dotnet sdk to call http or gRPC servers with all the necessary fields and be able to receive responses including error messages.

The InvokeRequest and InvokeResponse should include all fields that are specific to http and gRPC protocols. We will treat http as a superset instead of having separate methods to invoke http and gRPC which allows more flexibility to the user.

Service invocation can use the method below:-
public abstract Task<InvokeResponse> InvokeMethodRawAsync(
            InvokeRequest request, 
            CancellationToken cancellationToken);

The Invoke Request will look as below:-
```
    public class InvokeRequest
    {
[...]

PR #432, opened 2020-10-14
Rich error model

vinayada1:
# Description

Ability to return error messages from service invocation

#400 

Please reference the issue this PR will close: #400 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


src/Dapr.Client/InvokeResponse.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        /// <summary>
+        /// Gets or sets the HTTP status info.
+        /// </summary>
+        public int? HttpStatusCode { get; set; }
+
+        /// <summary>
+        /// Gets or sets the gRPC status info.
+        /// </summary>
+        public GRPCStatusInfo GrpcStatusInfo { get; set; }
```
rynowak, line 71: After reading the code I'm curious - about the semantics of what's populated between `HttpStatusCode` and `GrpcStatusInfo`.

What's populated when I'm calling HTTP vs GRPC on the other side?

I was under the impression that we'd populate one set of fields based on what protocol the callee is using.

src/Dapr.Client/Dapr.Client.csproj
```diff
@@ -17,6 +17,8 @@
         <PackageReference Include="Google.Protobuf" Version="3.13.0" />
         <PackageReference Include="Grpc.Net.Client" Version="2.32.0" />
         <PackageReference Include="Grpc.Tools" Version="2.32.0" PrivateAssets="All" />
+        <PackageReference Include="Google.Api.CommonProtos" Version="2.2.0" />
+        <PackageReference Include="moq" Version="4.14.7" />
```
rynowak, line 21: This should be a reference of the test project not here.

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -294,10 +355,100 @@ public override async ValueTask<IReadOnlyList<BulkStateItem>> GetBulkStateAsync(
...
+            var daprApiToken = Environment.GetEnvironmentVariable("DAPR_API_TOKEN");
+
+            if (daprApiToken != null)
+            {
+                callOptions.Headers.Add("dapr-api-token", daprApiToken);
+            }
+
+            // Common Exception Handling logic can be added here for all calls.
+            var grpcCall = client.InvokeServiceAsync(request, callOptions);
+            return grpcCall;
```
rynowak, line 370: I would really strongly suggest refactoring this code path.

Calling an async method without immediately awaiting the result has some significant gotchas - because exceptions *might* be thrown synchronously or asynchronously - and that could change without  being considered a breaking change (and break you). 

It's better if you could have `MakeInvokeRequest` return the values of `request` and `callOptions` and call `client.InvokeServiceAsync` where your error handling code is.

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -294,10 +355,100 @@ public override async ValueTask<IReadOnlyList<BulkStateItem>> GetBulkStateAsync(
...
+                {
+                    invokeResponse.HttpStatusCode = (HttpStatusCode)System.Enum.Parse(typeof(HttpStatusCode), Encoding.UTF8.GetString(headers[daprHTTPStatusHeader], 0, headers[daprHTTPStatusHeader].Length));
+                    invokeResponse.ContentType = Constants.ContentTypeApplicationJson;
+                }
+                else
+                {
+                    // Response is grpc
+                    invokeResponse.GrpcStatusInfo = new GRPCStatusInfo(grpcStatus.StatusCode, grpcStatus.Detail);
+                    invokeResponse.ContentType = Constants.ContentTypeApplicationGrpc;
+                }
```
rynowak, line 411: this loos like what I expected 👍

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -294,10 +355,100 @@ public override async ValueTask<IReadOnlyList<BulkStateItem>> GetBulkStateAsync(
...
+                var response = await grpcCall.ResponseAsync;
+                var responseHeaders = await grpcCall.ResponseHeadersAsync;
+                var trailers = grpcCall.GetTrailers();
+                var grpcStatus = grpcCall.GetStatus();
+
+                var headers = grpcCall.ResponseHeadersAsync.Result.ToDictionary(kv => kv.Key, kv => kv.ValueBytes);
+
+                    invokeResponse.Body = response.Data.Value.IsEmpty ? default : TypeConverters.FromAny<TResponse>(response.Data, this.jsonSerializerOptions);
+                    invokeResponse.Headers = headers;
+                    invokeResponse.Trailers = grpcCall.GetTrailers().ToDictionary(kv => kv.Key, kv => kv.ValueBytes);
```
rynowak, line 399: looks like formatting is off here

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -294,10 +355,100 @@ public override async ValueTask<IReadOnlyList<BulkStateItem>> GetBulkStateAsync(
...
+            var entry = ex.Trailers.Get(grpcStatusDetails);
+            if (entry != null)
+            {
+                var status = Google.Rpc.Status.Parser.ParseFrom(entry.ValueBytes);
+                foreach (var detail in status.Details)
+                {
+                    if (Google.Protobuf.WellKnownTypes.Any.GetTypeName(detail.TypeUrl) == grpcErrorInfoDetail)
+                    {
+                        var rpcError = detail.Unpack<Google.Rpc.ErrorInfo>();
+                        var grpcStatusCode = (StatusCode) System.Enum.ToObject(typeof(StatusCode), status.Code);
```
rynowak, line 432: You should be able to do this like `(StatusCode)status.Code`

rynowak: I've never even heard of `Enum.ToObject(..)` 😆

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -294,10 +355,100 @@ public override async ValueTask<IReadOnlyList<BulkStateItem>> GetBulkStateAsync(
...
+                        invokeResponse.GrpcStatusInfo = new GRPCStatusInfo(grpcStatusCode, status.Message, Convert.ToInt32(rpcError.Metadata[daprErrorInfoHTTPCodeMetadata]), rpcError.Metadata[daprErrorInfoHTTPErrorMetadata]);
+                    }
+                }
+            }
+
+            return new ServiceInvocationException<TRequest, TResponse>($"Exception while invoking {invokeResponse.Request.MethodName} on appId:{invokeResponse.Request.AppId}", ex, invokeResponse);
```
rynowak, line 438: Let's think about this more. I'm not sure we want to make the exception a generic type. 

Writing error handling that's general purpose doesn't really work when the exception has generics in it, because you'd have to predict all of the replacements of `TRequest, TResponse`. 

Generic exception are exceedingly rare for this reason.

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -294,10 +355,100 @@ public override async ValueTask<IReadOnlyList<BulkStateItem>> GetBulkStateAsync(
...
+        }
+
+        private bool IsResponseFromHttpCallee(Dictionary<string,byte[]> headers)
```
rynowak, line 441: You can replace this with `headers.ContainsKey(daprHttpStatusHeader)`. Dictionaries are like hash tables or maps

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -294,10 +355,100 @@ public override async ValueTask<IReadOnlyList<BulkStateItem>> GetBulkStateAsync(
...
+                var headers = grpcCall.ResponseHeadersAsync.Result.ToDictionary(kv => kv.Key, kv => kv.ValueBytes);
+
+                    invokeResponse.Body = response.Data.Value.IsEmpty ? default : TypeConverters.FromAny<TResponse>(response.Data, this.jsonSerializerOptions);
+                    invokeResponse.Headers = headers;
+                    invokeResponse.Trailers = grpcCall.GetTrailers().ToDictionary(kv => kv.Key, kv => kv.ValueBytes);
+
+                if (IsResponseFromHttpCallee(headers))
+                {
+                    invokeResponse.HttpStatusCode = (HttpStatusCode)System.Enum.Parse(typeof(HttpStatusCode), Encoding.UTF8.GetString(headers[daprHTTPStatusHeader], 0, headers[daprHTTPStatusHeader].Length));
+                    invokeResponse.ContentType = Constants.ContentTypeApplicationJson;
```
rynowak, line 404: Since `headers` is a dictionary, can you replace this pattern with:

```C#
if (headers.TryGetValue(daprHttpStatusHeader, out var httpStatus))
{
}
```

This lets you check for presence and bind the value to a variable in one line.

src/Dapr.Client/GRPCStatusInfo.cs
```diff
@@ -0,0 +1,48 @@
...
+{
+    using Dapr.Client.Http;
+    using System;
+    using System.Net;
+
+    /// <summary>
+    /// This class is only needed if the app you are calling app is listening on gRPC.
+    /// It contains propertes that represent status info that may be populated for an gRPC response.
+    /// </summary>
+    public class GRPCStatusInfo
```
rynowak, line 15: The .NET naming convention for this would be `GrpcStatusInfo`. This would also be consistent with how the .NET gRPC APIs are named (see line 20).

src/Dapr.Client/GRPCStatusInfo.cs
```diff
@@ -0,0 +1,48 @@
...
+    {
+        /// <summary>
+        /// The constructor.
+        /// </summary>
+        public GRPCStatusInfo(Grpc.Core.StatusCode statusCode, string message, int? innerHttpStatusCode = default, string InnerHttpErrorMessage = default)
+        {
+            this.GrpcStatusCode = statusCode;
+            this.GrpcErrorMessage = message;
+            this.InnerHttpStatusCode = innerHttpStatusCode;
+            this.InnerHttpErrorMessage = InnerHttpErrorMessage;
```
rynowak, line 25: What's the meaning of these two?

src/Dapr.Client/GRPCStatusInfo.cs
```diff
@@ -0,0 +1,48 @@
...
+    /// <summary>
+    /// This class is only needed if the app you are calling app is listening on gRPC.
+    /// It contains propertes that represent status info that may be populated for an gRPC response.
+    /// </summary>
+    public class GRPCStatusInfo
+    {
+        /// <summary>
+        /// The constructor.
+        /// </summary>
+        public GRPCStatusInfo(Grpc.Core.StatusCode statusCode, string message, int? innerHttpStatusCode = default, string InnerHttpErrorMessage = default)
```
rynowak, line 20: `InnerHttpErrorMessage` -> `innerHttpErrorMessage`

src/Dapr.Client/ServiceInvocationException.cs
```diff
@@ -0,0 +1,54 @@
...
+namespace Dapr.Client
+{
+    using Dapr.Client.Http;
+    using System;
+    using System.Net;
+
+    /// <summary>
+    /// This class is only needed if the app you are calling app is listening on gRPC.
+    /// It contains propertes that represent status info that may be populated for an gRPC response.
+    /// </summary>
```
rynowak, line 14: This seems like a bad copy-paste

src/Dapr.Client/ServiceInvocationException.cs
```diff
@@ -0,0 +1,54 @@
...
+            : base(message, innerException)
+        {
+            this.Response = response;
+        }
+
+
+        /// <summary>
+        /// The gRPC Error Message
+        /// </summary>
+        public string GrpcErrorMessage { get; set; }
```
rynowak, line 47: What's the purpose of this? how's it different from the message of the exception? Is this the message returned byt he dapr sidecar? example?

src/Dapr.Client/ServiceInvocationException.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        /// <summary>
+        /// The Response
+        /// </summary>
+        public ServiceInvocationResponse<TRequest, TResponse> Response { get; set; }
```
rynowak, line 52: Is there a reason for this to be get/set? It would be more conventional to be get-only. It's not typical to modify an exception after its thrown

src/Dapr.Client/ServiceInvocationRequest.cs
```diff
@@ -0,0 +1,69 @@
...
+        {
+            ArgumentVerifier.ThrowIfNull(appId, nameof(appId));
+            ArgumentVerifier.ThrowIfNull(methodName, nameof(methodName));
+
+            this.AppId = appId;
+            this.MethodName = methodName;
+            this.Body = body;
+            this.ContentType = contentType;
+            this.HttpExtension = httpExtension;
+        }
```
rynowak, line 42: As general design feedback, it's better to stick to one design pattern. If the class accepts a bunch of constructor arguments, I usually expect it to be immutable, or to have those fields immutable.

It's a little off to have a no-arg constructor and one that validates. Having something show up in the constructor (vs a property) doesn't really add convenience - both are easy to use.

test/Dapr.Client.Test/InvokeApiTest.cs
```diff
@@ -5,10 +5,14 @@
 
 namespace Dapr.Client.Test
 {
+    using System;
     using System.Collections.Generic;
     using System.Net;
+    using System.Text;
     using System.Text.Json;
     using System.Threading.Tasks;
+    using System.Runtime.Serialization.Formatters.Binary;
```
rynowak, line 14: What's this being used for? This sets off some alarm bells.

test/Dapr.Client.Test/InvokeApiTest.cs
```diff
@@ -18,6 +22,12 @@ namespace Dapr.Client.Test
     using Grpc.Net.Client;
     using Xunit;
     using System.Linq;
+    using ProtoBuf.Grpc.Client;
+    using Microsoft.AspNetCore.Mvc.Testing;
+    using Moq;
+    // using Moq.SetupAsync;
```
rynowak, line 28: delete

test/Dapr.Client.Test/InvokeApiTest.cs
```diff
@@ -18,6 +22,12 @@ namespace Dapr.Client.Test
...
+    using Autogenerated = Autogen.Grpc.v1;
+    using System.IO;
```
rynowak, line 30: sort

test/Dapr.Client.Test/InvokeApiTest.cs
```diff
@@ -413,6 +423,271 @@ public async Task InvokeMethodAsync_WithReturnTypeAndData_WithNonDefaultVerb_Wit
...
+        {
+            var client = new MockClient();
+            var data = new Response() { Name = "Look, I was invoked!" };
+            var invokeResponse = new InvokeResponse();
+            invokeResponse.Data = TypeConverters.ToAny(data);
+            
+            var response = 
+                client.Call<InvokeResponse>()
+                .SetResponse(invokeResponse)
+                .AddHeader("dapr-http-status", "200")
```
rynowak, line 467: Should we include a user-defined header here? (something that doesn't get special handling by dapr)

test/Dapr.Client.Test/InvokeApiTest.cs
```diff
@@ -413,6 +423,271 @@ public async Task InvokeMethodAsync_WithReturnTypeAndData_WithNonDefaultVerb_Wit
...
+
+            var body = new Request() { RequestParameter = "Hello " };
+            var task = client.DaprClient.InvokeMethodWithResponseHeadersAsync<Request, Response>("test", "testMethod", body);
+
+            // Validate Response
+            var invokedResponse = await task;
+            invokedResponse.Body.Name.Should().Be("Look, I was invoked!");
+            invokedResponse.HttpStatusCode.Should().BeNull();
+            invokedResponse.GrpcStatusInfo.Should().NotBeNull();
+            invokedResponse.GrpcStatusInfo.GrpcStatusCode.Should().Be(Grpc.Core.StatusCode.OK);
```
rynowak, line 453: we should verify the header here too

test/Dapr.Client.Test/InvokeApiTest.cs
```diff
@@ -413,6 +423,271 @@ public async Task InvokeMethodAsync_WithReturnTypeAndData_WithNonDefaultVerb_Wit
...
+            // Validate Response
+            var invokedResponse = await task;
+            invokedResponse.Body.Name.Should().Be("Look, I was invoked!");
+            invokedResponse.GrpcStatusInfo.Should().BeNull();
+            invokedResponse.HttpStatusCode.Should().NotBeNull();
+            invokedResponse.HttpStatusCode.Should().Be(HttpStatusCode.OK);
+        }
+
+        [Fact]
+        public async Task InvokeMethodAsync_CanInvokeMethodWithResponseHeaders_ServerThrowsRpcException()
```
rynowak, line 487: `ServerThrowsRpcException`- we should pick a better name here. Servers don't throw exceptions, they return responses.

test/Dapr.Client.Test/InvokeApiTest.cs
```diff
@@ -413,6 +423,271 @@ public async Task InvokeMethodAsync_WithReturnTypeAndData_WithNonDefaultVerb_Wit
...
+            catch(ServiceInvocationException<Request, Response> ex)
+            {
+                ex.Message.Should().Be("Exception while invoking testMethod on appId:test");
+                ex.InnerException.Message.Should().Be(rpcExceptionMessage);
+
+                ex.Response.GrpcStatusInfo.GrpcErrorMessage.Should().Be(grpcStatusMessage);
+                ex.Response.GrpcStatusInfo.GrpcStatusCode.Should().Be(grpcStatusCode);
+                ex.Response.GrpcStatusInfo.InnerHttpErrorMessage.Should().Be(grpcErrorInfoDetailHttpErrorMsg);
+                ex.Response.GrpcStatusInfo.InnerHttpStatusCode.Should().Be(grpcErrorInfoDetailHttpCode);
+                ex.Response.HttpStatusCode.Should().BeNull();
```
rynowak, line 553: I'm a little confused about this given that `http.code` is set in the code above - what is this intended to test exactly?

test/Dapr.Client.Test/InvokeApiTest.cs
```diff
@@ -413,6 +423,271 @@ public async Task InvokeMethodAsync_WithReturnTypeAndData_WithNonDefaultVerb_Wit
...
+            invokedResponse.Body.Name.Should().Be("Look, I was invoked!");
+            invokedResponse.GrpcStatusInfo.Should().BeNull();
+            invokedResponse.HttpStatusCode.Should().NotBeNull();
+            invokedResponse.HttpStatusCode.Should().Be(HttpStatusCode.OK);
+        }
+
+        [Fact]
+        public async Task InvokeMethodAsync_CanInvokeMethodWithResponseHeaders_ServerThrowsRpcException()
+        {
+            var client = new MockClient();
```
rynowak, line 489: I think we need really need to have a test for error handling with each of gRPC and HTTP on the callee side. This is the scenario that's been problematic/undertested.

test/Dapr.Client.Test/MockClient.cs
```diff
@@ -0,0 +1,91 @@
+using System.Collections.Generic;
```
rynowak, line 1: missing copyright

test/Dapr.Client.Test/MockClient.cs
```diff
@@ -0,0 +1,91 @@
...
+            .Setup(m => m.InvokeServiceAsync(It.IsAny<Autogen.Grpc.v1.InvokeServiceRequest>(), It.IsAny<CallOptions>()))
+            .Throws(rpcException);
+        }
+
+        // public void SetupMockToRespond(InvokeResponse response)
+        // {
+        //     this.Mock
+        //     .Setup(m => m.InvokeServiceAsync(It.IsAny<Autogen.Grpc.v1.InvokeServiceRequest>(), It.IsAny<CallOptions>()))
+        //     .Returns(response);
+        // }
```
rynowak, line 41: delete

rynowak (changes requested):

src/Dapr.Client/DaprClient.cs
```diff
@@ -134,6 +134,39 @@ public abstract class DaprClient
...
+        /// <summary>
+        /// Invokes a method on a Dapr app.
+        /// </summary>
+       /// <param name="appId">The Dapr application id to invoke the method on.</param>
+        /// <param name="methodName">The name of the method to invoke.</param>  
+        /// <param name="data">Data to pass to the method</param>      
+        /// <param name="httpExtension">Additional fields that may be needed if the receiving app is listening on HTTP.</param>
+        /// <param name="cancellationToken">A <see cref="CancellationToken" /> that can be used to cancel the operation.</param>
+        /// <returns>A <see cref="Task{InvokeResponse}" /> that will return the value when the operation has completed.</returns>
+        public abstract Task<ServiceInvocationResponse<TRequest, TResponse>> InvokeMethodWithResponseHeadersAsync<TRequest, TResponse>(
```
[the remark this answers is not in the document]
vinayada1: There is an Autogenerated.InvokeResponse and the name can get confusing which was the reason I used this name. We can use InvocationResponse

src/Dapr.Client/DaprClientGrpc.cs
vinayada1: My bad!

test/Dapr.Client.Test/InvokeApiTest.cs
vinayada1: Unused.

src/Dapr.Client/GRPCStatusInfo.cs
vinayada1: in the case of grpc to http, the innerHttpStatusCode and error message are the http status code and error message returned by the HTTP server. the grpcstatuscode and grpcerrormessage has the values returned by Dapr which are converted values from http to grpc

test/Dapr.Client.Test/InvokeApiTest.cs
vinayada1: When Dapr sidecar receives HTTP error code = 500 and error message in the response body, it maps the http status code to grpc status code. It preserves the original error code (as metadata key = http.code) and error message (as metadata key = http.error_message) and sends it in the trailer. This test is simulating the response from Dapr sidecar and checking whether the dotnet sdk is able to parse it correctly. This is similar to the sample code provided to Zeiss in https://github.com/dapr/dotnet-sdk/issues/399

src/Dapr.Client/ServiceInvocationException.cs
vinayada1: yes, that's the message returned by Dapr side car. https://github.com/dapr/dapr/blob/6af5d6e6a497aae2cf9deb4759f7be3be20193cf/pkg/messaging/v1/util.go#L319

and GrpcStatusCode is the grpc status code returned by Dapr sidecar which is a mapped value from the http status code.

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -43,6 +51,12 @@ internal DaprClientGrpc(GrpcChannel channel, JsonSerializerOptions jsonSerialize
             this.client = new Autogenerated.Dapr.DaprClient(channel);
         }
 
+        internal DaprClientGrpc(Autogenerated.Dapr.DaprClient inner, JsonSerializerOptions jsonSerializerOptions)
+        {
+            this.client = inner;
+            this.jsonSerializerOptions = jsonSerializerOptions;
+        }
+
```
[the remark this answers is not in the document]
vinayada1: had not removed references to the old one

src/Dapr.Client/DaprClientGrpc.cs
vinayada1: Valid point but we cannot include the Response in the exception then.

test/Dapr.Client.Test/InvokeApiTest.cs
vinayada1: I am trying to mock the response from dapr sidecar here.

test/Dapr.Client.Test/InvokeApiTest.cs
rynowak: so I think I'm still confused. If this test **represents** an HTTP 500 returned by an HTTP callee - then why is `response.HttpStatusCode` null?

vinayada1: I have updated the tests to separate the ones for GRPC callee and HTTP callee. So this should be null for grpc callee. This comment is referencing the old code?

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -157,7 +165,10 @@ private static T ConvertFromInvokeBindingResponse<T>(Autogenerated.InvokeBinding
             ArgumentVerifier.ThrowIfNullOrEmpty(appId, nameof(appId));
             ArgumentVerifier.ThrowIfNullOrEmpty(methodName, nameof(methodName));
 
-            _ = await this.MakeInvokeRequestAsync(appId, methodName, null, httpExtension, cancellationToken);
+            Autogenerated.InvokeServiceRequest request;
+            CallOptions callOptions;
+            (request, callOptions) = this.MakeInvokeRequestAsync(appId, methodName, null, httpExtension, cancellationToken);
```
rynowak, line 170: btw the simple way to write this is `var (request, callOptions) = ...`

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -35,12 +43,12 @@ internal class DaprClientGrpc : DaprClient
         /// <summary>
         /// Initializes a new instance of the <see cref="DaprClientGrpc"/> class.
         /// </summary>
-        /// <param name="channel">gRPC channel to create gRPC clients.</param>
+        /// <param name="inner">Autogenerated.Dapr.DaprClient.</param>
         /// <param name="jsonSerializerOptions">Json serialization options.</param>
-        internal DaprClientGrpc(GrpcChannel channel, JsonSerializerOptions jsonSerializerOptions = null)
+        internal DaprClientGrpc(Autogenerated.Dapr.DaprClient inner, JsonSerializerOptions jsonSerializerOptions)
```
rynowak, line 48: It might be beneficial to just delete the docs. There's no real value for doing that for an internal method (unless you have something interesting to say).

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -294,10 +358,123 @@ public override async ValueTask<IReadOnlyList<BulkStateItem>> GetBulkStateAsync(
...
+
+            return (request, callOptions);
+        }
+
+        private async Task<InvocationResponse<TResponse>> MakeInvokeRequestAsyncWithResponse<TRequest, TResponse>(
+            InvocationRequest<TRequest> request,
+            CancellationToken cancellationToken = default)
+        {
+            
+
```
rynowak, line 379: extra line police 🚨 🚨 🚨 🚨 🚨 🚨 🚨 🚨 🚨 🚨

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -294,10 +358,123 @@ public override async ValueTask<IReadOnlyList<BulkStateItem>> GetBulkStateAsync(
...
+        private async Task<InvocationResponse<TResponse>> MakeInvokeRequestAsyncWithResponse<TRequest, TResponse>(
+            InvocationRequest<TRequest> request,
+            CancellationToken cancellationToken = default)
+        {
+            
+            
+            Any serializedData = null;
+            if (request.Body != null)
+            {
+                if (typeof(TResponse) == typeof(byte[]))
```
rynowak, line 383: This is not a correct assumption. Serialization should  live outside this method.

It's totally valid to use `byte[]` with JSON serialization. It will result in a base64 encoded JSON string. I'd suggest that the callers of this method do the serialization.

Can you please add a test for `byte[]` with JSON so we don't regress this?

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -294,10 +358,123 @@ public override async ValueTask<IReadOnlyList<BulkStateItem>> GetBulkStateAsync(
...
+                var grpcCall = client.InvokeServiceAsync(invokeRequest, callOptions);
+
+                var response = await grpcCall.ResponseAsync;
+                var responseHeaders = await grpcCall.ResponseHeadersAsync;
+                var trailers = grpcCall.GetTrailers();
+                var grpcStatus = grpcCall.GetStatus();
+
+                var headers = grpcCall.ResponseHeadersAsync.Result.ToDictionary(kv => kv.Key, kv => kv.ValueBytes);
+
+                if (typeof(TResponse) == typeof(byte[]))
```
rynowak, line 410: Again the same point applies. `byte[]` is just another .NET type, it doesn't tell you any information about the format used.

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -294,10 +358,123 @@ public override async ValueTask<IReadOnlyList<BulkStateItem>> GetBulkStateAsync(
...
+                {
+                    // User wants to receive raw bytes
+                    var responseBytes = new byte[response.Data.Value.Length];
+                    response.Data.Value?.CopyTo(responseBytes, 0);
+                    invokeResponse.Body = (TResponse)(response.Data.Value.IsEmpty ? default : Convert.ChangeType(responseBytes, typeof(TResponse)));
```
rynowak, line 415: You can remove `Convert.ChangeType` here and just do a cast. You already know that `TResponse` is the same type as `byte[]`. `Convert.ChangeType` is for converting values between representations - like converting a floating point number to a decimal.

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -294,10 +358,123 @@ public override async ValueTask<IReadOnlyList<BulkStateItem>> GetBulkStateAsync(
...
+        {
+            
+            
+            Any serializedData = null;
+            if (request.Body != null)
+            {
+                if (typeof(TResponse) == typeof(byte[]))
+                {
+                    // User has passed in raw bytes
+                    var requestBytes = (byte[])Convert.ChangeType(request.Body, typeof(byte[]));
```
rynowak, line 386: You don't need `Convert.ChangeType` here, just do a cast.

rynowak (changes requested):

src/Dapr.Client/DaprClientGrpc.cs
vinayada1: I had tried the cast....getting this without the Convert.ChangeType. DaprClientGrpc.cs(377,40): error CS0030: Cannot convert type '(TRequest Body, System.Type)' to 'byte[]' [/Users/vinayada/dapr/dotnet-sdk/src/Dapr.Client/Dapr.Client.csproj]

src/Dapr.Client/DaprClientGrpc.cs
vinayada1: Raw explicity calls with a flag now

src/Dapr.Client/DaprClientGrpc.cs
vinayada1: I don't think the UT will catch anything because the server is mocked.Changed to instead make the raw format explicit rather than deducing from data type

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -176,7 +180,10 @@ private static T ConvertFromInvokeBindingResponse<T>(Autogenerated.InvokeBinding
                 serializedData = TypeConverters.ToAny(data, this.jsonSerializerOptions);
             }
 
-            _ = await this.MakeInvokeRequestAsync(appId, methodName, serializedData, httpExtension, cancellationToken);
+            Autogenerated.InvokeServiceRequest request;
+            CallOptions callOptions;
+            (request, callOptions) = this.MakeInvokeRequestAsync(appId, methodName, serializedData, httpExtension, cancellationToken);
```
rynowak, line 185: these can be cleaned up to `var (request, callOptions) = ` there's a few more cases

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -202,14 +212,61 @@ private static T ConvertFromInvokeBindingResponse<T>(Autogenerated.InvokeBinding
...
-            }
+                AppId = appId,
+                MethodName = methodName,
+                Body = data,
+                HttpExtension = httpExtension,
+            };
 
-            var response = await this.MakeInvokeRequestAsync(appId, methodName, serializedData, httpExtension, cancellationToken);
-            return response.Data.Value.IsEmpty ? default : TypeConverters.FromAny<TResponse>(response.Data, this.jsonSerializerOptions);
+            var invokeResponse = await this.MakeInvokeRequestAsyncWithResponse<TRequest, TResponse>(request, false, cancellationToken);
```
rynowak, line 223: Use a named argument if you want this to be more glanceable:

```C#
await this.MakeInvokeRequestAsyncWithResponse<TRequest, TResponse>(request, useRaw: false, cancellationToken);
```

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -294,10 +351,122 @@ public override async ValueTask<IReadOnlyList<BulkStateItem>> GetBulkStateAsync(
...
+                {
+                    // User has passed in raw bytes
+                    var requestBytes = (byte[])(object)(request.Body);
+                    serializedData = new Any { Value = ByteString.CopyFrom(requestBytes), TypeUrl = typeof(byte[]).FullName };
+                }
+                else
+                {
+                    serializedData = TypeConverters.ToAny(request.Body, this.jsonSerializerOptions);
+                }
+            }
```
rynowak, line 385: perfect

rynowak (approved):

Issue #416, opened 2020-10-07
Remove reliance on static instances in Actor infrastructure


rynowak:
## Expected Behavior

This is an issue tracking refactoring/design work internal to the Actor runtime. There may be bugs and UX gaps due to this issue but the expectation is that we just do some code cleanup.

## Actual Behavior

Dapr Actors relies on a number of **mutable** static objects in its infrastructure. These are effectively global variables and have all of the maintainability problems associated with mutable global variables.

We should refactor as needed to remove these. Having statics as part of the design makes it easy to design experiences that cannot scale horizontally as features are added. Ex: https://github.com/dapr/dotnet-sdk/issues/362

**High Priority:**

- [x] [ActorRuntime](https://github.com/dapr/dotnet-sdk/blob/6c271262231c41b21f3ca866eb0d55f7ce8b7dbc/src/Dapr.Actors/Runtime/ActorRuntime.cs#L23)
- [x] [ActorTrace](https://github.com/dapr/dotnet-sdk/blob/6c271262231c41b21f3ca866eb0d55f7ce8b7dbc/src/Dapr.Actors/ActorTrace.cs#L13) - this is used for logging and is not customizable today but *should* be made customizable
- [ ] [ActorProxyFactory](https://github.com/dapr/dotnet-sdk/blob/master/src/Dapr.Actors/Client/ActorProxy.cs#L23) - there's a PR out now that endeavors to make this mutable. If it stays immutable/stateless, we can keep the static.

**Low Priority:**

These cases are unlikely to have an impact and will be easy to mitigate in the future.

- Everything in `Dapr.Actors.Builder` - these act as caches for type information. It's possible as of .NET Core 2.X to unload assemblies and types. Using caches of type information that aren't "unload aware" will prevent unloading, which could considered a bug. However ASP.NET Core doesn't yet support this use case - so there would likely be no impact for us fixing it. 
- Everything in `Dapr.Actors.Communication` - these are caches for serialization functionality. This has the same justification as above - these are likely harmless.

rynowak:
Closing this as done. Changing ActorProxyFactory  would be really impactful without much benefit so we're not going to address that part right now.

PR #433, opened 2020-10-15
Remove static instance of ActorRuntime

rynowak:
# Description

This change removes a mutable static instance - this is low-hanging
fruit design improvement and helps us make these features more flexible
in the future.

## Issue reference

Contributes to: #416

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation (n/a)

### DIFF
--- a/src/Dapr.Actors.AspNetCore/WebHostBuilderExtensions.cs
+++ b/src/Dapr.Actors.AspNetCore/WebHostBuilderExtensions.cs
@@ -9,6 +9,7 @@ namespace Dapr.Actors.AspNetCore
     using Dapr.Actors.Runtime;
     using Microsoft.AspNetCore.Hosting;
     using Microsoft.Extensions.DependencyInjection;
+    using Microsoft.Extensions.Hosting;
 
     /// <summary>
     /// Class containing DaprActor related extension methods for Microsoft.AspNetCore.Hosting.IWebHostBuilder.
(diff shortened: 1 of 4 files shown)


src/Dapr.Actors.AspNetCore/WebHostBuilderExtensions.cs
```diff
@@ -36,7 +37,11 @@ public static IWebHostBuilder UseActors(this IWebHostBuilder hostBuilder, Action
                 return hostBuilder;
             }
 
-            configureActorRuntime.Invoke(ActorRuntime.Instance);
+            var runtime = new ActorRuntime();
+            if (configureActorRuntime != null)
+            {
+                configureActorRuntime.Invoke(runtime);
+            }
```
rynowak, line 44: this was a bug - it would have null-ref'ed

src/Dapr.Actors.AspNetCore/WebHostBuilderExtensions.cs
```diff
@@ -47,6 +52,8 @@ public static IWebHostBuilder UseActors(this IWebHostBuilder hostBuilder, Action
                 services.AddRouting();
                 services.AddHealthChecks();
                 services.AddSingleton<IStartupFilter>(new DaprActorSetupFilter());
+
+                services.AddSingleton<ActorRuntime>(runtime);
```
vinayada1, line 56: Isn't adding a singleton same as a static instance?

rynowak: No, it's not the same because it's only a singleton within the context of the DI container. You can't (from a static context) reach out and grab a singleton from DI - which solves all of the maintainability problems. 

The problem that this avoids is when multiple code paths want to configure something like the runtime - they have to do so by being passed an instance or an instance of a container. A true static means that you can't really audit or predict what code accesses what (it's a true global variable).

vinayada1: Thanks!

vinayada1 (dismissed):

rynowak:
I rebased it, if you want to take another quick look

vinayada1 (approved):

PR #439, opened 2020-10-17
Fixed NullReferenceException thrown from StateEntryApplicationModelProvider

tomasherceg:
# Description

The `StateEntryApplicationModelProvider` was throwing `NullReferenceException` from `MapControllers` when a parameter without model binding was present in any API method. The issue is the same as in #371 but it seems like it wasn't fixed on all places it was happening.

Dapr SDK version: `0.11.0-preview02` 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

I was not able to replicate the behavior in the test project on the `DaprController`. Even when I copied the exact controller methods that were causing the problem, I haven't run into the issue.

### DIFF
--- a/src/Dapr.AspNetCore/StateEntryApplicationModelProvider.cs
+++ b/src/Dapr.AspNetCore/StateEntryApplicationModelProvider.cs
@@ -28,7 +28,7 @@ namespace Dapr.AspNetCore
                     {
                         // Not bindable.
                     }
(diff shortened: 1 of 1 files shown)


vinayada1 (approved):

rynowak:
Thanks for sending this PR - but it looks like it grew stale despite being approved, I think we got multiple PRs for this issue and another one was merge first. Regardless, we appreciate the effort, thanks!

PR #440, opened 2020-10-17
Create gRPC serializer

akkie:
# Description

This pull request creates a gRPC serializer that can be injected into the dapr client builder. This fixes the issue that user code must implement the serialization/deserialization logic when building gRPC enabled APIs with dapr. The gRPC serializer can be configured by the user and then injected into all services. The dapr client builder will then pickup this implementation. If no gRPC serializer was configure, dapr will use the default serializer implementation.

The following test shows the current situation: https://github.com/dapr/dotnet-sdk/blob/master/test/Dapr.Client.Test/InvokeApiTest.cs#L498

If an application implements the AppCallback.AppCallbackBase class to receive gRPC requests, the application needs a way to serialize/deserialize the data send in the InvokeRequest and InvokeResponse classes. Currently the user must implement this logic itself, because the Dapr logic was only accessible internally. With this PR a user is able to inject the serializer implementation into the services and use the same logic as Dapr uses internally.

Unfortunately, the PR changes the API of the DaprClientBuilder, because the JSON options are now passed to the serializer which will be injected into the client builder. In my opinion the new implementation makes the code more cleaner, because the JSON options are now used at one place. Previously the options were scattered over the whole client implementation.

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


rynowak:
> This fixes the issue that user code must implement the serialization/deserialization logic when building gRPC enabled APIs with dapr. The gRPC serializer can be configured by the user and then injected into all services. The dapr client builder will then pickup this implementation. If no gRPC serializer was configure, dapr will use the default serializer implementation.

@akkie - can you please open an issue to discuss? Would like to know more about the use case and what's hard about it.

akkie:
@rynowak Please see #441 and also #442

rynowak:
@akkie - thanks!

rynowak:
Closing this as we're going to introduce a "serializer" abstraction in the Dapr .NET SDK as agreed here: https://github.com/dapr/dotnet-sdk/issues/442#issuecomment-714294726

Issue #436, opened 2020-10-16
[Unit Tests] dotnet-sdk: Add unit test code coverage reports to PRs


wcs1only:
## Describe the proposal

As part of dapr/dapr#2226, we need to be measuring unit test coverage with time. We should add codecov reports, along with rules against regressing unit test coverage.

PR #443, opened 2020-10-19
Added codecov integration (#436)

wcs1only:
# Description

Add codecov integration

## Issue reference

Please reference the issue this PR will close: #436

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/.github/workflows/sdk_build.yml
+++ b/.github/workflows/sdk_build.yml
@@ -27,10 +27,13 @@ jobs:
         uses: actions/setup-dotnet@v1
         with:
           dotnet-version: 3.1.301
+      - name: Test solution - release
+        run: |
+          unset GITHUB_ACTIONS #disable deterministic builds, just for test run. Deterministic builds break coverage for some reason
+          dotnet test test/test.sln --configuration release /p:CollectCoverage=true /p:CoverletOutputFormat=opencover
+          dotnet clean # remove the artifacts so they can be rebuild with deterministic builds enabled
       - name: Build solution - release
         run: dotnet build all.sln --configuration release
-      - name: Test solution - release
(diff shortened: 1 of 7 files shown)


youngbupark (approved):

tcnghia (approved):
Thanks for adding this

.github/workflows/sdk_build.yml
```diff
@@ -27,17 +27,22 @@ jobs:
         uses: actions/setup-dotnet@v1
         with:
           dotnet-version: 3.1.301
+      - name: Test solution - release
+        run: |
+          unset GITHUB_ACTIONS #disable deterministic builds, just for test run. Deterministic builds break coverage for some reason
```
mukundansundar, line 32: is this only for dotnet or does this need to be added for other repos also?

mukundansundar (approved):

.github/workflows/sdk_build.yml
wcs1only: There's an issue with the code coverage framework and deterministic builds
I tried the mitigation suggested in:
https://github.com/coverlet-coverage/coverlet/blob/master/Documentation/DeterministicBuild.md
but I couldn't get it to work correctly. This was the simplest mitigation that I could get to work.

rynowak: I had some discussion about this offline. I'm going to take a look and see if I can figure out a better fix in a short amount of time. If not then we'll ship this.

rynowak (commented):
Let's move forward with this for now.

The only downside to the workaround is a slower build by like 15 seconds.

Independently of this I'll follow up whether we *really* have deterministic builds working - this is something we care about because we need the supply chain to be verifiable.

Issue #446, opened 2020-10-21
Avoid using `async void` in server side programming and tests


rynowak:
## Expected Behavior

Async methods in our library code and tests avoid `async void` as a feature. 

## Actual Behavior

There are some places where `async void` is required - in very specialized cases and certain patterns in UI programming. For what we're building here, `async void` is almost a mistake.

- `async void` methods have no backpressure (no way to wait for completion)
- `async void` methods will **CRASH** the process on unhandled exception

I suspect that the usage of `async void` in our tests has been purely accidental.

## Steps to Reproduce the Problem

This is code cleanup in our unit tests. No library code has `async void` in it. Since these tests are working with in-memory bytes, while they are async code paths, no actual async behavior will be observed. If these tests were changed to use networking (E2E tests) they would have concurrency bugs 😆 

## Release Note

None: this is test code cleanup.

PR #447, opened 2020-10-21
Remove async void in tests

rynowak:
# Description

`async void` is almost always a bug in server-side programming. If these
tests were doing any networking then they would encounter concurrency
bugs or crash in the event of an exception.

## Issue reference

Fixes: #446 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


rynowak:
@vinayada1 - can I get a review on this test cleanup?

vinayada1 (approved):

Issue #422, opened 2020-10-09
InvokeMethodAsync reports cancellation using RpcException instead of OperationCanceledOperation


rynowak:
## Expected Behavior

InvokeMethodAsync and other APIs that accept a `CancellationToken` should report cancellation by the caller by throwing `OperationCanceledException`.

This behavior is important because application code that passes in a cancellation token expects to handle cancellation on a separate code path than other kinds of failures. Using the standard mechanism to report a canceled operation makes this straightforward.

## Actual Behavior

The DaprClient reports cancellation by propagating the underlying implementation's cancellation mechanism (`RpcException`). 

Example call-stack:

```txt
Error: Grpc.Core.RpcException: Status(StatusCode=Cancelled, Detail="Call canceled by the client.")
   at Dapr.Client.DaprClientGrpc.MakeGrpcCallHandleError[TResponse](Func`2 callFunc, Metadata headers, CancellationToken cancellationToken)
   at Dapr.Client.DaprClientGrpc.MakeInvokeRequestAsync(String appId, String methodName, Any data, HTTPExtension httpExtension, CancellationToken cancellationToken)
   at Dapr.Client.DaprClientGrpc.InvokeMethodAsync[TRequest,TResponse](String appId, String methodName, TRequest data, HTTPExtension httpExtension, CancellationToken cancellationToken)
   at Frontend.Shared.JoinGameScreen.TryJoinGameAsync() in /Users/ryan/github.com/rynowak/rock-paper-actors/src/Frontend/Shared/JoinGameScreen.razor:line 62
   at Frontend.Shared.JoinGameScreen.OnParametersSetAsync() in /Users/ryan/github.com/rynowak/rock-paper-actors/src/Frontend/Shared/JoinGameScreen.razor:line 30
   at Microsoft.AspNetCore.Components.ComponentBase.CallStateHasChangedOnAsyncCompletion(Task task)
   at Microsoft.AspNetCore.Components.ComponentBase.RunInitAndSetParametersAsync()
````

## Steps to Reproduce the Problem

Call `InvokeMethodAsync` and pass in a pre-cancelled cancellation token - or begin a long operation that hits the network and then cancel after a short period of time.

## Workaround

Example code that works around the problem:

```C#
            try
            {
                await client.InvokeMethodAsync<UserInfo, GameInfo>("matchmaker", "join", user, cancellationToken: cancellationToken);
            }
[...]

PR #455, opened 2020-11-03
Handle rpc exception thrown for operation cancellation

vinayada1:
# Description

Throw OperationCanceledException when rpc exception has status code = Cancelled

#422 

Please reference the issue this PR will close: #422 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -856,18 +860,25 @@ private Task<TResponse> MakeGrpcCallHandleError<TResponse>(Func<CallOptions, Asy
...
+                // add token for dapr api token based authentication
+                var daprApiToken = Environment.GetEnvironmentVariable("DAPR_API_TOKEN");
 
-            if (daprApiToken != null)
+                if (daprApiToken != null)
+                {
+                    callOptions.Headers.Add("dapr-api-token", daprApiToken);
+                }
+
+                // Common Exception Handling logic can be added here for all calls.
```
rynowak, line 875: looks like it has been already

test/Dapr.Client.Test/InvokeApiTest.cs
```diff
@@ -419,6 +420,47 @@ public async Task InvokeMethodAsync_WithReturnTypeAndData_WithNonDefaultVerb_Wit
...
+            invokeResponse.Data = TypeConverters.ToAny(data);
+            
+            var response = 
+                client.Call<InvokeResponse>()
+                .SetResponse(invokeResponse)
+                .Build();
+
+            const string rpcExceptionMessage = "Call canceled by client";
+            const StatusCode rpcStatusCode = StatusCode.Cancelled;
+            const string rpcStatusDetail = "Call canceled";
```
rynowak, line 438: We might not want to put these errors messages in tests because they are coming from the Grpc library. Libraries are free to change their error messages 😆 

Suggestions for improvement:

- don't test the error message
- OR add our own error message where we throw OperationCancelledException and test that

test/Dapr.Client.Test/InvokeApiTest.cs
```diff
@@ -419,6 +420,47 @@ public async Task InvokeMethodAsync_WithReturnTypeAndData_WithNonDefaultVerb_Wit
...
+            var ctSource = new CancellationTokenSource();
+            CancellationToken ct = ctSource.Token;
+            ctSource.Cancel();
+            try
+            {
+                Request request = new Request() { RequestParameter = "Hello " };
+                await client.DaprClient.InvokeMethodAsync<Request, Response>("test", "test", request, cancellationToken:ct);
+                Assert.False(true);
+            }
+            catch(OperationCanceledException ex)
```
rynowak, line 457: You can simplify this pattern: https://docs.shouldly.io/documentation/exceptions/throw

test/Dapr.Client.Test/InvokeApiTest.cs
```diff
@@ -419,6 +420,47 @@ public async Task InvokeMethodAsync_WithReturnTypeAndData_WithNonDefaultVerb_Wit
...
+            {
+                ex.Message.Should().Be(rpcExceptionMessage);
+                ((Grpc.Core.RpcException)ex.InnerException).Status.Equals(rpcStatus);
```
rynowak, line 460: nitpick: can also simplify this pattern: https://docs.shouldly.io/documentation/equality/oftype

rynowak (commented):
Some minor suggestions left - overall looks very thorough and correct

test/Dapr.Client.Test/InvokeApiTest.cs
```diff
@@ -419,6 +421,39 @@ public async Task InvokeMethodAsync_WithReturnTypeAndData_WithNonDefaultVerb_Wit
...
+            client.Mock
+                .Setup(m => m.InvokeServiceAsync(It.IsAny<Autogen.Grpc.v1.InvokeServiceRequest>(), It.IsAny<CallOptions>()))
+                .Throws(rpcException);
+
+            var ctSource = new CancellationTokenSource();
+            CancellationToken ct = ctSource.Token;
+            ctSource.Cancel();
+            Request request = new Request() { RequestParameter = "Hello " };
+            var task = client.DaprClient.InvokeMethodAsync<Request, Response>("test", "test", request, cancellationToken: ct);
+            (await FluentActions.Awaiting(async () => await task).Should().ThrowAsync<OperationCanceledException>()).WithInnerException<Grpc.Core.RpcException>();
```
rynowak, line 454: You should move line 453 inside the lambda here. This code is a little bit fragile - calling a task returning method and then awaiting it later is a pretty expert thing to do. 

The reason is that `client.DaprClient.InvokeMethodAsync(...)` might throw an exception immediately, or it might throw when you await the task. You want them both be covered by the test.

rynowak: If needed we can jump on a call and I can explain.

rynowak: note: this is the the case for all of the new tests

vinayada1: Thanks for taking the time to go over this with an example!

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -427,6 +427,10 @@ public override async ValueTask<IReadOnlyList<BulkStateItem>> GetBulkStateAsync(
                 return invokeResponse;
 
             }
+            catch (RpcException rpc) when (rpc.StatusCode == StatusCode.Cancelled && cancellationToken.IsCancellationRequested)
```
JamesNK, line 430: Why are you doing this instead of setting `GrpcChannelOptions.ThrowOperationCanceledOnCancellation` to true?

vinayada1: discussed with @rynowak again and used the solution that you suggested.

samples/Client/DaprClient/Readme.md
```diff
@@ -144,4 +144,14 @@ The controller sample has a route "/throwException" that returns a BadRequest re
                     Console.WriteLine("Grpc Exception: Http Status Code: " + rpcError.Metadata[daprErrorInfoHTTPCodeMetadata]);
                 }
             }
- ```
\ No newline at end of file
+ ```
+
+ ## Working with cancellation tokens
+
```
rynowak, line 150: Awesome 👍

rynowak (approved):

Issue #414, opened 2020-10-07
Dapr Actors should not use BinaryFormatter


rynowak:
## Expected Behavior

Dapr does not rely on deprecated features, or parts of .NET with known security vulnerabilities.

## Actual Behavior

Dapr uses BinaryFormatter - which **will throw an exception by default for web projects in .NET 5**. 

See: https://github.com/dotnet/designs/pull/141

This step is part of a long term plan to obsolete and remove BinaryFormatter from the framework. For .NET 5 it would be possible for a user to opt-in and allow it to function.

We should do the right thing and update any functionality that uses BinaryFormatter so that users aren't stung by this when they create their first .NET 5 Dapr project.

## Steps to Reproduce the Problem

Build a .NET 5 Dapr Actors Application

Trigger a codepath that uses `RemoteException`: https://github.com/dapr/dotnet-sdk/blob/ea51561f497dae80f411300f3399202715249330/src/Dapr.Actors/Communication/RemoteException.cs

rynowak:
## Ideas

We had a discussion with @grabyourpitchforks on our options here.

The major concern is that BinaryFormatter or any other mechanism that constructs arbitrary types can easily be used as a remote code execution vulnerability.  There's a proposal to build a general purpose *safe* serialization model for exceptions in .NET: https://github.com/dotnet/runtime/issues/43482 - unfortunately this is just a proposal and not something we can start using.

### Do nothing - continue using BinaryFormatter

The major drawback of this is that be broken by default in .NET 5. It doesn't improve the security situation and we would leave users in a broken state immediately when they create a new project.

.NET does not allow libraries to control feature switches like this because the application author needs to understand the security risks and make the decision. Changing the defaults for users is not something we can do.

### Throw a single ActorException type and don't serialize full exception details

This is the fallback case today when the server's exception type can't be serialized. 

We'd include the exception message and name of the original exception type so that it's still possible to handle different cases using `catch (...) when (...)`

### Try to implement our own serializer with limitations that avoid security flaws

Any approach to implement our own serializer will need to avoid numerous security risks to handle polymorphism correctly. We would need to build our own opt-in system for tracking which types can be serialized as well as our own allow-list of framework-provided exception types. Due to the size and amount of moving pieces this does not viable.

## Decision

Ultimately we decided on option 2 - to simplify what we're sending over the wire to the minimum and revisit this scenario when the general purpose exception serialization mechanism is available in .NET.

Issue #389, opened 2020-09-09
Add Integration tests to .net sdk repo


vinayada1:
## Expected Behavior

Add integration tests that can be run end-to-end on every PR.

## Actual Behavior

No end-to-end tests

vinayada1:
Especially actors

rynowak:
I added a project for in-memory integration tests as part of the DI pr - so we're unblocked for writing tests that don't use the Dapr sidecar. However it would be even better if we can include the interactions with `daprd` like the Java SDK does. We've definitely seen bugs where drift between our expectations and the behavior of `dapr` was the root cause.

rynowak:
Closing this because it has no actionable information. We'll add tests as we make changes and investments going forward.

PR #456, opened 2020-11-03
remove binaryformatter

vinayada1:
# Description

Remove use of BinaryFormatter

#414 

Please reference the issue this PR will close: #414 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


src/Dapr.Actors/Communication/RemoteException.cs
```diff
@@ -147,8 +171,24 @@ private static bool TryDeserializeException(Stream data, out Exception result)
...
-                result = (Exception)binaryFormatter.Deserialize(data);
-                return true;
+                using (var sr = new StreamReader(data, Encoding.UTF8))
+                {
+                    var str = sr.ReadToEnd();
+                    var dict = (Dictionary<string, object>)JsonSerializer.Deserialize(str, typeof(Dictionary<string, object>));
+                    var innerExceptionDict = (Dictionary<string, object>)JsonSerializer.Deserialize(dict[ExceptionInner].ToString(), typeof(Dictionary<string, object>));
+
+                    var innerExceptionType = Type.GetType(innerExceptionDict[ExceptionClassName].ToString());
+                    var innerException = (Exception)Activator.CreateInstance(innerExceptionType, innerExceptionDict[ExceptionMessage].ToString());
```
rynowak, line 181: This is the part of binary formatter that needs to be avoided. This allows a caller to send a type name over and HTTP request and we'll new up one of those. It's literally remote code execution.

What I expected to see was a design proposal for what the experience looks like when you subtract binary formatter. I don't think we should go down the path of trying to serialize arbitrary exception types. Normally (for ASP.NET Core) we would never send exception details like stack traces or inner exceptions over the wire.

src/Dapr.Actors/Communication/RemoteException.cs
```diff
@@ -136,28 +120,19 @@ internal static byte[] FromExceptionString(Exception exception)
                 CultureInfo.CurrentCulture,
                 SR.ErrorExceptionSerializationFailed2,
                 exception);
-            var exceptionData = new ServiceExceptionData(exception.GetType().FullName, exceptionStringBuilder.ToString());
+            return SerializeServiceException(exception);
+        }
+
+        internal static byte[] SerializeServiceException(Exception exception)
+        {
+            var exceptionData = new ServiceExceptionData(exception.GetType().FullName, exception.Message);
```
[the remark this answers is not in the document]
rynowak: Looks like these are existing types. The design of them is fine, but the naming isn't great. In particular `ServiceException` is super generic, and they are in a namespace that most users won't have referenced in code (`.Communication`).

src/Dapr.Actors/Communication/RemoteException.cs
```diff
@@ -13,10 +13,12 @@ namespace Dapr.Actors.Communication
     using System.Runtime.Serialization.Formatters;
     using System.Runtime.Serialization.Formatters.Binary;
     using System.Text;
+    using System.Text.Json;
     using System.Xml;
     using Dapr.Actors;
     using Dapr.Actors.Resources;
     using Microsoft.Extensions.Logging;
+    using System.Reflection;
```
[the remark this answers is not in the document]
vinayada1: fixed

src/Dapr.Actors/Communication/RemoteException.cs
```diff
@@ -57,10 +54,7 @@ public static (byte[], string) FromException(Exception exception)
         {
             try
             {
-                using var stream = new MemoryStream();
-                binaryFormatter.Serialize(stream, exception);
-                stream.Flush();
-                return (stream.ToArray(), string.Empty);
+                return (SerializeServiceException(exception), String.Empty);
```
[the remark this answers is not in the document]
vinayada1: in the exception case, we append these extra things which might be useful to the user:-
exceptionStringBuilder.AppendFormat(
                CultureInfo.CurrentCulture,
                SR.ErrorExceptionSerializationFailed1,
                exception.GetType().FullName);

            exceptionStringBuilder.AppendLine();

            exceptionStringBuilder.AppendFormat(
                CultureInfo.CurrentCulture,
                SR.ErrorExceptionSerializationFailed2,
                exception);

I have refactored the common code.

test/Dapr.Actors.Test/ActorMethodInvocationExceptionTests.cs
```diff
@@ -38,8 +39,10 @@ public void TestThrowActorMethodInvocationException()
             isDeserialzied.Should().BeTrue();
             var ex = this.ThrowRemoteException(message, remoteMethodException);
             ex.Should().BeOfType<ActorMethodInvocationException>();
-            ex.InnerException.Should().BeOfType<InvalidOperationException>();
+            ex.InnerException.Should().BeOfType<ServiceException>();
```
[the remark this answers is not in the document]
vinayada1: added

src/Dapr.Actors/Communication/ActorCommunicationException.cs
```diff
@@ -0,0 +1,134 @@
...
+    using System.Text;
+    using System.Xml;
+    using Dapr.Actors.Resources;
+    using Microsoft.Extensions.Logging;
+
+    /// <summary>
+    /// Provides an information about an exception from the service. This exception is thrown when the actual
+    /// exception from the service cannot be serialized for transferring to client.
+    /// </summary>
+    public class ActorCommunicationException : Exception
```
rynowak, line 21: Suggestion - `ActorInvokeException` or `ActorMethodException` - the term *Communication* makes it seem like a networking error, when it's actually not.

src/Dapr.Actors/Communication/ActorCommunicationException.cs
```diff
@@ -0,0 +1,134 @@
...
+    using System.IO;
+    using System.Runtime.Serialization;
+    using System.Text;
+    using System.Xml;
+    using Dapr.Actors.Resources;
+    using Microsoft.Extensions.Logging;
+
+    /// <summary>
+    /// Provides an information about an exception from the service. This exception is thrown when the actual
+    /// exception from the service cannot be serialized for transferring to client.
```
rynowak, line 19: > This exception is thrown when the actual exception from the service cannot be serialized for transferring to client.

Is this actually true for all cases?

src/Dapr.Actors/Communication/ActorCommunicationException.cs
```diff
@@ -0,0 +1,134 @@
...
+        /// Initializes a new instance of the <see cref="ActorCommunicationException"/> class.
+        /// </summary>
+        public ActorCommunicationException()
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ActorCommunicationException" /> class with appropriate message.
+        /// </summary>
+        /// <param name="actualExceptionType">the ActualExceptionType of exception thrown.</param>
```
rynowak, line 33:
```suggestion
        /// <param name="actualExceptionType">The fully-qualified name of the exception type throw by the actor.</param>
```

src/Dapr.Actors/Communication/ActorCommunicationException.cs
```diff
@@ -0,0 +1,134 @@
...
+        /// Gets the ActualExceptionType is the type of actual exception thrown.
+        /// </summary>
+        public string ActualExceptionType { get; private set; }
+
+        /// <summary>
+        /// Factory method that constructs the ActorCommunicationException from an exception.
+        /// </summary>
+        /// <param name="exception">Exception.</param>
+        /// <returns>Serialized bytes.</returns>
+        public static (byte[], string) FromException(Exception exception)
```
rynowak, line 52: Does this need to be public?

src/Dapr.Actors/Communication/ActorCommunicationException.cs
```diff
@@ -0,0 +1,134 @@
...
+        /// <param name="message">The error message that explains the reason for this exception.
+        /// </param>
+        public ActorCommunicationException(string actualExceptionType, string message)
+            : base(message)
+        {
+            this.ActualExceptionType = actualExceptionType;
+        }
+
+        /// <summary>
+        /// Gets the ActualExceptionType is the type of actual exception thrown.
```
rynowak, line 43: Similar suggestion to the above on the doc comment

src/Dapr.Actors/Communication/ActorCommunicationException.cs
```diff
@@ -0,0 +1,134 @@
...
+        /// </summary>
+        /// <param name="exception">Exception.</param>
+        /// <returns>Serialized bytes.</returns>
+        public static (byte[], string) FromException(Exception exception)
+        {
+            try
+            {
+                return (Serialize(exception), String.Empty);
+            }
+            catch (Exception e)
```
rynowak, line 58: Do we expect this to ever happen? Aren't we just serializing two strings?

src/Dapr.Actors/Communication/ActorCommunicationException.cs
```diff
@@ -0,0 +1,134 @@
...
+                CultureInfo.CurrentCulture,
+                SR.ErrorExceptionSerializationFailed1,
+                exception.GetType().FullName);
+
+            exceptionStringBuilder.AppendLine();
+
+            exceptionStringBuilder.AppendFormat(
+                CultureInfo.CurrentCulture,
+                SR.ErrorExceptionSerializationFailed2,
+                exception);
```
rynowak, line 100: what's the purpose of this string builder? it's contents aren't used?

src/Dapr.Actors/Communication/ActorCommunicationExceptionData.cs
```diff
@@ -0,0 +1,73 @@
...
+{
+    using System;
+    using System.IO;
+    using System.Runtime.Serialization;
+    using System.Xml;
+    using Dapr.Actors;
+    using Microsoft.Extensions.Logging;
+
+    [DataContract(Name = "ActorCommunicationExceptionData", Namespace = Constants.Namespace)]
+    internal class ActorCommunicationExceptionData
```
rynowak, line 16: Similar comments about naming to the exception type

src/Dapr.Actors/Communication/ActorCommunicationExceptionData.cs
```diff
@@ -0,0 +1,73 @@
...
+            return (ActorCommunicationExceptionData)ActorCommunicationExceptionDataSerializer.ReadObject(reader);
+        }
+
+        internal byte[] Serialize()
+        {
+            using var stream = new MemoryStream();
+            using var writer = XmlDictionaryWriter.CreateBinaryWriter(stream);
+            ActorCommunicationExceptionDataSerializer.WriteObject(writer, this);
+            writer.Flush();
+            return stream.ToArray();
```
rynowak, line 70: Why this serialization format? This is a binary-encoded XML format - why not something well known like JSON?

I'm not sure whether or not this is suitable for untrusted data - It's not an API I am familar with

src/Dapr.Actors/Communication/ActorCommunicationExceptionData.cs
```diff
@@ -0,0 +1,73 @@
...
+            this.Message = message;
+        }
+
+        [DataMember]
+        public string Type { get; private set; }
+
+        [DataMember]
+        public string Message { get; private set; }
+
+        internal static bool TryDeserialize(Stream data, out ActorCommunicationExceptionData result, ILogger logger = null)
```
rynowak, line 32: This is redundant with a similar method on ActorCommunicationException

test/Dapr.Actors.Test/ActorMethodInvocationExceptionTests.cs
```diff
@@ -25,21 +25,25 @@ public class ActorMethodInvocationExceptionTests
...
                                                      out var remoteMethodException);
             isDeserialzied.Should().BeTrue();
             var ex = this.ThrowRemoteException(message, remoteMethodException);
             ex.Should().BeOfType<ActorMethodInvocationException>();
-            ex.InnerException.Should().BeOfType<InvalidOperationException>();
+            ex.InnerException.Should().BeOfType<ActorCommunicationException>();
+            ((ActorCommunicationException)ex.InnerException).ActualExceptionType.Should().Be("System.InvalidOperationException");
+            ex.InnerException.InnerException.Should().BeNull();
             ex.Message.Should().Be(message);
+            ex.InnerException.Message.Should().Be(message);
```
rynowak, line 46: We should probably start adding integration tests for Actors - it might be a good thing to follow up on after this PR (which is already large).

src/Dapr.Actors/Communication/ActorCommunicationException.cs
```diff
@@ -0,0 +1,134 @@
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+// ------------------------------------------------------------
+
+namespace Dapr.Actors.Communication
```
rynowak, line 6: Is there a good reason to use this namespace? I don't think this is something we expect users to already ahve a `using` for.

Can we put this in the same namespace as all of the other classes that users use when calling into an Actor? It's better to keep related functionality together based on *usage*.

src/Dapr.Actors/Communication/ActorCommunicationException.cs
vinayada1: I will make this change with the other comments.

test/Dapr.Actors.Test/ActorMethodInvocationExceptionTests.cs
vinayada1: Marked this for milestone 3: https://github.com/dapr/dotnet-sdk/issues/389

src/Dapr.Actors/Communication/ActorInvokeException.cs
```diff
@@ -0,0 +1,97 @@
...
+        /// <returns>true if there was a valid exception, false otherwise.</returns>
+        internal static bool ToException(Stream bufferedStream, out Exception result)
+        {
+            // try to de-serialize the bytes in to exception requestMessage and create service exception
+            if (ActorInvokeException.TryDeserialize(bufferedStream, out result))
+            {
+                return true;
+            }
+
+            bufferedStream.Dispose();
```
rynowak, line 73: This is non-idiomatic. 

You generally don't dispose a stream that someone else gives you unless it's via a constructor. Think of this as the equivalent of passing a reference in C++. The caller owns the lifetime of the stream

src/Dapr.Actors/Communication/ActorCommunicationExceptionData.cs
rynowak: I'm fine with decision as long as we plan to address it before 1.0. 

Beyond issues like security, and configuration - there's more at stake here as well - cost to load and JIT additional types, memory consumption from the same, and binary size for anyone doing a trimmed publish.

src/Dapr.Actors/Communication/ActorInvokeExceptionData.cs
```diff
@@ -0,0 +1,52 @@
...
+namespace Dapr.Actors
+{
+    using System;
+    using System.IO;
+    using System.Runtime.Serialization;
+    using System.Xml;
+    using Dapr.Actors;
+    using Microsoft.Extensions.Logging;
+
+    [DataContract(Name = "ActorInvokeExceptionData", Namespace = Constants.Namespace)]
```
rynowak, line 15: As discussed, we'll get rid of DCS in a follow-up.

src/Dapr.Actors/Communication/ActorInvokeExceptionData.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        internal static ActorInvokeExceptionData Deserialize(Stream buffer)
+        {
+            if ((buffer == null) || (buffer.Length == 0))
+            {
+                return null;
+            }
+
+            using var reader = XmlDictionaryReader.CreateBinaryReader(buffer, XmlDictionaryReaderQuotas.Max);
+            return (ActorInvokeExceptionData)ActorInvokeExceptionDataSerializer.ReadObject(reader);
```
rynowak, line 40: As discussed offline - we'll change this to a serializer we're using elsewhere before 1.0 ships.

src/Dapr.Actors/Communication/ActorInvokeExceptionData.cs
```diff
@@ -0,0 +1,52 @@
...
+        public string Message { get; private set; }
+
+        internal static ActorInvokeExceptionData Deserialize(Stream buffer)
+        {
+            if ((buffer == null) || (buffer.Length == 0))
+            {
+                return null;
+            }
+
+            using var reader = XmlDictionaryReader.CreateBinaryReader(buffer, XmlDictionaryReaderQuotas.Max);
```
rynowak, line 39: You have to be careful here with usings on these readers. For instance I believe this class will dispose the underlying stream.

This isn't a problem if you're using memory stream, but with streams that actually do I/O you actually care about the lifetimes.

src/Dapr.Actors/Communication/ActorInvokeExceptionData.cs
```diff
@@ -0,0 +1,52 @@
...
+            this.Message = message;
+        }
+
+        [DataMember]
+        public string Type { get; private set; }
+
+        [DataMember]
+        public string Message { get; private set; }
+
+        internal static ActorInvokeExceptionData Deserialize(Stream buffer)
```
rynowak, line 32: I'd suggest some naming consistency for these parameters. `Stream stream` or `byte[] buffer`. It's a little bit confusing to see `Stream buffer` - a `Stream` represents a read/writeable thing with no guarantees about where the data is - whereas a `byte[]` is actual buffered memory.

src/Dapr.Actors/Communication/ActorInvokeExceptionData.cs
rynowak: This is one of the reasons why I want to limit the number and kind of serializers we're using. The S.T.Json API for instance is really simple, and doesn't bump into concerns like this.

rynowak (approved):

Issue #411, opened 2020-10-07
Multiple calls to UseActors should throw an exception or compose


rynowak:
## Expected Behavior

Calling `UseActors(...)` multiple times should either throw an exception, or compose by applying all of the provided configuration delegates.

## Actual Behavior

`UseActors(...)` will no op each time after the first call. This is a problem because if you call it multiple times you rely on the configuration delegates running. 

See: https://github.com/dapr/dotnet-sdk/blob/master/src/Dapr.Actors.AspNetCore/WebHostBuilderExtensions.cs#L34

## Steps to Reproduce the Problem

```C#
        public static IWebHostBuilder CreateWebHostBuilder(string[] args) =>
            WebHost.CreateDefaultBuilder(args)
                .UseStartup<Startup>()
            .UseActors(actorRuntime =>
            {
                actorRuntime.RegisterActor<SomeActor>();
            })
            .UseActors(actorRuntime =>
            {
                actorRuntime.RegisterActor<AnotherActor>();
            });
```

Only `SomeActor` will be registered `AnotherActor` will be forgotten 🤣 

This code is plausible if you think about how someone might abstract away these calls into their own helper functions in different "modules".

rynowak:
I suspect this was fixed during the logging changes. I'll take this one with the goal of adding a test for it.

PR #453, opened 2020-10-28
Logging changes to use ILogger

vinayada1:
# Description

Use ILogger instead of logging to console. The logging level can now be controlled by adding an appsettings.json file as below in the user actor project root. For example:-
{
    "Logging": {
      "Debug": {
        "LogLevel": {
          "Default": "Information"
        }
      },
      "Console": {
        "LogLevel": {
          "Microsoft": "Error",
          "Default": "Information"
        }
      },
      "LogLevel": {
        "Default": "Information",
        "Microsoft": "Error",
        "Microsoft.Hosting.Lifetime": "Error"
      }
    }
  }

## Issue reference

We strive to have all PR being opened based on an issue, where the problem or feature have been discussed prior to implementation.

Please reference the issue this PR will close: #406 

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [x] Code compiles correctly
* [x] Created/updated tests
* [ ] Extended the documentation

PR #459, opened 2020-11-04
Add test for multiple calls to UseActors

rynowak:
# Description

The bug was actually addressed in PR #453 - this change adds a test that
verifies that we actually fixed it.

## Issue reference

Fixes: #411

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


test/Dapr.Actors.AspNetCore.Test/ActorHostingTest.cs
```diff
@@ -0,0 +1,84 @@
...
+// Licensed under the MIT License.
+// ------------------------------------------------------------
+
+using System.Linq;
+using Dapr.Actors.Runtime;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.DependencyInjection;
+using Xunit;
+
+namespace Dapr.Actors.AspNetCore
```
vinayada1, line 12: .Test in the namespace like other tests?

rynowak: I didn't do that because it's not a useful practice. Putting tests in a different namespace than the library code just results in more usings. 

If you think it's important I can put it back, for consistency - but I think it's overall not a useful thing.

vinayada1: Just a consistency thing....I am fine either ways.

vinayada1 (approved):

Issue #171, opened 2019-11-24
Dependency injection for Actors


Xorpio:
Ass the basis of the Actor project is a .net core webapp we should be able to use the asp.net Dependecy Injection framework. However if you try to use it now the ActorService cannot create the Actor as it is not setup to create actors with other constructors.
see:
https://github.com/dapr/dotnet-sdk/blob/master/src/Dapr.Actors/Runtime/ActorService.cs#L43

amanbha:
@Xorpio When you register your actor, you can provide a Func to create Actor instances, it will be sued to create your actor instances, you can inject through this Func.
`actorRuntime.RegisterActor<DemoActor>(
                (type) =>
                new ActorService(type, (actorService, actorId) =>
                new DemoActor(actorService, actorId)));`

Xorpio:
Great! Missed that one. Works like a charm. Thanks!

neildonkin:
I like this approach, but how does it work if you want to inject an instance of `ILogger` into your actor's constructor?  How do you obtain an instance of ILogger within the `CreateHostBuilder` method?

guihlk:
> I like this approach, but how does it work if you want to inject an instance of `ILogger` into your actor's constructor? How do you obtain an instance of ILogger within the `CreateHostBuilder` method?

Hi @neildonkin,
I instead followed the steps provided here [https://github.com/dapr/dotnet-sdk/issues/255#issuecomment-624382344](https://github.com/dapr/dotnet-sdk/issues/255#issuecomment-624382344) in #255 and it worked. Basically you do not need to inject the ILogger from withing the CreateHostBuilder. It can simply turn into:
```
.UseActors(actorRuntime =>
                {
                })
```

And then, together with the actor registration, you define a factory. Not the ideal answer to your question, but maybe you end up achieving the same outcome?

The registration then ends up basically like:
. . .
`services.AddActor(MyActorFactory)`
. . .
```
private static MyActor MyActorFactory(ActorService service, ActorId id, IServiceProvider provider) =>
          new MyActor (
              service,
              id,
              provider.GetRequiredService<IMyCustomService>());
```
Just remember registering your custom services before adding them within the actor factory.

rynowak:
I'm reopening this issue - I was playing with this over the weekend and I didn't end up finding a solution that made me happy. We shouldn't require as much code as https://github.com/dapr/dotnet-sdk/issues/255#issuecomment-624382344 to solve this. 

[...]

PR #463, opened 2020-11-06
Implement dependency injection for Actors

rynowak:
# Description

This change allows Actors to accept dependency-injected services through
constructor parameters.

There are numerous breaking changes in this PR that are visible to user
code. In most cases these are simplifications.

For example the Actor base class accepted 3 constructor parameters,
which meant that every subclass (every actor type) also had 3 constructor
 parameters. This isn't fun and becomes way less fun now that users can
add their own constructor parameters in a meaningful way.

As another example, we had multiple places where "runtime options" lived
that had different interaction patterns. Now there's a single options
type and you interactive with it through `UseActors(...)`.

## Issue reference

Fixes: #171

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation

PR #484, opened 2020-11-19
First approach adding Dependency Injection to Actor activation

iskandersierra:
# Description

This is a first approach on adding dependency injection to `Dapr.Actors`. I just what to put a solution I came up today to discuss it, after that some code cleanup would be necessary and some unit tests would be required to validate this approach.

There are two API changes that are additive, hence backward compatible:
- Actor constructors MUST have parameters `ActorService service` and `ActorId actorId`, as before, but also MAY have any other parameters as long as their types are registered as services in the implicit `IServiceProvider` used to create the `ActorRuntime` (during the call to `services.UseActors(...)`)
- `ActorService` have a reference to the `IServiceProvider` instance, so that an actor type can have a factory function with access to the service provider through the `ActorService`, when using the method `RegisterActor`.

This propositions MUST be tested yet, but the changes are actual and backward compatible with existing test sets.

I modified existing tests (4) to check that my proposed changes do not affect existing code.

There are some changes that can be avoided, like adding a dependency to package `Microsoft.Extensions.Options` from `Dapr.Actors` project, I just wanted to have a tested solution and an approval to move this PR forward, to do the rest of the work.

## Issue reference

Please reference the issue this PR will close: #171

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -87,3 +87,4 @@ bld/
 
 # coverlet code coverage results
 coverage.json
+/.ionide
--- a/src/Dapr.Actors.AspNetCore/WebHostBuilderExtensions.cs
+++ b/src/Dapr.Actors.AspNetCore/WebHostBuilderExtensions.cs
@@ -59,7 +59,7 @@ namespace Dapr.Actors.AspNetCore
 
                 services.AddSingleton<ActorRuntime>(s =>
                 {
-                    return new ActorRuntime(s.GetRequiredService<IOptions<ActorRuntimeOptions>>().Value, s.GetRequiredService<ILoggerFactory>());
+                    return new ActorRuntime(s);
                 });
             });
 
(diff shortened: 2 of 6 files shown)


rynowak:
Hi @iskandersierra - there's a more complete PR for this issue here https://github.com/dapr/dotnet-sdk/pull/463 that implements the design proposal from the issue. Do you think there's anything missing there?

iskandersierra:
Hy @rynowak. Yes I noticed you were already working on this issue, in a more complete solution. For now I'm using a singleton of `IServiceProvider` to access services from my actors until this PR is merged. 

Thanks and keep the awesome work going. Let me know if I can help with something.

iskandersierra:
There is work done in this issue already. No need to proceed with this PR.

Issue #435, opened 2020-10-16
Timers invocation fails after app restarts


artursouza:
## Expected Behavior

After the app crashes (not Dapr sidecar), Actor timers should continue to be fired successfully.

## Actual Behavior

After the app crashes, the context of all registered timers is lost . See: https://github.com/dapr/dotnet-sdk/blob/26651756d0ca997218fa3a89b18b153326649b39/src/Dapr.Actors/Runtime/Actor.cs#L29

## Steps to Reproduce the Problem

Start the app that has an Actor.
Start Dapr sidecar.
Invoke a method in Actor that will register a timer.
See timer being logged in App.
Restarted App.
See exception when trying to trigger timer in App.

## Release Note

RELEASE NOTE: **FIX** Timer invocation after Actor runtime app restarts.

rynowak:
This will be an impactful change. The dotnet SDK allows you to register an arbitrary `Func<object, Task>` as a timer.

```
protected async Task<IActorTimer> RegisterTimerAsync(
            string timerName,
            Func<object, Task> asyncCallback,
            object state,
            TimeSpan dueTime,
            TimeSpan period)
{
}
```

There's no way that an arbitrary `Func<>` object can be serialized and restored from memory because it could be a closure holding on to arbitrary state.

Two approaches emerge:

- Require the callback to refer to a non-overloaded instance/static method declared on the actor type (or its base-class hierarchy) - this would be equivalent to registering a callback by name. This is the approach the Java takes.
- Requires the Actor class to associate timer names with functions during the `OnActivatedAsync` lifecycle method. No timers would be started at this point - only the associations between timer names and callbacks. The `RegisterTimerAsync` method would no longer accept a callback. The side effect would be that it its no longer possible to dynamically register a timer with a generated name.

Personally, I think the first option is a little bit more primitive, but it's simpler to reason about from a user's POV. The second options requires users to understand more about how timers work under the covers.

vinayada1:
@amanbha  - Are you aware of the history why we used delegates?

amanbha:
> @amanbha - Are you aware of the history why we used delegates?

Delegate seemed more user friendly, rather than just taking a method name and it follows Servic3e Fabric actor model.
Ideally, timers should not survive a restart as they are not persisted.

artursouza:
[...]

PR #485, opened 2020-11-20
Timers

vinayada1:
# Description

Remove in-memory storage for timers as the timer info is lost upon app restart

#435 

Please reference the issue this PR will close: #435 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


samples/Actor/DemoActor/DemoActor.cs
```diff
@@ -81,10 +82,23 @@ public Task ReceiveReminderAsync(string reminderName, byte[] state, TimeSpan due
...
         {
-            return this.RegisterTimerAsync("TestTimer", this.TimerCallBack, null, TimeSpan.FromSeconds(3), TimeSpan.FromSeconds(3));
+            var timerParams = new TimerParams
+            {
+                IntParam = 100,
+                StringParam = "timer test",
+            };
+
+            var state = JsonSerializer.SerializeToUtf8Bytes(timerParams);
+            return this.RegisterTimerAsync("TestTimer", "TimerCallBack", state, TimeSpan.FromSeconds(3), TimeSpan.FromSeconds(3));
```
rynowak, line 101: you can use `nameof(TimeCallBack)` here.

samples/Actor/DemoActor/DemoActor.cs
```diff
@@ -120,11 +134,15 @@ protected override Task OnDeactivateAsync()
         /// </summary>
         /// <param name="data">Timer input data.</param>
         /// <returns>A task that represents the asynchronous operation.</returns>
-        private Task TimerCallBack(object data)
+        public Task TimerCallBack(object data)
```
rynowak, line 137: nitpick: spelling - usually this would be `Callback` instead of `CallBack` because it's a single word when used as a programming term

src/Dapr.Actors/Runtime/Actor.cs
```diff
@@ -252,7 +242,7 @@ protected Task UnregisterReminderAsync(string reminderName)
         /// <summary>
         /// Registers a Timer for the actor. A timer name is autogenerated by the runtime to keep track of it.
         /// </summary>
-        /// <param name="asyncCallback">
+        /// <param name="callback">
```
rynowak, line 245: The text here needs to be updated as well to match the new behavior

src/Dapr.Actors/Runtime/Actor.cs
```diff
@@ -267,24 +257,20 @@ protected Task UnregisterReminderAsync(string reminderName)
...
         /// <returns>Returns IActorTimer object.</returns>
         protected Task<IActorTimer> RegisterTimerAsync(
-            Func<object, Task> asyncCallback,
+            string callback,
             object state,
             TimeSpan dueTime,
             TimeSpan period)
         {
-            return this.RegisterTimerAsync(null, asyncCallback, state, dueTime, period);
+            return this.RegisterTimerAsync(null, callback, Encoding.UTF8.GetBytes(state.ToString()), dueTime, period);
```
rynowak, line 265: `state.ToString()`

Unfortunately, this isn't idiomatic in .NET. There are APIs that accept an object `state` parameter, but the expectation is that they will give you the literal same object when your timer fires.  This doesn't follow that pattern.

Think about about the code that you'd write - you'd pass in a `Widget` here, and then your callback would have to accept bytes, not a `Widget`.

My suggestion is to just remove it this overload, and leave the `byte[]` one. That makes it clear that your data needs to be serialized/deserialized.

src/Dapr.Actors/Runtime/ActorManager.cs
```diff
@@ -185,18 +186,35 @@ async Task<byte[]> RequestFunc(Actor actor, CancellationToken ct)
             }
         }
 
-        internal Task FireTimerAsync(ActorId actorId, string timerName, CancellationToken cancellationToken = default)
+        internal async Task FireTimerAsync(ActorId actorId, string timerName, Stream requestBodyStream, CancellationToken cancellationToken = default)
         {
+            var timerData = await TimerInfo.DeserializeAsync(requestBodyStream);
+
+            dynamic awaitable;
```
rynowak, line 193: I suggest avoiding dynamic in all cases. It almost never is what you want.

src/Dapr.Actors/Runtime/ActorManager.cs
```diff
@@ -185,18 +186,35 @@ async Task<byte[]> RequestFunc(Actor actor, CancellationToken ct)
...
+
+                var parameters = methodInfo.GetParameters();
+                if (parameters.Length == 0)
+                {
+                    awaitable = methodInfo.Invoke(actor, null);
+                }
+                else
+                {
+                    var deserializedType = timerData.Data;
+                    awaitable = methodInfo.Invoke(actor, new object[] { deserializedType });
```
rynowak, line 210: Note, that this will blow up if the method returns void or anything other than `Task`.

I see a few options:
- Require the method to return `Task` and throw an exception explaining that it's required if it doesn't.
- Copy in https://github.com/dotnet/aspnetcore/tree/master/src/Shared/ObjectMethodExecutor and attribute it correctly (this code is Apache licensed, and the we need to retain the copyright info)
- Reverse engineer from https://github.com/dotnet/aspnetcore/blob/master/src/Shared/ObjectMethodExecutor/ObjectMethodExecutor.cs based on the 1-2 cases we need to handle (complicated)

I'd choose option 1.

src/Dapr.Actors/Runtime/ActorManager.cs
```diff
@@ -185,18 +186,35 @@ async Task<byte[]> RequestFunc(Actor actor, CancellationToken ct)
...
+
+            dynamic awaitable;
             // Create a Func to be invoked by common method.
             async Task<byte[]> RequestFunc(Actor actor, CancellationToken ct)
             {
-                await
-                    actor.FireTimerAsync(timerName);
+                var actorTypeName = this.actorService.ActorTypeInfo.ActorTypeName;
+                var actorType = this.actorService.ActorTypeInfo.ImplementationType;
+                MethodInfo[] methods = actorType.GetMethods(BindingFlags.Public | BindingFlags.Instance);
```
rynowak, line 199: this is dead code - I'd suggest passing the binding flags in at L200

src/Dapr.Actors/Runtime/ActorManager.cs
```diff
@@ -185,18 +186,35 @@ async Task<byte[]> RequestFunc(Actor actor, CancellationToken ct)
...
+                var methodInfo = actorType.GetMethod(timerData.Callback);
+
+                var parameters = methodInfo.GetParameters();
```
rynowak, line 202: We should probably harden this code. 

- Method return type
- Number and type of parameters
- Uniqueness (no overloading allowed)

src/Dapr.Actors/Runtime/ActorMethodContext.cs
```diff
@@ -40,6 +54,18 @@ public ActorCallType CallType
             get { return this.actorCallType; }
         }
 
+        /// <summary>
+        /// Gets the type of call by actor runtime (e.g. actor interface method, timer callback etc.).
```
rynowak, line 58: This is really right? It doesn't specify the type of call, it specifies the method being called.

src/Dapr.Actors/Runtime/ActorMethodContext.cs
```diff
@@ -14,10 +18,20 @@ public struct ActorMethodContext
...
             this.actorMethodName = methodName;
             this.actorCallType = callType;
+            this.callback = null;
+        }
+
+        private ActorMethodContext(Type callback, string methodName)
+        {
+            this.actorMethodName = "";
+            this.actorCallType = ActorCallType.ActorInterfaceMethod;
+            this.callback = callback.GetMethod(methodName);
```
rynowak, line 34: Is this really the right logic (`GetMethod)`? Most cases where reflection is happening on user code are a lot more specialized. For instance if this represents Actor interface method, this is definitely not the right logic.

It feels like this code needs a refactor to handle new capabilities - or maybe it shouldn't be shared between timers and actor methods.

src/Dapr.Actors/Runtime/ActorRuntime.cs
```diff
@@ -168,13 +168,14 @@ internal Task FireReminderAsync(string actorTypeName, string actorId, string rem
         /// <param name="actorTypeName">Actor type name to invokde the method for.</param>
         /// <param name="actorId">Actor id for the actor for which method will be invoked.</param>
         /// <param name="timerName">The name of timer provided during registration.</param>
+        /// <param name="requestBodyStream">The name of timer provided during registration.</param>
```
rynowak, line 171: Bad copy-paste of docs. I'm generally in favor of removing doc comments on internal code. They end up being boilerplate. It's better to just use comments to call out the truly important details.

src/Dapr.Actors/Runtime/ActorTimer.cs
```diff
@@ -18,49 +18,38 @@ internal class ActorTimer : IActorTimer
...
-            this.State = state;
-            this.Period = period;
-            this.DueTime = dueTime;
+            this.TimerInfo = timerInfo;
         }
 
         public string Name { get; }
 
-        public TimeSpan DueTime { get; }
+        internal TimerInfo TimerInfo { get; }
```
rynowak, line 30: Any good reason why this is internal visibility? The type is already internal, and now it has a mix of public and internal properties

src/Dapr.Actors/Runtime/IActorTimer.cs
```diff
@@ -13,12 +13,6 @@ namespace Dapr.Actors.Runtime
     /// </summary>
     public interface IActorTimer
     {
-        /// <summary>
-        /// Gets the time when timer is first due.
-        /// </summary>
-        /// <value>Time as <see cref="System.TimeSpan"/> when timer is first due.</value>
-        TimeSpan DueTime { get; }
```
rynowak, line 20: Do you know what this interface is used for? (I don't). If it's not something users use, then it should be removed or non-public.

src/Dapr.Actors/Runtime/TimerInfo.cs
```diff
@@ -0,0 +1,136 @@
...
+
+        public string Callback { get; private set; }
+
+        public TimeSpan DueTime { get; private set; }
+
+        public TimeSpan Period { get; private set; }
+
+        public byte[] Data { get; private set; }
+
+        internal static async Task<TimerInfo> DeserializeAsync(Stream stream)
```
rynowak, line 43: I'd suggest turning this functionality into a JsonConverter, and putting the `JsonConverterAttribute` on the `TimerInfo` class. You can keep all the same functionality, but it's composable.

rynowak (commented):
This seems like it's generally on the right track

src/Dapr.Actors/Runtime/Actor.cs
```diff
@@ -267,24 +257,20 @@ protected Task UnregisterReminderAsync(string reminderName)
         /// Specify negative one (-1) milliseconds to disable periodic signaling.</param>
         /// <returns>Returns IActorTimer object.</returns>
         protected Task<IActorTimer> RegisterTimerAsync(
-            Func<object, Task> asyncCallback,
+            string callback,
```
rynowak, line 260: Should we validate the method when the timer is created? Then you would find out you made a mistake immediately.

src/Dapr.Actors/Runtime/ActorManager.cs
vinayada1: Will go with option 1

src/Dapr.Actors/Runtime/ActorMethodContext.cs
vinayada1: removed....was unused

src/Dapr.Actors/Runtime/Actor.cs
vinayada1: Agree

samples/Actor/DemoActor/DemoActor.cs
```diff
@@ -100,13 +115,21 @@ protected override Task OnDeactivateAsync()
...
-        // This method is called when the timer is triggered based on its registration.
-        // It updates the PropertyA value.
-        private Task TimerCallBack(object data)
+        /// <summary>
+        /// This method is called when the timer is triggered based on its registration.
+        /// It updates the PropertyA value.
+        /// </summary>
+        /// <param name="data">Timer input data.</param>
+        /// <returns>A task that represents the asynchronous operation.</returns>
+        public Task TimerCallback(object data)
```
rynowak, line 124: did we talk about changing the signature of this to just be `byte[]`? that would make it really clear what the contract is.

src/Dapr.Actors.AspNetCore/EndpointRouteBuilderExtensions.cs
```diff
@@ -100,13 +103,14 @@ public static IEndpointConventionBuilder AddTimerRoute(this IEndpointRouteBuilde
             var runtime = endpoints.ServiceProvider.GetRequiredService<ActorRuntime>();
             return endpoints.MapPut("actors/{actorTypeName}/{actorId}/method/timer/{timerName}", async context =>
             {
+                // context.Request.EnableBuffering();
```
rynowak, line 106: delete :)

src/Dapr.Actors/Runtime/Actor.cs
```diff
@@ -285,24 +249,27 @@ protected Task UnregisterReminderAsync(string reminderName)
...
             if (string.IsNullOrEmpty(timerName))
             {
-                timerName = $"{this.Id}_Timer_{this.timers.Count + 1}";
+                timerName = $"{this.Id}_Timer_{Guid.NewGuid()}";
             }
 
-            var actorTimer = new ActorTimer(this, timerName, asyncCallback, state, dueTime, period);
-            var serializedTimer = await actorTimer.SerializeAsync();
+            var timerInfo = new TimerInfo(callback, callbackParams, dueTime, period);
+            var actorTimer = new ActorTimer(this, timerName, timerInfo);
```
rynowak, line 269: Is this line dead code now? It doesn't look like the value is used.

rynowak: NVM I'm a sillypants.

src/Dapr.Actors/Runtime/Actor.cs
```diff
@@ -328,9 +291,39 @@ protected async Task UnregisterTimerAsync(IActorTimer timer)
...
             await ActorRuntime.DaprInteractor.UnregisterTimerAsync(this.actorTypeName, this.Id.ToString(), timerName);
-            if (this.timers.ContainsKey(timerName))
+        }
+
+        private void ValidateTimerCallback(string callback)
+        {
+            var actorTypeName = this.Host.ActorTypeInfo.ActorTypeName;
+            var actorType = this.Host.ActorTypeInfo.ImplementationType;
+
+            // GetMethod will throw an AmbiguousMatchException if more than one methods are found with the same name
```
rynowak, line 301: Is the error message clear when that happens?

src/Dapr.Actors/Runtime/Actor.cs
```diff
@@ -328,9 +291,39 @@ protected async Task UnregisterTimerAsync(IActorTimer timer)
...
+            var methodInfo = actorType.GetMethod(callback, BindingFlags.Public | BindingFlags.Instance);
+
+            // Check if the method exists
+            if (methodInfo == null)
+            {
+                throw new ArgumentException($"Timer callback method: {callback} does not exist in the Actor class: {actorTypeName}");
```
rynowak, line 307: The other possibility is that it's not a public instance method.

src/Dapr.Actors/Runtime/Actor.cs
```diff
@@ -328,9 +291,39 @@ protected async Task UnregisterTimerAsync(IActorTimer timer)
...
+            }
+
+            // The timer callback should return only Task return type
+            if (methodInfo.ReturnType != typeof(Task))
+            {
+                throw new ArgumentException("Timer callback can only return type Task");
+            }
+
+            // The timer callback cannot be overridden in a derived class
+            if (methodInfo.IsVirtual)
```
rynowak, line 324: Why is that a problem?

src/Dapr.Actors/Runtime/ActorMethodContext.cs
```diff
@@ -40,6 +47,18 @@ public ActorCallType CallType
...
+        /// </summary>
+        /// <value>
+        /// An <see cref="Callback"/> representing the call type.
+        /// </value>
+        public MethodInfo Callback
+        {
+            get { return this.callback; }
+        }
+
+
```
rynowak, line 61: Are the changes to this file used? This seems like its from an earlier iteration

src/Dapr.Actors/Runtime/ActorRuntime.cs
```diff
@@ -156,19 +125,12 @@ internal Task FireReminderAsync(string actorTypeName, string actorId, string rem
...
-        /// <summary>
-        /// Fires a timer for the Actor.
-        /// </summary>
-        /// <param name="actorTypeName">Actor type name to invokde the method for.</param>
-        /// <param name="actorId">Actor id for the actor for which method will be invoked.</param>
-        /// <param name="timerName">The name of timer provided during registration.</param>
-        /// <param name="cancellationToken">The token to monitor for cancellation requests.</param>
-        /// <returns>A <see cref="Task"/> representing the result of the asynchronous operation.</returns>
-        internal Task FireTimerAsync(string actorTypeName, string actorId, string timerName, CancellationToken cancellationToken = default)
+        // Fires a timer for the Actor.
```
rynowak, line 128: Doesn't it feel great to free yourself from all this boilerplate :)

rynowak (commented):
The code changes look good overall - we should figure out why the method is not allowed to be non-virtual - that's an odd requirement.

src/Dapr.Actors/Runtime/Actor.cs
vinayada1: I caught the exception and threw with a more understandable error message

src/Dapr.Actors/Runtime/Actor.cs
vinayada1: I changed the binding flags to look for static and private methods as well which makes private method as a valid case

src/Dapr.Actors/Runtime/Actor.cs
vinayada1: Removed this check. I don't think we need to check the abstract flag as well since we will invoke this method on an instance.

src/Dapr.Actors/Runtime/Actor.cs
```diff
@@ -328,9 +291,46 @@ protected async Task UnregisterTimerAsync(IActorTimer timer)
         protected async Task UnregisterTimerAsync(string timerName)
         {
             await ActorRuntime.DaprInteractor.UnregisterTimerAsync(this.actorTypeName, this.Id.ToString(), timerName);
-            if (this.timers.ContainsKey(timerName))
+        }
+
+        internal MethodInfo GetMethodInfoUsingReflection(Type actorType, string callback)
+        {
+            return actorType.GetMethod(callback, BindingFlags.Public | BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.Static | BindingFlags.InvokeMethod);
```
rynowak, line 298: FYI - I think that `InvokeMethod` will be ignored here: https://docs.microsoft.com/en-us/dotnet/api/system.reflection.bindingflags?view=net-5.0

src/Dapr.Actors/Runtime/Actor.cs
```diff
@@ -328,9 +291,46 @@ protected async Task UnregisterTimerAsync(IActorTimer timer)
...
+        {
+            var actorTypeName = host.ActorTypeInfo.ActorTypeName;
+            var actorType = host.ActorTypeInfo.ImplementationType;
+
+            MethodInfo methodInfo = default;
+            try
+            {
+                methodInfo = this.GetMethodInfoUsingReflection(actorType, callback);
+            }
+            catch(AmbiguousMatchException)
```
rynowak, line 311: nit: fix spacing

src/Dapr.Actors/Runtime/Actor.cs
```diff
@@ -328,9 +291,46 @@ protected async Task UnregisterTimerAsync(IActorTimer timer)
...
+        {
+            return actorType.GetMethod(callback, BindingFlags.Public | BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.Static | BindingFlags.InvokeMethod);
+        }
+
+        internal void ValidateTimerCallback(ActorHost host, string callback)
+        {
+            var actorTypeName = host.ActorTypeInfo.ActorTypeName;
+            var actorType = host.ActorTypeInfo.ImplementationType;
+
+            MethodInfo methodInfo = default;
```
rynowak, line 306: nit: you don't need the initializer here (`MethodInfo methodInfo;` will do)

The compiler will verify that the variable is initialized before its accessed.

src/Dapr.Actors/Runtime/Actor.cs
```diff
@@ -328,9 +291,46 @@ protected async Task UnregisterTimerAsync(IActorTimer timer)
...
+            if (parameters.Length > 1)
+            {
+                throw new ArgumentException("Timer callback can accept only zero or one parameters");
+            }
+
+            // The timer callback should return only Task return type
+            if (methodInfo.ReturnType != typeof(Task))
             {
-                this.timers.Remove(timerName);
+                throw new ArgumentException("Timer callback can only return type Task");
```
rynowak, line 333: This is much improved compared to the original 👍

test/Dapr.Actors.Test/ITestActor.cs
```diff
@@ -65,5 +66,55 @@ public void ResetTestStateAsync()
...
+        {
+            return default;
+        }
+
+        protected Task TimerCallbackProtected()
+        {
+            return default;
+        }
+
+        protected Task TimerCallbackInternal()
```
rynowak, line 90: nope - this is `protected` :)

rynowak (approved):
There are 1-2 little things to fix here, overall looking great! 👍

src/Dapr.Actors/Runtime/Actor.cs
vinayada1: I guess so because we have restrictions on return type....so it anyways cannot be a constructor

test/Dapr.Actors.Test/ITestActor.cs
vinayada1: fixed

src/Dapr.Actors/Runtime/Actor.cs
rynowak: `GetMethod` only returns non-constructor methods regardless of that flag.

rynowak (approved):

Issue #421, opened 2020-10-09
HTTPExtension feedback


rynowak:
## Describe the proposal

The design of the HTTPExtension (for passing HTTP details to a Dapr Invoke) class could be improved

- The extra namespace segment `Dapr.Client.Http` should be removed - there's no need to hide these types in a separate namespace from the main client - they are part of the same logical task (invoking other services)
- The naming should reflect .NET practices - `HttpExtension` instead of `HTTPExtension` (golang convention)
- We should give this a name that reflects Dapr's terminology instead of generic `HTTPExtension` - suggesting `HttpInvocationOptions` - `HttpInvocationVerb`
- It would be better to drop our own type for HTTP verbs and use a framework one (`System.Net.Http.HttpMethod`). This is what users would expect to find.

rynowak:
Another reason to change this - these names collide with the generated gRPC types. This means that tooling will suggest importing the namespace with all of the generated code - picking the wrong one is an easy mistake to make.

PR #488, opened 2020-11-29
Rename and cleanup of HTTPExtension

rynowak:
##  Description

This change renames `HTTPExtension`->`HttpInvocationOptions`. This is more
idiomatic in .NET and reflects existing naming conventions (options or
settings for parameters).

Also changes to use the `System.Net.Http.HttpMethod` type for the http
method, and globally prefer the term `method` over `verb`. Every .NET
API that I'm aware of uses 'method', as does rfc7231.

## Issue reference

Fixes: #421

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -251,7 +263,7 @@ private static T ConvertFromInvokeBindingResponse<T>(Autogenerated.InvokeBinding
            string appId,
            string methodName,
            byte[] data,
-           HTTPExtension httpExtension = default,
+           HttpInvocationOptions httpExtension = default,
```
vinayada1, line 266: rename?

vinayada1 (dismissed):

src/Dapr.Client/DaprClientGrpc.cs
rynowak: yep! missed this one.

vinayada1 (approved):

Issue #491, opened 2020-12-01
ETag is empty


jjindrich:
Hi, I'm using State store (Azure Cosmos) in my application. When I call GetStateEntry, ETag is empty.
var state = await daprClient.GetStateEntryAsync<ArticleItem>(StoreName, vote.articleid)
state.ETag is empty

I need to avoid optimistic concurency. I need to get exception in case of another process updated entity.

I'm using this commands to save data:
var options = new StateOptions() {Concurrency = ConcurrencyMode.FirstWrite};
await state.SaveAsync(options);

RELEASE NOTE: rc1

rynowak:
> I'm using this commands to save data:

```
var options = new StateOptions() {Concurrency = ConcurrencyMode.FirstWrite};
await state.SaveAsync(options);
```

I realize that you mentioned the SDK didn't include an ETag - however when you *do* have the ETag you'll want use `TrySaveAsync` to ensure that it is passed to the state store.

jjindrich:
Thanks, I tried to use for saving data:
```
var options = new StateOptions() {Concurrency = ConcurrencyMode.FirstWrite};
await state.TrySaveAsync(options);
```
But the problem remain. Data are overriden.

vinayada1:
You might want to use the GetStateAndETagAsync API to get the state entry and the etag and as @rynowak suggested, TrySaveStateAsync API to save.

jjindrich:
@vinayada1 thanks for that. Can you help me how to use it ? state.TrySaveAsync doesn't exists
```
var state = await daprClient.GetStateAndETagAsync<MyItem>(StoreName, key);
...
var options = new StateOptions() {Concurrency = ConcurrencyMode.FirstWrite};
await state.TrySaveAsync(options);
```

vinayada1:
@jjindrich: Please take a look at the sample in the linked PR. I will check this in for future reference

jjindrich:
Excellent, many thanks !

PR #498, opened 2020-12-07
add sample for state apis with etags

vinayada1:
# Description

Add sample code to use State APIs with etags

#491 

Please reference the issue this PR will close: #491 
## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


samples/Client/DaprClient/Program.cs
```diff
@@ -169,6 +172,29 @@ internal static async Task ExecuteStateTransactionAsync(DaprClient client)
             Console.WriteLine("Executed State Transaction!");
         }
 
+         internal static async Task ReadAndSaveStateWithEtagsAsync(DaprClient client)
+        {
```
rynowak, line 176: nit: spacing is off here

samples/Client/DaprClient/Program.cs
```diff
@@ -169,6 +172,29 @@ internal static async Task ExecuteStateTransactionAsync(DaprClient client)
...
+
+            origRetrievedState.Color = "orange";
+            var updatedState = await client.TrySaveStateAsync<Widget>(storeName, stateKeyName, origRetrievedState, origEtag);
+            Console.WriteLine($"Saved modified state with etag: {origEtag} successfully? {updatedState}");
+
+             var (updatedRetrievedState, updatedEtag) = await client.GetStateAndETagAsync<Widget>(storeName, stateKeyName);
+            Console.WriteLine($"Retrieved state: {updatedRetrievedState.Size}  {updatedRetrievedState.Color} with etag: {updatedEtag}");
+
+            updatedRetrievedState.Color = "purple";
+            var isSaveStateSuccess = await client.TrySaveStateAsync<Widget>(storeName, stateKeyName, updatedRetrievedState, origEtag);
```
rynowak, line 191: I feel like this needs some comments for explanation.

One of the challenges for sample code is that you want to show how stuff works (and sometimes doesn't) but users will often just copy-paste code and use it verbatim, which is not something you want if you're providing a "bad example".

Can you add some comments that explain that this is intended to *simulate* what happens when you have a concurrency violation.

rynowak (approved):

Issue #510, opened 2020-12-16
DaprClient: Implement GetBulkSecretAsync


cmendible:
## Describe the proposal

Since dapr/dapr#1493 is fixed and both dapr/components-contrib#310 and dapr/dapr#2419 are merged we should implement the `GetBulkSecretAsync` operation to get all the secrets in a single call. 

``` csharp
        public abstract ValueTask<Dictionary<string, string>> GetBulkSecretAsync(
            string storeName,
            Dictionary<string, string> metadata = default,
            CancellationToken cancellationToken = default);
```

Also the `DaprSecretStoreConfigurationProvider` and corresponding extension methods should also allow for the application to retrieve all secrets on startup using `GetBulkSecretAsync`

``` csharp
        public static IConfigurationBuilder AddDaprSecretStore(
            this IConfigurationBuilder configurationBuilder,
            string store,
            Dictionary<string, string> metadata,
            DaprClient client)
        {
            ...

            configurationBuilder.Add(new DaprSecretStoreConfigurationSource()
            {
                Store = store,
                Metadata = metadata,
                Client = client
            });

            return configurationBuilder;
        }
```

PR #511, opened 2020-12-16
DaprClient: Implementing GetBulkSecretAsync

cmendible:
# Description

Implemented `GetBulkSecretAsync`

## Issue reference

Please reference the issue this PR will close: #510

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


amolenk:
Very nice if we don't need to specify the secret keys upfront anymore for the configuration provider!

Do we still need the `AddDaprSecretStore(this IConfigurationBuilder configurationBuilder, string store, IEnumerable<DaprSecretDescriptor> secretDescriptors, DaprClient client)` overload though? If you don't want your application to have access to all the secrets, the preferred way would be to use the Dapr configuration file to allow/deny access to specific secret keys. That would also ensure that the application cannot get the secrets by sneaking around the SDK and making direct HTTP calls to the sidecar. The current behaviour of the GetBulkSecret API is blocking this, though (see https://github.com/dapr/dapr/issues/2600)

Also, does this implementation automatically refresh the secrets just like the Azure Key Vault config provider does? #wishlist :-)

cmendible:
Hi @amolenk for the time being I would keep the `AddDaprSecretStore(this IConfigurationBuilder configurationBuilder, string store, IEnumerable<DaprSecretDescriptor> secretDescriptors, DaprClient client)` overload because most of the secret providers require at least 1+ n trips to get the secret values, and I think that in some cases where you do not need all the secrets and performance is key the `secretDescriptors` will be helpful. 

Regarding https://github.com/dapr/dapr/issues/2600 I respected the way it was implemented in the `GetSecret` API but I agree that a better solution will be to return just the secrets that are allowed for the application. I'll add some comments on the issue.

And there is no automatic refresh yet so yes I guess it is on the `whishlist` 😁

rynowak:
Sorry for the delay on this, I've been fairly busy over the winter holidays. Assigning myself so I don't forget.

src/Dapr.Client/DaprClient.cs
```diff
@@ -324,5 +324,17 @@ public async ValueTask<StateEntry<TValue>> GetStateEntryAsync<TValue>(string sto
             string key,
             Dictionary<string, string> metadata = default,
             CancellationToken cancellationToken = default);
+
+        /// <summary>
+        /// Gets all secret values from the secret store.
+        /// </summary>
+        /// <param name="storeName">Secret store name.</param>
+        /// <param name="metadata">An key/value pair that may be consumed by the secret store.  This depends on the secret store used.</param>
```
rynowak, line 332: Can you update this to be the same as the other `metadata` parameters on this class? We did a review pass here after you created this PR.

src/Dapr.Client/DaprClient.cs
```diff
@@ -324,5 +324,17 @@ public async ValueTask<StateEntry<TValue>> GetStateEntryAsync<TValue>(string sto
...
+        /// <param name="cancellationToken">A <see cref="CancellationToken" /> that can be used to cancel the operation.</param>
+        /// <returns>A <see cref="ValueTask{T}" /> that will return the value when the operation has completed.</returns>
+        public abstract ValueTask<Dictionary<string, string>> GetBulkSecretAsync(
```
rynowak, line 335: Should this be `GetAllSecretsAsync()`? It's not doing any filtering, it's getting all of the ones you have access to.

src/Dapr.Client/DaprClient.cs
```diff
@@ -324,5 +324,17 @@ public async ValueTask<StateEntry<TValue>> GetStateEntryAsync<TValue>(string sto
             string key,
             Dictionary<string, string> metadata = default,
             CancellationToken cancellationToken = default);
+
+        /// <summary>
+        /// Gets all secret values from the secret store.
```
rynowak, line 329:
```suggestion
        /// Gets all secret values that the application is allowed to access from the secret store.
```

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -836,6 +836,28 @@ private Autogenerated.StateItem ToAutogeneratedStateItem(StateTransactionRequest
...
             return response.Data.ToDictionary(kv => kv.Key, kv => kv.Value);
         }
+
+        /// <inheritdoc/>
+        public async override ValueTask<Dictionary<string, string>> GetBulkSecretAsync(
+            string storeName,
+            Dictionary<string, string> metadata = default,
+            CancellationToken cancellationToken = default)
+        {
+            var envelope = new Autogenerated.GetBulkSecretRequest(){
```
rynowak, line 846: move brace to new line

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -836,6 +836,28 @@ private Autogenerated.StateItem ToAutogeneratedStateItem(StateTransactionRequest
 
             return response.Data.ToDictionary(kv => kv.Key, kv => kv.Value);
         }
+
+        /// <inheritdoc/>
+        public async override ValueTask<Dictionary<string, string>> GetBulkSecretAsync(
```
rynowak, line 841: Is the return value correct here? `GetSecretAsync` returns `ValueTask<Dictionary<string, string>>`, so I would expect this to be `ValueTask<Dictionary<string, Dictionary<string, string>>>`

src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationExtensions.cs
```diff
@@ -44,6 +44,34 @@ public static class DaprSecretStoreConfigurationExtensions
...
+        /// </summary>
+        /// <param name="configurationBuilder">The <see cref="IConfigurationBuilder"/> to add to.</param>
+        /// <param name="store">Dapr secret store name.</param>
+        /// <param name="metadata">The secrets metadata.</param>
+        /// <param name="client">The Dapr client</param>
+        /// <returns>The <see cref="IConfigurationBuilder"/>.</returns>
+        public static IConfigurationBuilder AddDaprSecretStore(
+            this IConfigurationBuilder configurationBuilder,
+            string store,
+            Dictionary<string, string> metadata,
```
rynowak, line 58: It would be really great if we could make metadata optional here - since usually there's not going to be anything in there. If this is a required arg, then everyone has to stop, look it up, figure out they don't care, and then move with their life :)

src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationSource.cs
```diff
@@ -24,6 +26,11 @@ public class DaprSecretStoreConfigurationSource : IConfigurationSource
         /// </summary>
         public IEnumerable<DaprSecretDescriptor> SecretDescriptors { get; set; }
 
+        /// <summary>
+        /// Gets or sets the secrets metadata.
```
rynowak, line 30: Can you copy a similar description for this from the updated version of DaprClient.cs?

src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationSource.cs
```diff
@@ -32,7 +39,19 @@ public class DaprSecretStoreConfigurationSource : IConfigurationSource
         /// <inheritdoc />
         public IConfigurationProvider Build(IConfigurationBuilder builder)
         {
-            return new DaprSecretStoreConfigurationProvider(Store, SecretDescriptors, Client);
+            if (SecretDescriptors != null)
+            {
+                if (Metadata != null)
+                { 
+                    throw new ArgumentException($"{nameof(Metadata)} must be null when {nameof(SecretDescriptors)} is set", nameof(Metadata));
```
rynowak, line 46: Is this really true? I don't think these are exclusive in the Dapr API

rynowak:
Do we still have issues here with waiting until the sidecar starts? Does that need to be addressed or does HttpClient/gRPC do enough retries?

rynowak:
This is overall looking pretty good. It would be great to complete this and get it merged within the next week if possible.

src/Dapr.Client/DaprClient.cs
cmendible: The `GetBulk` prefix was added to keep consistency with naming used by Dapr: https://github.com/dapr/components-contrib/pull/310#discussion_r504195884

src/Dapr.Client/DaprClient.cs
cmendible: done

src/Dapr.Client/DaprClient.cs
cmendible: done

src/Dapr.Client/DaprClientGrpc.cs
cmendible: done

src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationExtensions.cs
cmendible: Great point 👍

Done!

src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationSource.cs
cmendible: done

src/Dapr.Client/DaprClientGrpc.cs
cmendible: This is an eye opener. They way I implemented this in **components-contrib** is returning the `map[string]string`, so for instance when using [kubernetes](https://github.com/dapr/components-contrib/blob/4d3e1be105ef2c10f64d6b80f87341244a6751e0/secretstores/kubernetes/kubernetes.go#L78) the store is only returning the underlying secret maps and it's ignoring the original secret name.

I guess all this means that we'll have to revisit both the **dapr** and **components-contrib** methods and Bulk Secrets API in order to return the expected information.

src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationSource.cs
cmendible: `Metadata` parameter is only useful when using the bulk secret API, this means that if descriptors (Each one holds it's own metadata) are specified the parameter is not used.

That's the reason behind the 2 constructors in `DaprSecretStoreConfigurationProvider`.

cmendible:
> Do we still have issues here with waiting until the sidecar starts? Does that need to be addressed or does HttpClient/gRPC do enough retries?

This is working as expected in the provided sample and I also removed the code that waited for Dapr. On the other hand there could be an issue when running inside kubernetes, but I think the provider should not be responsible for this check.

src/Dapr.Client/DaprClientGrpc.cs
rynowak: So if I'm understanding this right - the implementation of "Bulk Secrets" currently will - flatten all of the secrets into a single map, which is why the API looks like it does.

Also based on the code you linked, it looks like it ignores duplicates when flattening.

rynowak: Did you create an issue for this?

src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationSource.cs
rynowak: I guess what I'm really asking is this. **Both** the get secret and get bulk secret APIs in Dapr support metadata
- https://github.com/dapr/dapr/blob/master/dapr/proto/runtime/v1/dapr.proto#L238
- https://github.com/dapr/dapr/blob/master/dapr/proto/runtime/v1/dapr.proto#L257

So why are we modeling the .NET SDK around a different feature set - it seems like we should support metadata for both kinds of requests.

src/Dapr.Client/DaprClientGrpc.cs
cmendible: I'll create an issue in both the components and dapr repos to change the behavior and then be able to return `ValueTask<Dictionary<string, Dictionary<string, string>>>`  as expected here.

Agree?

rynowak: sounds good, let's start a discussion there - I don't claim to know everything, but the current design is not what I expected.

Also note, we just changed to use `Task` so when you do another round `ValueTask<>` -> `Task<>`

src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationSource.cs
cmendible: Not sure I understand your question. It does allow for metadata to be used in both kind of requests. When using descriptors the api allow to specify metadata for each  requested secret, via it´s descriptor) since the underlying API is GetSecrets. Then if you just want all the secrets (using GetBulSecrets) you can specify the metadata for that call if needed.

Perhaps the confusion is that current implementation  allows the use of both APIs depending on the parameters.

src/Dapr.Client/DaprClientGrpc.cs
cmendible: Waiting for https://github.com/dapr/components-contrib/issues/590 and https://github.com/dapr/dapr/issues/2652 to be fixed.

src/Dapr.Client/DaprClient.cs
```diff
@@ -365,5 +365,17 @@ public async Task<StateEntry<TValue>> GetStateEntryAsync<TValue>(string storeNam
             string key,
             IReadOnlyDictionary<string, string> metadata = default,
             CancellationToken cancellationToken = default);
+
+        /// <summary>
+        /// Gets all secret values that the application is allowed to access from the secret store.
+        /// </summary>
+        /// <param name="storeName">Secret store name.</param>
+        /// <param name="metadata">A key/value pair that may be consumed by the secret store. This depends on the secret store used.</param>
```
rynowak, line 373:
```suggestion
        /// <param name="metadata">A collection of metadata key-value pairs that will be provided to the secret store. The valid metadata keys and values are determined by the type of secret store used.</param>
```

We did a polish pass on these docs recently, this is similar to the text we're going with for metadata paramters

src/Dapr.Client/DaprClient.cs
```diff
@@ -365,5 +365,17 @@ public async Task<StateEntry<TValue>> GetStateEntryAsync<TValue>(string storeNam
...
+        /// <param name="cancellationToken">A <see cref="CancellationToken" /> that can be used to cancel the operation.</param>
+        /// <returns>A <see cref="Task{T}" /> that will return the value when the operation has completed.</returns>
+        public abstract Task<Dictionary<string, string>> GetBulkSecretAsync(
+            string storeName,
+            Dictionary<string, string> metadata = default,
```
rynowak, line 378: Use `IReadOnlyDictionary<>` here (recent polish pass)

rynowak: Dictionary is fine as the return type

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -919,6 +919,29 @@ private Autogenerated.StateItem ToAutogeneratedStateItem(StateTransactionRequest
 
             return response.Data.ToDictionary(kv => kv.Key, kv => kv.Value);
         }
+
+        /// <inheritdoc/>
+        public async override Task<Dictionary<string, string>> GetBulkSecretAsync(
```
rynowak, line 924: Waiting on update to the signature based on runtime changes.

src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationExtensions.cs
```diff
@@ -44,6 +44,33 @@ public static class DaprSecretStoreConfigurationExtensions
...
+        /// <param name="configurationBuilder">The <see cref="IConfigurationBuilder"/> to add to.</param>
+        /// <param name="store">Dapr secret store name.</param>
+        /// <param name="metadata">The secrets metadata.</param>
+        /// <param name="client">The Dapr client</param>
+        /// <returns>The <see cref="IConfigurationBuilder"/>.</returns>
+        public static IConfigurationBuilder AddDaprSecretStore(
+            this IConfigurationBuilder configurationBuilder,
+            string store,
+            DaprClient client,
+            Dictionary<string, string> metadata = null)
```
rynowak, line 59: Use `IReadOnlyDictionary<>` here and copy the doc string for `metadata` from elsewhere

src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationSource.cs
rynowak: OK makes sense 👍

rynowak (approved):

rynowak:
Thanks @cmendible

Issue #426, opened 2020-10-12
TrySaveStateAsync and TryDeleteStateAsync should only handle "rejections" based on ETag, not all RPC exceptions


rynowak:
## Expected Behavior

`TrySaveStateAsync` or `TryDeleteStateAsync` should return false **iff** the operation was rejected due to a concurrency violation. The design intent of these methods is to treat expected failures (etag mismatch) separately from unexpected failures (network hiccup).

Users should be able to use the true/false return value to distinguish whether state was saved, and only use try/catch if they choose to handle arbitrary exceptions.

## Actual Behavior

`TrySaveStateAsync` swallows **all** `RpcException`s

`TryDeleteStateAsync` swallows **all** `Exceptions`

## Steps to Reproduce the Problem

Issue a call to `TryDeleteState` that's totally invalid (say an invalid key format), or when the network is disconnected. It will be surfaced as a `false` return value instead of an exception.

rynowak:
Tagging this as a breaking change (not sure why it was missing). Moving from one valid (but not ideal) to another valid behavior is a breaking change 😆

PR #531, opened 2021-01-07
do not swallow all exceptions

vinayada1:
# Description

TrySaveStateAsync and TryDeleteStateAsync APIs should not swallow all exceptions

#426 

Please reference the issue this PR will close: #426 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -588,7 +588,7 @@ public override async ValueTask<(TValue value, string etag)> GetStateAndETagAsyn
                 await this.MakeSaveStateCallAsync(storeName, key, value, etag, stateOptions, metadata, cancellationToken);
                 return true;
             }
-            catch (RpcException)
+            catch (RpcException rpc) when (rpc.Status.Detail.Contains(ErrStateSave))
```
rynowak, line 591: We need to find a better way to do this. These strings are intended for human consumption and will change in the future. These should be programmatic error codes not human readable text.

vinayada1: I agree but I guess that would need changes in the runtime which might affect all sdks.

rynowak: Let's start a discussion about it then. I don't think this is a viable approach.

amolenk: If we only want `TrySaveStateAsync` to return false if there's an OCC conflict (which I think is the intended behavior), then the `ErrStateSave` string won't even be precise enough to determine what happened. Any error returning from a state component while saving keys will be surfaced as an ERR_STATE_SAVE message.
I think the only way to really solve this is to implement https://github.com/dapr/dapr/issues/2619

vinayada1: I tested this with redis and Dapr runtime seems to be getting the error below from redis component. I checked the component code and this seems to be what the redis is returning:-
== DAPR == @@@@@@ state err: failed to set key DaprClient||mykey: ERR Error running script (call to f_83e03ec05d6a3b6fb48483accf5e594597b6058f): @user_script:1: user_script:1: failed to set key DaprClient||mykey

Not very useful :(

rynowak: yeah I dug into the code in components contrib, and the runtime doesn't make a distinction. Any error returned by the state store is treated the same. Let's discuss this using dapr/dapr#2619

rynowak: https://github.com/dapr/dapr/pull/2647/files#diff-ce3f68305f0eb28b6b8911c281186536c80d8cac9dff3cece2c8bdf3dd955315R499

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -656,7 +656,7 @@ public override async Task<IReadOnlyList<BulkStateItem>> GetBulkStateAsync(strin
                 await this.MakeSaveStateCallAsync(storeName, key, value, etag, stateOptions, metadata, cancellationToken);
                 return true;
             }
-            catch (RpcException)
+            catch (RpcException rpc) when (rpc.StatusCode == StatusCode.Aborted && (rpc.Status.Detail.Contains(ETagInvalid) || rpc.Status.Detail.Contains(ETagMismatch)))
```
rynowak, line 659: We should not be comparing error messages here - checking the status code is enough.

Remember, these are user-facing descriptive error messages. If we update them to change the phrasing we don't want this code to break. Using the status code is enough 😄

vinayada1: That's why I am not comparing the entire error string but I thought the constants should not change. I hope Aborted error code does not get used for something else because it still sounds pretty generic. Thoughts?

rynowak: If a string in code is readable as a sentence (like these are) then it's for humans. Strings for computers look like `ERR_STATE_CONFLICT` or something similar.

The gRPC Aborted status code has a **very** specific meaning, and it's what we should rely on. The name sounds generic but it has a precise meaning that maps to our use case.  https://grpc.github.io/grpc/core/md_doc_statuscodes.html

vinayada1: ok...makes sense

rynowak (approved):

Issue #434, opened 2020-10-16
Actors HTTP handlers don't work with some common frontend server patterns


rynowak:
## Expected Behavior

Actors HTTP handlers that are registered by the runtime can be called from the Dapr sidecar without much user knowledge.

## Actual Behavior

Certain use cases for routing don't work with the way Actors is wired up. This requires a little bit of explanation to make sense, but I assure you it does 😆 

It's important to understand how the actor runtime communicates with the Dapr sidecar - it injects HTTP handlers into the application in way that *tries* to be transparent to the developer. The tool that's used is `IStartupFilter`- specifically to inject an additional router into the HTTP pipeline (at the end).

Using Actors in ASP.NET Core is **as-if** the user wrote:

```C#
            app.UseRouting();
            app.UseEndpoints(endpoints =>
            {
                // application endpoints go here
            });

            // this is not a mistake, routing is configured again
            app.UseRouting();
            app.UseEndpoints(endpoints =>
            {
                  // actors goes here
            });
```

This is how Actors tries to be transparent to the user. This breaks when the user configures a *route of last resort* - for instance in a server-side blazor project.

This is a common routing setup when an application is hosting a client-side SPA - the application handles URLs that don't match a "server side" route by passing them off to the endpoint that serves the SPA. This allows the client-side framework to handle deep linking, and also use all of it's client-side goodies for "not found" pages.

```C#
            app.UseRouting();
            app.UseEndpoints(endpoints =>
            {
                endpoints.MapSubscribeHandler();
                endpoints.MapHealthChecks("/healthz");
                endpoints.MapControllers();
                endpoints.MapBlazorHub();
                endpoints.MapFallbackToPage("/_Host"); // YOU SHALL NOT PASS
            });

            // this is not a mistake, routing is configured again
            app.UseRouting();
            app.UseEndpoints(endpoints =>
            {
                  // actors goes here
            });
```

There isn't really a good solution users can apply (without a bug fix) other than hosting Actors in a different application.

This **isn't** a problem with other Dapr functionality, because it's part of the users processing pipeline explicitly (see above, this sample was already using pub/sub).

## Steps to Reproduce the Problem

Create a server side blazor project (`dotnet new blazorserver`)

Add Actors

Your actors will fail to enregister with the runtime, because the application is returning HTML from `/dapr/config`

vinayada1:
@rynowak - Could you please add a specific example of how the new way of registering actors would look like?

rynowak:
[...]

PR #533, opened 2021-01-08
Use Startup to register Actors

rynowak:
# Description

This change introduces new APIs for registering Actors DI and with
the HTTP pipeline based on the ASP.NET Core Startup.cs model. This
replaces `UseActors()` in Program.cs and delivers better integration
between the ASP.NET Core framework and Actors.

The bug that motivated this change was that the actors HTTP handlers
were separate from the application's main routing table. This meant that
certain routing patterns were totally broken (such as Blazor Server)
when used in the same project as actors. There was no real way to fix
this without a change in how we register the framework to make it more
consistent with other features that build on ASP.NET Core.

## Issue reference

Fixes: #434

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


docs/get-started-dapr-actor.md
```diff
@@ -437,19 +427,19 @@ In order to validate and debug actor service and client, we need to run actor se
 1. Run Dapr Runtime via Dapr cli
 
    ```bash
-   $ dapr run --app-id myapp --app-port 3000 --dapr-http-port 3500 dotnet run
+   $ dapr run --app-id myapp --app-port 5000 --dapr-http-port 3500 dotnet run
```
rynowak, line 430: For some reason this was using a non-default port 😕 - its an extra step to explain how to get asp.net to listen on another port, and it seems kinda unnecessary to put in **our** tutorial.

samples/Actor/DemoActor/Program.cs
```diff
@@ -5,42 +5,21 @@
...
-        /// <summary>
-        /// Creates a IWebHostBuilder.
-        /// </summary>
-        /// <param name="args">Arguments.</param>
-        /// <returns>IWebHostBuilder instance.</returns>
-        public static IWebHostBuilder CreateWebHostBuilder(string[] args) =>
-            WebHost.CreateDefaultBuilder(args)
-                .UseStartup<Startup>()
-                .ConfigureServices(services => 
+        public static IHostBuilder CreateHostBuilder(string[] args) =>
```
rynowak, line 18: note: our examples were using some outdated code. Projects based on 3.1 or 5.0 will have `IHostBuilder` here not `IWebHostBuilder`.

src/Dapr.Actors.AspNetCore/ActorsEndpointRouteBuilderExtensions.cs
```diff
@@ -3,31 +3,60 @@
 // Licensed under the MIT License.
 // ------------------------------------------------------------
 
-namespace Dapr.Actors.AspNetCore
+namespace Microsoft.AspNetCore.Builder
```
rynowak, line 6: It's the common convention to stick these things in the same namespace as built in stuff so it's easy to find. (yes this is OK in this case, but not in general).

src/Dapr.Actors.AspNetCore/ActorsEndpointRouteBuilderExtensions.cs
```diff
@@ -3,31 +3,60 @@
...
     {
-        public static IEndpointConventionBuilder AddDaprConfigRoute(this IEndpointRouteBuilder endpoints)
+        /// <summary>
+        /// Maps endpoints for Dapr Actors into the <see cref="IEndpointRouteBuilder" />.
+        /// </summary>
+        /// <param name="endpoints">The <see cref="IEndpointRouteBuilder" />.</param>
+        /// <returns>
+        /// An <see cref="IEndpointConventionBuilder" /> that can be used to configure the endpoints.
+        /// </returns>
+        public static IEndpointConventionBuilder MapActorsHandlers(this IEndpointRouteBuilder endpoints)
```
rynowak, line 27: As a bonus to this approach, all kinds of different configuration knobs are now available for the endpoints we create. You can configure CORS, Auth, or any other policy via these endpoints.

For instance it would be possible to do port filtering and only accept traffic from the dapr sidecar for actors while serving other content to anyone.

src/Dapr.Actors.AspNetCore/ActorsEndpointRouteBuilderExtensions.cs
```diff
@@ -80,10 +109,10 @@ public static IEndpointConventionBuilder AddActorMethodRoute(this IEndpointRoute
                         await context.Response.CompleteAsync();
                     }
                 }
-            });
+            }).WithDisplayName(b => "Dapr Actors Invoke");
```
rynowak, line 112: These descriptive names show in the logs produced by asp.net core.

vinayada1 (approved):
LGTM!

codeputer:
I create a proof of concept some time ago, and I'm looking forward to getting back to it and the use of DAPR.  I'm seeing a lot of changes.  I was hoping the RC would slow that down, but there still seems to be a lot of churn.... are we close to hitting the eye of the storm? When should I venture back in to DAPR? :)

rynowak:
The next release RC3 will be a really good one to pick up if you want something stable that's not going to change much. We're not expecting many impactful changes after this milestone.

tomkerkhove:
Any ETA for RC3 @rynowak ?

rynowak:
Soon™

I know that's not super helpful. We've got about 1.5 more weeks of working time and then whatever stabilization is necessary.

tomkerkhove:
Cool, thanks!

Issue #410, opened 2020-10-07
Actor Runtime should not report Exception text in HTTP responses


rynowak:
## Expected Behavior

Runtime failures should report sanitized or *known safe* content in error responses. ASP.NET Core has separate knobs for configuring the reporting of exception details in responses that should be used. 

## Actual Behavior

The actor runtime unconditionally reports the full exception details including stack trace in certain types of failures. This is a security bug due to the risk of disclosure. 

See: https://github.com/dapr/dotnet-sdk/blob/master/src/Dapr.Actors.AspNetCore/EndpointRouteBuilderExtensions.cs#L72

## Steps to Reproduce the Problem

Get the `DispatchWithoutRemotingAsync` code path to throw an exception. I found this through code review so I don't know the exact repo.

It should be possible to just delete the exception handling code here and this will do the right thing. It's likely this code is not tested because `context.Response.Headers.Add("Connection: close", default(string));` is not correct and will probably **also** throw an exception.

PR #535, opened 2021-01-08
Do not write exception to response body

rynowak:
# Description

This change removes code that would write the exception message and call
stack to the response body when a non-remoting invoke is used. While
this kind of thing can be helpful for debugging, the best thing to do is
actually just let the exception bubble. This allows users to use the
ASP.NET Core error handling features that are much richer and have
security related guidance built-in.

## Issue reference

Fixes: #410

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


src/Dapr.Actors.AspNetCore/ActorsEndpointRouteBuilderExtensions.cs
```diff
@@ -95,19 +95,7 @@ private static IEndpointConventionBuilder MapActorMethodEndpoint(this IEndpointR
...
                 else
                 {
-                    // write exception info in response.
-                    try
-                    {
-                        await runtime.DispatchWithoutRemotingAsync(actorTypeName, actorId, methodName, context.Request.Body, context.Response.Body);
-                    }
-                    catch (Exception e)
-                    {
-                        context.Response.Headers.Add("Connection: close", default(string));
```
rynowak, line 105: This case was very not likely having the desired effect anyway. This isn't a valid header 😆

vinayada1 (approved):

Issue #524, opened 2021-01-02
ModelBinding throws ArgumentNullException when state not found in store


amolenk:
## Expected Behavior

In the sample below, I'd expect `forecast` to be `null` when the value isn't found in the underlying data store.

```c#
[HttpGet("{city}")]
public ActionResult<WeatherForecast> Get([FromState("statestore", "city")] WeatherForecast forecast)
{
    if (forecast == null)
    {
        return NotFound();
    }

    return forecast;
}
```

## Actual Behavior

The code throws an `ArgumentNullException`. This happens at line 76 in dotnet-sdk/src/Dapr.AspNetCore/StateEntryModelBinder.cs (master):

```c#
bindingContext.ValidationState.Add(bindingContext.Result.Model, new ValidationStateEntry()
{
    // Don't do validation since the data came from a trusted source.
    SuppressValidation = true,
});
```

Here, `bindingContext.Result.Model` is `null`, which causes the exception. I guess we don't need to add the value to the validation state dictionary at all when it is `null`.

BTW if I use `StateEntry<WeatherForecast>` instead of `WeatherForecast`, everything works of course, because `StateEntry` itself is never `null`.

## Release Note
RELEASE NOTE:

rynowak:
Thanks for the report - if there's an exception then this is definitely a bug. 

I think there's an interesting discussion to be had about the right semantics for when this happens. MVC's set of behaviors for model binding are pretty complex and idiosyncratic. 

Some options:

- Validation error + WeatherForecast is non-null
- Validation error + WeatherForecase is null
- no validation error + WeatherForecast is non-null

I think the best analog for consistency would be the behavior of `[FromBody]` because both of these constructs delegate their work to the JSON serializer. `[FromBody]` results in a null model object and a validation errors (bodies are never optional). We don't have a requirement or guarantee with `[FromBody]` binding to be able to construct the value like we do with standard modelbinding.

However, this means for users using `[ApiController]` that you'd see an HTTP 400 returned automatically with an error is the state key is not found. That doesn't seem like a helpful behavior since missing state is more like a 404 than a business validation failure.

Interested in hearing more thoughts (with examples).

vinayada1:
I tried this with: https://github.com/dapr/dotnet-sdk/blob/master/samples/AspNetCore/ControllerSample/Readme.md#code-samples

[HttpGet("{account}")]
public ActionResult<Account> Get(StateEntry<Account> account)
{
    if (account.Value is null)
    {
[...]

PR #537, opened 2021-01-09
Fix the ArgumentNull exception

vinayada1:
# Description

When state entry is missing fix argument null exception

#524 

Please reference the issue this PR will close: #524 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


src/Dapr.AspNetCore/StateEntryModelBinder.cs
```diff
@@ -71,13 +71,20 @@ public async Task BindModelAsync(ModelBindingContext bindingContext)
...
+            }
+            else
+            {
+                bindingContext.Result = ModelBindingResult.Success(obj);
+                bindingContext.ValidationState.Add(bindingContext.Result.Model, new ValidationStateEntry()
+                {
+                    // Don't do validation since the data came from a trusted source.
+                    SuppressValidation = true,
+                });
+            }
```
rynowak, line 87: We should add integration tests for this using https://github.com/dapr/dotnet-sdk/blob/master/test/Dapr.AspNetCore.IntegrationTest.App/DaprController.cs

I'd expect the following to happen when the state isn't found in the state store:
- If the parameter type is `StateEntry<Widget>` the `StateEntry<>` is non-null, but `StateEntry<>.Value` is null
- If the parameter type is `Widget` then the argument should be null.

For both of these cases, check `ModelState.IsValid` (should return true)

rynowak: I expect that this is already the behavior but the tests will help a lot.

rynowak: It would also really help to add some comments here about the expected behavior from a user POV (basically what I said above). The comments are useful because the relationship between this code and the outcome a user sees isn't obvious just from reading the code.

rynowak (approved):

PR #538, opened 2021-01-09
Add F# BankActor sample

cartermp:
~~Pair~~ Teams programming with @rynowak , for his eyes only right now


cartermp:
Since I can't get docker to install on Windows (lol classic, right?) I relied on @rynowak to run this and prove out that it actually worked. We may want his royal spiciness to run it again to make sure my `AggregateException` stuff actually works, and/or tweak it from there.

giuliohome:
Where are the F# samples currently?

PR #540, opened 2021-01-11
Fixes for the build

rynowak:
# Description

I discovered when running some tests locally, that we had merged a failing test, but it wasn't filing in the actual CI runs. The root cause was that we were maintaining multiple solution files for different purposes, and the tests didn't get added to the test solution. This isn't really necessary and is a complex way to organize things.

This change:

- Fixes the broken test
- Reduces the number of solution files so we don't break again
- Removes obsolete `azure-pipelines.yml`

## Issue reference

Please reference the issue this PR will close: no issue, just realized we weren't running all of the tests 😢 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


properties/dapr_managed_netcore.props
```diff
@@ -5,6 +5,7 @@
     <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <WarningLevel>4</WarningLevel>
+    <WarnOnPackingNonPackableProject>false</WarnOnPackingNonPackableProject>
```
rynowak, line 8: This is needed when you run `dotnet pack` on a solution - otherwise you get a bunch of warnings about test projects and samples which aren't packable.

test/Dapr.Actors.AspNetCore.IntegrationTest/HostingTests.cs
```diff
@@ -80,7 +80,7 @@ public async Task MapActorsHandlers_WorksWithFallbackRoute()
             await Assert2XXStatusAsync(response);
         }
 
-        private static IHost CreateHost<TStartup>()
+        private static IHost CreateHost<TStartup>() where TStartup : class
```
rynowak, line 83: Thanks @lukepammant for the clue about this being wrong 😆

vinayada1 (approved):

Issue #526, opened 2021-01-02
Reimagining service invocation for dotnet-sdk


rynowak:
## Describe the proposal

One of the things that's evolved over the last year of Dapr's existence has been how the SDKs interface with service invocation. In particular the .NET, Java and Go SDKs use gRPC as an implementation detail to mediate between a client API surface (.NET in our case) and what's most likely an HTTP service being invoked on the other side. Invoke is interesting to talk about because it's one of the few places in Dapr where your code is probably on both sides of the interaction. 

The fundamental problem is that the .NET APIs we provide in this project aren't really *the truth* - the underlying system is defined by both the [Dapr API for service invocation](https://docs.dapr.io/reference/api/service_invocation_api/) and the callee (another web app). It's most likely that developer writing applications with Dapr wrote both sides of the operation when using service invocation (both the caller and callee).  The underlying Dapr APIs intend to be a passthrough that can preserve the semantics of the HTTP communication protocol in use - however the problem is that the SDK and the choice to use gRPC as a detail result in an *impedance mismatch* - we're not currently doing a good job being a passthough.

I think we could leverage more of users' existing experience and the .NET community's existing assets by making the standard `HttpClient` API surface featured more prominently in Dapr's samples and guidance. I want to start a discussion about this and see what kinds of changes make sense. This proposal includes some ideas about how we could make using Dapr with HTTP simpler within the scope of the .NET SDK without changes to Dapr's protocol or runtime.

## The current landscape

The current design approach is that the `DaprClient.InvokeMethodAsync` family of APIs map to use of the [Dapr gRPC APIs for service invocation](https://github.com/dapr/dapr/blob/master/dapr/proto/runtime/v1/dapr.proto#L22). The Dapr API for service invocation provides an abstraction over HTTP and gRPC - in theory a caller does not need to know what protocol is used in the callee. 

Remember that *the big idea* here is that we can provide a set of .NET APIs that *map to either HTTP or gRPC* on the remote end *without the caller needing to know*. I'll briefly enumerate some of the issues that we've found with this approach - I think for most serious users, you will have encountered one of these limitations.

### HTTP and gRPC are different

The headline problem is that the *style* of designing services in HTTP and gRPC are different, especially when you look at non-success cases. If it helps, think of gRPC as an opinionated subset of HTTP. 

[...]

PR #544, opened 2021-01-14
Add HttpClient support for Dapr service invocation

rynowak:
# Description

- Adds a new message handler for HttpClient interop
- Adds an easy way to create HttpClient instances
- A bunch of sample cleanup/polish/simplification

The main thing here is the added ability to interop with HttpClient. We'll be making some future changes to the service invocation APIs in DaprClient to address https://github.com/dapr/dapr/issues/2342 since that issues greatly simplifies the interface we're dealing with. 

## Issue reference

Fixes: #526

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/samples/Client/DaprClient/DaprClient.csproj
+++ b/samples/Client/DaprClient/DaprClient.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net5</TargetFramework>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
     <RootNamespace>Samples.Client</RootNamespace>
     <Nullable>enable</Nullable>
   </PropertyGroup>
@@ -17,6 +17,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="System.Net.Http.Json" Version="5.0.0" />
     <PackageReference Include="Google.Protobuf" Version="3.13.0" />
     <PackageReference Include="Google.Api.CommonProtos" Version="2.2.0" />
   </ItemGroup>
--- a/samples/Client/DaprClient/DaprClient.csproj
+++ b/samples/Client/DaprClient/DaprClient.csproj
@@ -2,7 +2,9 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <TargetFramework>net5</TargetFramework>
+    <RootNamespace>Samples.Client</RootNamespace>
+    <Nullable>enable</Nullable>
   </PropertyGroup>
 
   <ItemGroup>
--- a/samples/Client/DaprClient/Program.cs
+++ b/samples/Client/DaprClient/Program.cs
@@ -15,6 +15,7 @@ namespace Samples.Client
         {
             new InvokeServiceGrpcExample(),
             new InvokeServiceHttpExample(),
+            new InvokeServiceHttpClientExample(),
             new PublishEventExample(),
             new StateStoreExample(),
             new StateStoreTransactionsExample(),
--- a/samples/Client/DaprClient/Program.cs
+++ b/samples/Client/DaprClient/Program.cs
@@ -3,353 +3,42 @@
 // Licensed under the MIT License.
 // ------------------------------------------------------------
 
-namespace DaprClient
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Samples.Client
 {
-    using System;
-    using System.Collections.Generic;
-    using System.Text.Json;
-    using System.Threading.Tasks;
-    using Dapr;
-    using Dapr.Client;
-    using Grpc.Core;
-    using Microsoft.Extensions.Configuration;
-
-    /// <summary>
-    /// Shows Dapr client calls.
-    /// </summary>
-    public static class Program
+    class Program
     {
-        private static readonly string stateKeyName = "mykey";
-        private static readonly string storeName = "statestore";
-        private static readonly string pubsubName = "pubsub";
-        private static readonly string daprErrorInfoHTTPCodeMetadata  = "http.code";
-        private static readonly string daprErrorInfoHTTPErrorMetadata = "http.error_message";
-        private static readonly string grpcStatusDetails = "grpc-status-details-bin";
-        private static readonly string grpcErrorInfoDetail = "google.rpc.ErrorInfo";
-
-        /// <summary>
-        /// Main entry point.
-        /// </summary>
-        /// <param name="args">
-        /// set parameter "useRouting" true to use routing service;
-        /// set parameter "useGrpcsample" true to use grpcsample service
-        /// </param>
-        /// <returns>A <see cref="Task"/> representing the asynchronous operation.</returns>
-        public static async Task Main(string[] args)
+        private static readonly Example[] Examples = new Example[]
         {
-            var builder = new ConfigurationBuilder();
-            builder.AddCommandLine(args);
+            new InvokeServiceGrpcExample(),
+            new InvokeServiceHttpExample(),
+            new PublishEventExample(),
+            new StateStoreExample(),
+            new StateStoreTransactionsExample(),
+            new StateStoreETagsExample(),
+        };
 
-            var config = builder.Build();
-
-            var jsonOptions = new JsonSerializerOptions()
+        static async Task<int> Main(string[] args)
+        {
+            if (args.Length > 0 && int.TryParse(args[0], out var index) && index >= 0 && index < Examples.Length)
             {
-                PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
-                PropertyNameCaseInsensitive = true,
-            };
+                var cts = new CancellationTokenSource();
+                Console.CancelKeyPress += (object? sender, ConsoleCancelEventArgs e) => cts.Cancel();
 
-            var client = new DaprClientBuilder()
-                .UseJsonSerializationOptions(jsonOptions)
-                .Build();
-
-            await PublishEventAsync(client);
-
-            // Save State
-            await SaveStateAsync(client);
-
-            // Read State
-            await GetStateAsync(client);
-
-            // Delete State
-            await DeleteStateAsync(client);
-
-            // State Transaction
-            await ExecuteStateTransactionAsync(client);
-
-            // Read State
-            await GetStateAfterTransactionAsync(client);
-
-            // Read and save state with etags
-            await ReadAndSaveStateWithEtagsAsync(client);
-
-            if (config.GetValue("rpc-exception", false))
-            {
-                // Invoke /throwException route on the Controller sample server
-                await InvokeThrowExceptionOperationAsync(client);
+                await Examples[index].RunAsync(cts.Token);
+                return 0;
             }
 
-            // Invoke deposit operation on ControllerSample or RoutingSample or GrpcServiceSample service by publishing event.
-            await PublishDepositeEventAsync(client);
-
-            await Task.Delay(TimeSpan.FromSeconds(1));
-
-            #region Service Invoke - Required RoutingService
-            // This provides an example of how to invoke a method on another REST service that is listening on http.
-            // To use it run RoutingService in this solution.
-
-            if (config.GetValue("useRouting", false))
+            Console.WriteLine("Hello, please choose a sample to run:");
+            for (var i = 0; i < Examples.Length; i++)
             {
-                //Invoke deposit operation on RoutingSample service by POST.
-                await InvokeDepositServiceOperationAsync(client);
-
-                //Invoke withdraw operation on RoutingSample service by POST.
-                await InvokeWithdrawServiceOperationAsync(client);
-
-                //Invoke balance operation on RoutingSample service by GET.
-                await InvokeBalanceServiceOperationAsync(client);
+                Console.WriteLine($"{i}: {Examples[i].DisplayName}");
             }
-            #endregion
-
-            #region Service Invoke via GRPC - Required GrpcServiceSample
-
-            if (config.GetValue("useGrpcsample", false))
-            {
-                //Invoke deposit operation on GrpcServiceSample service by GRPC.
-                await InvokeGrpcDepositServiceOperationAsync(client);
-
-                //Invoke withdraw operation on GrpcServiceSample service by GRPC.
-                await InvokeGrpcWithdrawServiceOperationAsync(client);
-
-                //Invoke balance operation on GrpcServiceSample service by GRPC.
-                await InvokeGrpcBalanceServiceOperationAsync(client);
-            }
-
-            #endregion
-
-            Console.WriteLine("Done");
-        }
-
-        internal static async Task PublishDepositeEventAsync(DaprClient client)
-        {
-            var eventData = new { Id = "17", Amount = (decimal)10, };
-            await client.PublishEventAsync(pubsubName, "deposit", eventData);
-            Console.WriteLine("Published deposit event!");
-        }
-
-        internal static async Task PublishEventAsync(DaprClient client)
-        {
-            var eventData = new Widget() { Size = "small", Color = "yellow", };
-            await client.PublishEventAsync(pubsubName, "TopicA", eventData);
-            Console.WriteLine("Published Event!");
-        }
-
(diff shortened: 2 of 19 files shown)


samples/Client/DaprClient/Readme.md
```diff
@@ -2,177 +2,54 @@
...
-
-The following example invokes another service:
-
-```c#
-
-var res = await client.InvokeMethodAsync<object>(serviceName, methodName);
-```
-
-In this example, the HttpGet method is invoked on the service with name *serviceName*. This argument specifies the name of the service, which was used when in conjunction with *dapr run –app-id serviceName*.The *route* argument specifies the route of the service endpoint.
-For example, if your service listens on HTTP, the route is typically defined by attribute *HttpPost(“route”)* or *HttpGet(“route”)*. Last argument is called metadata and it specifies GET operation in its example.
```
rynowak, line 85: It felt to me like the time was right to really dramatically simplify the content here and let the samples + API docs speak for themselves. A lot of this information reads like a tutorial - and makes sense when the richer content doesn't yet exist. However there have been several books written about this now.

samples/Client/DaprClient/DaprClient.csproj
```diff
@@ -2,7 +2,9 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <TargetFramework>net5</TargetFramework>
```
vinayada1, line 5: I thought we are going to target net5 only for the tests. The other samples are also targeting 3.1

samples/Client/DaprClient/Readme.md
```diff
@@ -2,177 +2,54 @@
...
-```
-
-or
+ Press Ctrl+C to exit, and then run the command again and provide a sample number to run the samples.
 
-``` sh
- dapr run --app-id DaprClient -- dotnet run --useGrpcsample true
-```
+  ```sh
+ dapr run --app-id DaprClient -- dotnet run 0
```
vinayada1, line 25: Remove this or maybe add a line: "for example, run this command to invoke the grpc sample"

samples/Client/DaprClient/Program.cs
```diff
@@ -3,353 +3,42 @@
...
-
-            if (config.GetValue("useGrpcsample", false))
-            {
-                //Invoke deposit operation on GrpcServiceSample service by GRPC.
-                await InvokeGrpcDepositServiceOperationAsync(client);
-
-                //Invoke withdraw operation on GrpcServiceSample service by GRPC.
-                await InvokeGrpcWithdrawServiceOperationAsync(client);
+            new InvokeServiceGrpcExample(),
+            new InvokeServiceHttpExample(),
```
vinayada1, line 17: InvokeServiceHttpClientExample is missing

samples/Client/DaprClient/Program.cs
```diff
@@ -3,353 +3,42 @@
...
-                // Invokes a POST method named "throwException" that takes input of type "Transaction" as defined in the ControllerSample.            
-                await client.InvokeMethodAsync("controller", "throwException", data, HttpInvocationOptions.UsingPost());
-            }
-            catch (RpcException ex)
+            Console.WriteLine("Hello, please choose a sample to run:");
+            for (var i = 0; i < Examples.Length; i++)
             {
-                var entry = ex.Trailers.Get(grpcStatusDetails);
-                var status = Google.Rpc.Status.Parser.ParseFrom(entry.ValueBytes);
-                Console.WriteLine("Grpc Exception Message: " + status.Message);
```
vinayada1, line 287: The sample code to handle grpc exceptions got removed

samples/Client/DaprClient/Readme.md
```diff
@@ -2,177 +2,54 @@
 The client sample shows how to make Dapr calls to publish events, save state, get state and delete state using a Dapr client apis.
```
vinayada1, line 2: invoking services is missing

samples/Client/DaprClient/Readme.md
```diff
@@ -2,177 +2,54 @@
...
-
-``` sh
- dapr run --app-id DaprClient -- dotnet run --useRouting true --useGrpcsample true
-```
+ Samples that use HTTP-based service invocation will require running the [RoutingService](../../AspNetCore/RoutingSample).
+ 
+ Samples that use gRPC-based service invocation will require running [GrpcService](../../AspNetCore/GrpcServiceSample).
 
 ## Invoking Services
+
```
vinayada1, line 33: Moving this description to line 10 would make more sense

samples/Client/DaprClient/Readme.md
```diff
@@ -2,177 +2,54 @@
...
- ``` 
-
-The method *PublishDepositeEventToRoutingSampleAsync* demonstrates how to publish an event to the dapr runtime, which triggers the operation 'deposit'.
- ```c#
-            await client.PublishEventAsync("deposit", new  { id = "17", amount = (decimal)10, });          
- ```
-
-
-## Handling RpcException
+See: `InvokeServiceHttpClientExample.cs` for an example of using `HttpClient` to invoke another service through Dapr.
```
vinayada1, line 46: Remove this and the next line (or move it with the addition of grpcsample)

samples/Client/DaprClient/StateStoreETagsExample.cs
```diff
@@ -0,0 +1,56 @@
...
+            var (updated, updatedETag) = await client.GetStateAndETagAsync<Widget>(storeName, stateKeyName, cancellationToken: cancellationToken);
+            Console.WriteLine($"Retrieved state: {updated.Size}  {updated.Color} with etag: {updatedETag}");
+
+            // Modify the state and try saving it with the old etag. This will fail
+            updated.Color = "purple";
+            var isSaveStateSuccess = await client.TrySaveStateAsync<Widget>(storeName, stateKeyName, updated, originalETag, cancellationToken: cancellationToken);
+            Console.WriteLine($"Saved state with old etag: {originalETag} successfully? {isSaveStateSuccess}");
+
+            // Modify the state and try saving it with the updated etag. This will succeed
+            isSaveStateSuccess = await client.TrySaveStateAsync<Widget>(storeName, stateKeyName, updated, updatedETag, cancellationToken: cancellationToken);
```
vinayada1, line 46: This belongs to the etag sample?

src/Dapr.Client/DaprClient.cs
```diff
@@ -15,6 +17,50 @@ namespace Dapr.Client
...
+            var handler = new InvocationHandler(){ InnerHandler = new HttpClientHandler(), };
+
+            if (daprEndpoint is string)
+            {
+                handler.DaprEndpoint = daprEndpoint;
+            }
+
+            var httpClient = new HttpClient(handler);
+            
+            if (appId is string)
```
vinayada1, line 56: Should this be string.IsNullOrEmpty()

src/Dapr.Client/DaprClient.cs
```diff
@@ -15,6 +17,50 @@ namespace Dapr.Client
...
+        /// The <see cref="HttpClient" /> object is intended to be a long-lived and holds access to networking resources.
+        /// Since the value of <paramref name="daprEndpoint" /> will not change during the lifespan of the application,
+        /// a single client object can be reused for the life of the application.
+        /// </para>
+        /// </remarks>
+        public static HttpClient CreateInvokeClient(string appId = null, string daprEndpoint = null)
+        {
+            var handler = new InvocationHandler(){ InnerHandler = new HttpClientHandler(), };
+
+            if (daprEndpoint is string)
```
vinayada1, line 49: Should this be string.IsNullOrEmpty()

src/Dapr.Client/DaprClient.cs
```diff
@@ -15,6 +17,50 @@ namespace Dapr.Client
...
+        /// <param name="daprEndpoint">The HTTP endpoint of the Dapr process to use for service invocation calls.</param>
+        /// <returns>An <see cref="HttpClient" /> that can be used to perform service invocation requests.</returns>
+        /// <remarks>
+        /// <para>
+        /// The <see cref="HttpClient" /> object is intended to be a long-lived and holds access to networking resources.
+        /// Since the value of <paramref name="daprEndpoint" /> will not change during the lifespan of the application,
+        /// a single client object can be reused for the life of the application.
+        /// </para>
+        /// </remarks>
+        public static HttpClient CreateInvokeClient(string appId = null, string daprEndpoint = null)
```
vinayada1, line 45: Why are we not using the builder pattern here like for the grpc case?

vinayada1, line 45: CreateInvokeHttpClient?

src/Dapr.Client/InvocationHandler.cs
```diff
@@ -0,0 +1,137 @@
...
+        {
+            var original = request.RequestUri;
+            if (!this.TryRewriteUri(request.RequestUri, out var rewritten))
+            {
+                throw new ArgumentException($"The request URI '{original}' is not a valid Dapr service invocation destination.", nameof(request));
+            }
+
+            try
+            {
+                if (this.apiToken is string)
```
vinayada1, line 97: Should this be string.IsNullOrEmpty

src/Dapr.Client/InvocationHandler.cs
```diff
@@ -0,0 +1,137 @@
...
+                request.RequestUri = original;
+                request.Headers.Remove("dapr-api-token");
+            }
+        }
+
+        // Internal for testing
+        internal bool TryRewriteUri(Uri uri, [NotNullWhen(true)] out Uri? rewritten)
+        {
+            // For now the only invalid cases are when the request URI is missing or just silly.
+            // We may support additional cases for validation in the future (like an allow-list of App-Ids).
```
vinayada1, line 116: Why should the allowlist be in the sdk and not dapr?

samples/Client/DaprClient/DaprClient.csproj
rynowak: Good point, I'll revert this part.

test/Dapr.Client.Test/DaprClientTest.cs
```diff
@@ -0,0 +1,26 @@
...
+            var client = DaprClient.CreateInvokeClient(appId: "bank", daprEndpoint: "http://localhost:3500");
+            Assert.Equal("http://bank/", client.BaseAddress.AbsoluteUri);
+        }
+
+        [Fact]
+        public void CreateInvokeClient_WithoutAppId()
+        {
+            var client = DaprClient.CreateInvokeClient(daprEndpoint: "http://localhost:3500");
+            Assert.Null(client.BaseAddress);
+        }
```
vinayada1, line 24: UTs with/without dapr endpoint?

test/Dapr.Client.Test/InvocationHandlerTests.cs
```diff
@@ -0,0 +1,143 @@
...
+#nullable enable
+
+namespace Dapr.Client
+{
+    public class InvocationHandlerTests 
+    {
+        [Fact]
+        public void TryRewriteUri_FailsForInvalidUris()
+        {
+            var uri = new Uri("/test"); // We don't handle many invalid cases today, just non-absolute URIs
```
vinayada1, line 24: and non-http scheme

samples/Client/DaprClient/Program.cs
rynowak: This sample was added before we had support in the box for automatically unpacking the status details. This was to document the workaround. Now we have this support in the box, and it's actually not meaningful anymore because we're getting rid of the gRPC -> HTTP invoke path wholesale.

samples/Client/DaprClient/Readme.md
```diff
@@ -2,177 +2,54 @@
...
-Retrieved state: small  yellow with etag: 1
-Saved modified state : small  orange
-Retrieved state: small  orange with etag: 2
-Saved state with old etag: 1 successfully? False
-Saved state with latest etag: 2 successfully? True
-Published deposit event!
-Done
+Running the following command will output a list of the samples included. 
+ ```sh
+ dapr run --app-id DaprClient -- dotnet run
```
vinayada1, line 19: This organization of samples looks much better!

src/Dapr.Client/InvocationHandler.cs
```diff
@@ -0,0 +1,137 @@
...
+                return false;
+            }
+
+
+            var builder = new UriBuilder(uri)
+            {
+                Scheme = this.parsedEndpoint.Scheme,
+                Host = this.parsedEndpoint.Host,
+                Port = this.parsedEndpoint.Port,
+                Path = $"/v1.0/invoke/{uri.Host}/method" + uri.AbsolutePath,
```
vinayada1, line 130: Should "v1.0" be a const now that we are approaching v1?

samples/Client/DaprClient/Program.cs
rynowak: Good catch

samples/Client/DaprClient/Readme.md
rynowak: Cleaned this up a bit.

src/Dapr.Client/DaprClient.cs
rynowak: If someone passes empty string, it will just result in an invalid configuration. I'm not sure why anyone would think that empty string is a valid value.

rynowak: btw if you pass empty string, it will produce an invalid URL on like 58 and throw a good descriptive exception

src/Dapr.Client/DaprClient.cs
rynowak: I'm going to add validation here using Uri so we can throw a good exception.

src/Dapr.Client/DaprClient.cs
rynowak: Builder APIs are complex and undiscoverable. I'd prefer not to use a sledgehammer to squash a fly. 

This is a convenience API that does glue code for something the user already can do (create an HttpClient). If it becomes complex then there's no value.

rynowak: Sure, I think that would be clearer.

src/Dapr.Client/InvocationHandler.cs
rynowak: I'll handle this where we read the environment variable. In general it's not a good idea to equate null and empty, but for an env-var it makes sense.

src/Dapr.Client/InvocationHandler.cs
rynowak: Speculating about future feature ideas, but an allowlist at the SDK level would apply to the set of URLs we transform, not to access control.

It's a fairly common pattern with `HttpClient` where auth is involved to be able to specify what domains that auth applies to so you don't leak tokens to inappropriate domains.

A similar idea applies here... right now we assume this `HttpClient` is only used for Dapr service invocation. We might in the future add support for an allow list, so users could use the same HttpClient instance for non-Dapr cases.

Since this is effectively a TODO, there's no action to take here.

src/Dapr.Client/DaprClient.cs
rynowak: This validation is already done by DaprEndpoint on the handler.

test/Dapr.Client.Test/DaprClientTest.cs
rynowak: There's no reason to test all possible combinations of arguments when they don't interact. We should test argument validation when there's something meaningful happening, these just set properties.

test/Dapr.Client.Test/InvocationHandlerTests.cs
rynowak: added

src/Dapr.Client/InvocationHandler.cs
rynowak: I don't think there's any value in trying to optimize for that. I'm **certain** that whatever we do for versioning post-1.0 will be more complex than just updating a constant.

samples/Client/DaprClient/Readme.md
rynowak: Since you're not around to talk about your comments, I'm going to just do some general cleanup on this and we can talk more later.

vinayada1 (approved):

yaron2 (approved):
lgtm

Issue #545, opened 2021-01-14
Support null etag


artursouza:
## Expected Behavior

Comply with new proto for etag as per https://github.com/dapr/dapr/pull/2666/files

## Actual Behavior

eTag is still handled as a simple string.

## Steps to Reproduce the Problem

Use master version of daprd and use any API that uses etag.

## Release Note

RELEASE NOTE: **FIXED** Support for null and empty etag.

PR #553, opened 2021-01-20
handle empty etag

vinayada1:
# Description

Differentiate between null and empty etag

#545 

Please reference the issue this PR will close: #545 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


rynowak:
What are the actual expections for ETags? A null ETag means *unspecified*. What does it mean to do `TrySaveStateAsync` but not specify an ETag? 

Is an *empty* ETag an illegal value? Does the Dapr runtime reject it?

vinayada1:
> Is an _empty_ ETag an illegal value? Does the Dapr runtime reject it?

Yes Dapr runtime treats null as a valid value (i.e. no etag specified) but empty would treat the tag as invalid and throw:-
switch e.Kind() {
	case state.ETagMismatch:
		return status.Errorf(codes.Aborted, format, args...)
	case state.ETagInvalid:
		return status.Errorf(codes.InvalidArgument, format, args...)

I thought it might be better to check for empty etag in the sdk code instead of checking the exception from runtime

rynowak:
> > Is an _empty_ ETag an illegal value? Does the Dapr runtime reject it?
> 

BTW that's not the code that validates ETags, that's the code that responds to a state store saying the ETag is invalid 😆 

I think this question still needs answering: 

> What does it mean to do TrySaveStateAsync but not specify an ETag?

The `Try{Save,Delete}StateAsync` methods are specifically meant for use with OCC. There's other methods `SaveStateAsync` that don't perform OCC and don't accept an ETag.

samples/Client/DaprClient/StateStoreETagsExample.cs
```diff
@@ -21,30 +21,63 @@ public override async Task RunAsync(CancellationToken cancellationToken)
...
-            await client.SaveStateAsync(storeName, stateKeyName, new Widget() { Size = "small", Color = "yellow", }, cancellationToken: cancellationToken);
+            // Save state with a null etag. This will throw an exception
+            try
+            {
+                await client.TrySaveStateAsync<Widget>(storeName, stateKeyName,  new Widget() { Size = "small", Color = "yellow", }, null, cancellationToken: cancellationToken);
+            }
+            catch(ArgumentException)
+            {
+                Console.WriteLine($"TrySaveStateAsync with null etag threw ArgumentException");
+            }
```
rynowak, line 32: I don't think this belongs in the sample. How does it help users to see code that will never succeed?

samples/Client/DaprClient/StateStoreETagsExample.cs
```diff
@@ -21,30 +21,63 @@ public override async Task RunAsync(CancellationToken cancellationToken)
...
+            var isDeleteStateSuccess = await client.TryDeleteStateAsync(storeName, stateKeyName, originalETag, cancellationToken: cancellationToken);
+            Console.WriteLine($"Deleted state with old etag: {originalETag} successfully? {isDeleteStateSuccess}");
+
+            // Read the updated state and etag
+            (updated, updatedETag) = await client.GetStateAndETagAsync<Widget>(storeName, stateKeyName, cancellationToken: cancellationToken);
+            Console.WriteLine($"Retrieved state: {updated.Size}  {updated.Color} with etag: {updatedETag}");
+
+            // Delete with updated etag. This will succeed
+            isDeleteStateSuccess = await client.TryDeleteStateAsync(storeName, stateKeyName, updatedETag, cancellationToken: cancellationToken);
+            Console.WriteLine($"Deleted state with updated etag: {updatedETag} successfully? {isDeleteStateSuccess}");
```
rynowak, line 80: Can you explain what the changes are to this sample? There are changes here, but I'm not really sure how it makes this a better sample.

rynowak (changes requested):
The actual library changes and tests here look good. I'm not sure quite what's up with the changes to the sample, so I want to understand more 😆

samples/Client/DaprClient/StateStoreETagsExample.cs
vinayada1: I added the TryDelete API because realized that the earlier sample only used Save.

samples/Client/DaprClient/StateStoreETagsExample.cs
vinayada1: earlier, I had some handling for empty etags as well and the sample was trying to demonstrate that null and empty etags are handled differently. but now that I have taken out the empty etag stuff from here, I think this can be removed too.

rynowak (approved):

rynowak (approved):

Issue #532, opened 2021-01-07
Provide a single way to configure System.Text.Json serializer options for all Dapr Client and Actors features


rynowak:
### The problem

We've gotten numerous pieces of feedback that the serializer options need to be configurable, and that there are too many places to configure them.

- #442 
- #441
- #476
- #362

The main issue is that the defaults don't work for certain types like F#'s unions in S.T.J. There are known workarounds for this, but it requires configuring the settings. It's unlikely that this problem will go away because in general writing a custom converter is the best solution for lack of serializer support. I'm sure we'll continue to find more cases and we need to allow users to unblock themselves.

### Proposal

The proposal is to add configuration of serializer options via DI. We'll introduce a new options type for Dapr's app-wide settings, and use DI-enabled entry points to get at the configuration. This means we'll need to change guidance for some features like `ActorProxy` which lead the user to a static entry point.

At the non-ASP.NET Core layer features like `DaprClientBuilder` will still have their own APIs for configuring settings, so usage of dotnet-sdk features in isolation will be the same as it is today.

virtualcca:
Is there an update to this issue, now I try to use the dateonly type under Net7, and I will find that if you call through an actor, it will report an XML-related serialization failure (From DCS)

It stands to reason that under net7, STJ can already support dateonly

Now you can only temporarily use DateOnly to be converted to DateTime as an Actor's interface to external calls

PR #556, opened 2021-01-22
Add IActorProxyFactory to DI

rynowak:
# Description

This change adds IActorProxyFactory to DI when used with ASP.NET Core,
and makes it also available as a property on the Actor base class. This
way if you've configured the ActorRuntimeOptions with your JSON
settings, its easy to have them picked up in Actors client use cases as
well.

This doesn't quite go all the way to unifying the settings with
DaprClient - we don't want to make a change that disruptive this close
to release.

## Issue reference

Fixes: #532

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/src/Dapr.Actors.AspNetCore/ActorsServiceCollectionExtensions.cs
+++ b/src/Dapr.Actors.AspNetCore/ActorsServiceCollectionExtensions.cs
@@ -4,6 +4,7 @@
 // ------------------------------------------------------------
 
 using System;
+using Dapr.Actors.Client;
 using Dapr.Actors.Runtime;
 using Microsoft.Extensions.DependencyInjection.Extensions;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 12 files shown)


src/Dapr.Actors.AspNetCore/ActorsServiceCollectionExtensions.cs
```diff
@@ -38,7 +39,22 @@ public static void AddActors(this IServiceCollection services, Action<ActorRunti
...
+            });
+
+            services.TryAddSingleton<ActorProxyFactory>(s =>
+            {
+                var options = s.GetRequiredService<IOptions<ActorRuntimeOptions>>().Value;
+                var factory = new ActorProxyFactory() 
+                { 
+                    DefaultOptions =
+                    {
+                        JsonSerializerOptions = options.JsonSerializerOptions,
```
vinayada1, line 53: add ut

rynowak: Sure

vinayada1 (approved):

Issue #442, opened 2020-10-18
Provide alternatives for raw bytes in all Dapr Client APIs and remove automatic protobuf serialization


akkie:
Currently Dapr internally used serialization/deserialization in different components. It uses it in the DaprClient, in the Actor and I think also in the state implementation. For all the components the JSON serialization options can be set at different places. I propose a single serializer implementation that can be reused internally by Dapr and which easily allows the user to set the JSON serialization/deserialzation option once. The serializer implementation should also be exposed so that the internal serialization/deserilzation logic can be used in user code to enable gRPC APIs.

There exists currently two open issues:
https://github.com/dapr/dotnet-sdk/issues/362
https://github.com/dapr/dotnet-sdk/issues/441

I think this could be implemented as proposed in https://github.com/dapr/dotnet-sdk/pull/440 but with a more general approach. As example by implementing a `DaprSerializer` which can handle all the possible serialization/deserialization tasks.

rynowak:
Thanks @akkie for opening this and #441 per my request for discussion. I'm responding here, because this issue seems to also include addressing #441 inside its scope. I agree that this area can be improved and there are a lot of variables to discuss because there are many possible solutions.

### Background

Some background my POV as a long time contributor to the Dapr dotnet-sdk....

Early on we took a dependency on System.Text.Json in the dotnet-sdk and added a bunch of methods that use JSON serialization to interact with the Dapr sidecar (over HTTP). This seemed like the obvious thing to do because it was both convenient and useful. 

Since then we've seen features related to gRPC come online, the transport used in dotnet-sdk has swapped from HTTP->gRPC, and we've seen that *other* SDKS for Dapr expose the ability to work with raw bytes.

The truth is that it was bytes all along, but the dotnet-sdk has really biased its API surface towards JSON and specifically `System.Text.Json`. 

At some point the code to handle `IMessage`-implementing types specially using Protobuf got added, because that seemed like the simplest thing to do for gRPC cases. At this point it would be fair to say that the dotnet-sdk now has *"serialization logic"* - whereas before this point we just called a static method.

The point I want to make is that it was never a goal to have *serialization logic*. Dapr is for interoperability, and the thing that matters most in that regard is clarity and predictability.  Piggybacking on top of a well-known and documented implementation without layering our own logic is the best way to have clear behavior.

### Concerns

My main concern is that I think the idea of a *Dapr serializer* as an extensibility point that comes with the wrong assumptions.

[...]

PR #557, opened 2021-01-23
Remove automatic protobuf conversion

rynowak:
# Description

We're making it absolutely clear which methods do JSON (most of them)
and which do gRPC. We have the ability to do both 'raw' functionality
and optimized paths for JSON+HTTP and gRPC+Protobuf for invoke. We're
not going to add 'raw' or Protobuf support for other building blocks
until there's a specific use case.

For the most part the functionality that's being removed here isn't used
anymore since that was part of the invoke changes.

## Issue reference

Fixes: #442

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


src/Dapr.Client/TypeConverters.cs
```diff
@@ -21,65 +21,36 @@ internal static class TypeConverters
         /// <param name="options">The JSON serialization options.</param>
         /// <typeparam name="T">The type of the given data.</typeparam>
         /// <returns>The given data as JSON based byte string.</returns>
-        public static ByteString ToJsonByteString<T>(T data, JsonSerializerOptions options = null)
+        public static ByteString ToJsonByteString<T>(T data, JsonSerializerOptions options)
         {
-            if (data == null)
-            {
-                return ByteString.Empty;
-            }
```
rynowak, line 29: This change because `null` is absolutely a meaningful value in JSON. Different parsers handle "empty data" differently and some of them will barf on "empty" which makes it a bad candidate for silent conversion.

src/Dapr.Client/TypeConverters.cs
```diff
@@ -21,65 +21,36 @@ internal static class TypeConverters
...
-
-            var t = typeof(T);
+            return new Any() 
+            { 
+                Value = ToJsonByteString<T>(data, options),
 
-            return typeof(IMessage).IsAssignableFrom(t)
-                ? Any.Pack((IMessage) data)
-                : new Any {Value = ToJsonByteString(data, options), TypeUrl = t.FullName};
```
rynowak, line 57: `TypeUrl = t.FullName` was doing something wierd and invalid. The Dapr runtime gives special meaning to "empty" `TypeUrl` even though that's technically invalid also. There's no real reason for us to put the .NET type name here, it has no meaning.

src/Dapr.Client/TypeConverters.cs
```diff
@@ -21,65 +21,36 @@ internal static class TypeConverters
...
-        /// <param name="options">The JSON serialization options.</param>
-        /// <typeparam name="T">The type to convert to.</typeparam>
-        /// <returns>An instance of T.</returns>
-        public static T FromAny<T>(Any any, JsonSerializerOptions options = null)
+        public static T FromJsonByteString<T>(ByteString bytes, JsonSerializerOptions options)
         {
-            var t = typeof(T);
-
-            if (typeof(IMessage).IsAssignableFrom(t) && t.GetConstructor(System.Type.EmptyTypes) != null)
+            if (bytes.Length == 0)
```
rynowak, line 43: being permissive in what we accept.

src/Dapr.Client/TypeConverters.cs
```diff
@@ -21,65 +21,36 @@ internal static class TypeConverters
...
-                var generic = method.MakeGenericMethod(t);
-                return (T) generic.Invoke(any, null);
+                return default;
             }
+            
+            return JsonSerializer.Deserialize<T>(bytes.Span, options);
+        }
 
-            return JsonSerializer.Deserialize<T>(any.Value.ToStringUtf8(), options);
+        public static T FromJsonAny<T>(Any any, JsonSerializerOptions options)
```
rynowak, line 51: I changed these to have `options` as a require parameter because it bug if we don't pass the options in library code. This makes tests a little more messy, but IMO worth it.

test/Dapr.Client.Test/DaprClientTest.InvokeMethodGrpcAsync.cs
```diff
@@ -203,7 +203,7 @@ public async Task InvokeMethodGrpcAsync_CanInvokeMethodWithNoReturnTypeAndData_T
             var data = new Response() { Name = "Look, I was invoked!" };
             var invokeResponse = new InvokeResponse
             {
-                Data = TypeConverters.ToAny(data)
+                Data = Any.Pack(data),
```
rynowak, line 206: Working with protobuf and gRPC types is already easy, no need for custom logic.

test/Dapr.Client.Test/TypeConvertersTest.cs
```diff
@@ -38,11 +20,13 @@ public void AnyConversion_JSON_Serialization_Deserialization()
...
-            var any = TypeConverters.ToAny(response);
-            var type = TypeConverters.FromAny<Response>(any);
+            var options = new JsonSerializerOptions(JsonSerializerDefaults.Web);
+
+            var any = TypeConverters.ToJsonAny(response, options);
+            var type = TypeConverters.FromJsonAny<Response>(any, options);
 
             type.Should().BeEquivalentTo(response);
-            any.TypeUrl.Should().Be("Dapr.Client.Test.TypeConvertersTest+Response");
+            any.TypeUrl.Should().Be(string.Empty);
```
rynowak, line 29: note, this property sets itself to empty string. I didn't make it do that, it does it on its own.

vinayada1 (approved):

Issue #551, opened 2021-01-20
Expose Dapr API token to users as a settable value


rynowak:
## Describe the proposal

Currently the .NET SDK reads the `DAPR_API_TOKEN` environment variable and uses that token in the HTTP and gRPC calls it generates. 

This isn't really the right design because the API token is totally hidden from users. Dapr doesn't provide any built-in opinions or conventions around setting this environment variable for applications.

We should surface this though an API and speak to it through docs and samples.

PR #558, opened 2021-01-23
API to set DAPR_API_TOKEN

vinayada1:
# Description

Add API to set DAPR_API_TOKEN

#551 

Please reference the issue this PR will close: #551 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


src/Dapr.Actors/Client/ActorProxy.cs
```diff
@@ -305,5 +314,15 @@ protected bool CheckIfItsWrappedRequest(IActorRequestMessageBody requestMessageB
...
         }
+
+        /// <summary>
+        /// Sets the Dapr Api Token which will be added to the header in all requests
+        /// </summary>
+        /// <param name="apiToken">Dapr Api Token value.</param>
+        public void SetDaprApiToken(string apiToken)
+        {
+            ArgumentVerifier.ThrowIfNull(apiToken, nameof(apiToken));
+            Environment.SetEnvironmentVariable("DAPR_API_TOKEN", apiToken);
```
rynowak, line 325: We shouldn't set environment variables for this process - this needs to get passed through to the right location in our code so that it's more flexible than an env-var. The user already has the ability to set an env-var without us providing an API :)

src/Dapr.Actors/Client/ActorProxyOptions.cs
```diff
@@ -26,5 +28,15 @@ public JsonSerializerOptions JsonSerializerOptions
...
                     throw new ArgumentNullException(nameof(JsonSerializerOptions), $"{nameof(ActorProxyOptions)}.{nameof(JsonSerializerOptions)} cannot be null");
         }
+
+        /// <summary>
+        /// The Dapr Api Token that is added to the header for all requests.
+        /// </summary>>
+        public string DaprApiToken
+        {
+            get => this.daprApiToken;
+            set => this.daprApiToken = value ??
```
rynowak, line 38: What makes null an invalid value?

src/Dapr.Client/DaprClient.cs
```diff
@@ -431,5 +431,12 @@ public async Task<StateEntry<TValue>> GetStateEntryAsync<TValue>(string storeNam
             string storeName,
             IReadOnlyDictionary<string, string> metadata = default,
             CancellationToken cancellationToken = default);
+
+        /// <summary>
+        /// Sets the Dapr Api Token which will be added to the header in all requests
+        /// </summary>
+        /// <param name="apiToken">Dapr Api Token value.</param>
+        public abstract void SetDaprApiToken(
```
rynowak, line 439: This should be part of the builder, not part of the client. It's something that's fixed for the lifetime of the app.

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -1038,6 +1034,16 @@ private static Autogenerated.StateOptions.Types.StateConcurrency GetStateConcurr
                 _ => throw new ArgumentException($"{concurrencyMode} Concurrency Mode is not supported.")
             };
         }
+
+        private void AddDaprApiTokenHeader(Metadata headers)
+        {
+            var daprApiToken = Environment.GetEnvironmentVariable(Constants.DaprApiTokenEnvironmentVariable);
```
rynowak, line 1040: Similar to a comment elsewhere, the source of truth should be a field on this class, not the env-var. The reason for providing an API for this is that it adds flexibility to do other things.

Example: Some people **really really** don't like security-related things in environment variables, they have an organization policy to ready them from files on an in memory file-system.

samples/AspNetCore/GrpcServiceSample/GrpcServiceSample.csproj
```diff
@@ -5,8 +5,16 @@
     <RunAnalyzersDuringBuild>true</RunAnalyzersDuringBuild>
   </PropertyGroup>
 
+  <ItemGroup>
```
rynowak, line 8: Can you rebase this PR or something? I'm seeing a bunch of commits/changes from things that have already been merged to master.

vinayada1: done

docs/api-tokens.md
```diff
@@ -0,0 +1,33 @@
...
+    var remotingProxy = factory.CreateActorProxy<IDemoActor>(actorId, "DemoActor");
+
+    // Making calls without Remoting, this shows method invocation using InvokeMethodAsync methods, the method name and its payload is provided as arguments to InvokeMethodAsync methods.
+    var nonRemotingProxy = factory.Create(actorId, "DemoActor");
+```
+
+The calls made using the proxy created with the above code will have the request headers with:-
+"dapr-api-token":"your_token"
+
+Alternatively, the user can set an environment variable "DAPR_API_TOKEN" with the token value. In case, the user has specified the token using code as well as the environment variable is set, the value set via code will be used.
```
rynowak, line 33: 👍 great!

src/Dapr.Actors/Client/ActorProxy.cs
```diff
@@ -45,6 +45,7 @@ protected internal ActorProxy()
 
         internal IActorMessageBodyFactory ActorMessageBodyFactory { get; set; }
         internal JsonSerializerOptions JsonSerializerOptions = JsonSerializerDefaults.Web;
+        internal string DaprApiToken;
```
rynowak, line 48: nit: `daprApiToken`

rynowak: If this needs to be internal, it would be great to make this a property instead so it's not settable outside the class.

src/Dapr.Actors/Client/ActorProxyFactory.cs
```diff
@@ -15,22 +16,27 @@ namespace Dapr.Actors.Client
...
 
         /// <summary>
         /// Initializes a new instance of the <see cref="ActorProxyFactory"/> class.
         /// </summary>
-        public ActorProxyFactory()
+        public ActorProxyFactory(ActorProxyOptions options = null)
         {
-            this.daprInteractor = new DaprHttpInteractor();
+            this.defaultOptions = options ?? new ActorProxyOptions();
+            this.daprInteractor = new DaprHttpInteractor(innerHandler:options?.HttpClientMessageHandler, apiToken:options?.DaprApiToken);
```
rynowak, line 39: So this introduces a bunch of problems, having a class that that's mutable, and shared, but allow has per-proxy state. The http interactor is shared (one per factory), but has it's settings modified per-proxy you create (line 51).

For instance if you called `Create` twice with two different values of the API token then it would crosstalk. 

My thoughts on how you should address this design-wise:
- Get rid of `options.HttpClientMessageHandler` - this addition breaks the design.
- If you need to make the message handler customizable for tests, make it possible pass in in constructor of `ActorProxyFactory` - otherwise fallback to a **static** instance on `ActorProxyFactory`.
- Create new instance of `DaprHttpInteractor` inside `Create` but **do not** store them

src/Dapr.Actors/Client/ActorProxyOptions.cs
```diff
@@ -17,6 +18,9 @@ public class ActorProxyOptions
 
         private JsonSerializerOptions jsonSerializerOptions = JsonSerializerDefaults.Web;
 
+                private string daprApiToken;
```
rynowak, line 21: nit: fix indent

src/Dapr.Actors/Client/ActorProxyOptions.cs
```diff
@@ -26,5 +30,24 @@ public JsonSerializerOptions JsonSerializerOptions
...
         }
+
+        /// <summary>
+        /// The Dapr Api Token that is added to the header for all requests.
+        /// </summary>>
+        public string DaprApiToken
+        {
+            get => this.daprApiToken;
+            set => this.daprApiToken = value;
+        }
```
rynowak, line 41: you can just use `{ get; set; }` and there's no need for an explicit backing field

src/Dapr.Actors/Client/ActorProxyOptions.cs
```diff
@@ -26,5 +30,24 @@ public JsonSerializerOptions JsonSerializerOptions
             set => this.jsonSerializerOptions = value ??
                     throw new ArgumentNullException(nameof(JsonSerializerOptions), $"{nameof(ActorProxyOptions)}.{nameof(JsonSerializerOptions)} cannot be null");
         }
+
+        /// <summary>
+        /// The Dapr Api Token that is added to the header for all requests.
+        /// </summary>>
```
rynowak, line 36: extra `>` character

src/Dapr.Actors/Client/ActorProxyOptions.cs
```diff
@@ -26,5 +30,24 @@ public JsonSerializerOptions JsonSerializerOptions
...
+        public string DaprApiToken
+        {
+            get => this.daprApiToken;
+            set => this.daprApiToken = value;
+        }
+
+        /// <summary>
+        /// The HTTP message handler. Used for testing purpose
+        /// </summary>>
```
rynowak, line 45: extra `>` character

src/Dapr.Actors/Client/ActorProxyOptions.cs
```diff
@@ -26,5 +30,24 @@ public JsonSerializerOptions JsonSerializerOptions
...
+        public HttpClientHandler HttpClientMessageHandler
+        {
```
rynowak, line 47: What's the purpose of adding this? This breaks a lot of the design work we did earlier.

src/Dapr.Actors/DaprHttpInteractor.cs
```diff
@@ -461,5 +459,26 @@ private HttpClient CreateHttpClient()
...
+        private void AddDaprApiTokenHeader(HttpRequestMessage request)
+        {
+            if (!string.IsNullOrWhiteSpace(this.daprApiToken))
+            {
+                request.Headers.Add("dapr-api-token", this.daprApiToken);
+                return;
+            }
+
+            // Check if the env variable DAPR_API_TOKEN is set and use it as the default value
+            var apiToken = Environment.GetEnvironmentVariable("DAPR_API_TOKEN");
```
rynowak, line 472: We should be able to isolate this code to the options class. Putting it here makes this class have more dependencies taht are hard to test.

src/Dapr.Actors/Client/ActorProxyOptions.cs
rynowak: I would expect this to have its value initialized by the env-var

src/Dapr.Actors/DaprHttpInteractor.cs
rynowak: Whereas, it's easy to rationalize a get/set property without unit tests 😆

src/Dapr.Actors/Runtime/ActorRuntime.cs
```diff
@@ -55,7 +55,7 @@ internal ActorRuntime(ActorRuntimeOptions options, ILoggerFactory loggerFactory,
         /// </summary>
         public IReadOnlyList<ActorRegistration> RegisteredActors => this.options.Actors;
 
-        internal static IDaprInteractor DaprInteractor => new DaprHttpInteractor();
+        internal static IDaprInteractor DaprInteractor => new DaprHttpInteractor(new System.Net.Http.HttpClientHandler());
```
rynowak, line 58: this instance should be non-static, and you can use `ActorRuntimeOptions` to initialize it. Remember, only the `HttpClientHandler` needs to be long-lived, these objects are cheap.

src/Dapr.Client/DaprClient.cs
```diff
@@ -49,9 +50,13 @@ public abstract class DaprClient
         /// a single client object can be reused for the life of the application.
         /// </para>
         /// </remarks>
-        public static HttpClient CreateInvokeHttpClient(string appId = null, string daprEndpoint = null)
+        public static HttpClient CreateInvokeHttpClient(string appId = null, string daprEndpoint = null, string daprApiToken = null)
```
rynowak, line 53: 👍

src/Dapr.Client/DaprClientBuilder.cs
```diff
@@ -111,6 +112,18 @@ public DaprClientBuilder UseGrpcChannelOptions(GrpcChannelOptions grpcChannelOpt
...
         }
 
+        /// <summary>
+        /// Adds the provided <paramref name="apiToken" /> on every request to the Dapr runtime.
+        /// </summary>
+        /// <param name="apiToken">The token to be added to the request headers/>.</param>
+        /// <returns>The <see cref="DaprClientBuilder" /> instance.</returns>
+        public DaprClientBuilder UseDaprApiToken(string apiToken)
+        {
+            ArgumentVerifier.ThrowIfNullOrEmpty(apiToken, nameof(apiToken));
```
rynowak, line 122: Why not allow null or empty here? There's no way to unset after setting?

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -906,6 +905,20 @@ private static Autogenerated.StateOptions.Types.StateConcurrency GetStateConcurr
...
+            {
+                headers.Add("dapr-api-token", this.daprApiToken);
+                return;
+            }
+
+            var apiToken = Environment.GetEnvironmentVariable("DAPR_API_TOKEN");
+            if(!string.IsNullOrWhiteSpace(apiToken))
+            {
+                headers.Add("dapr-api-token", apiToken);
+            }
```
rynowak, line 920: Can we put this logic elsewhere? This logic appears in a bunch of places. Ideally the builder would be responsible for this, and the client would just do as its told.

src/Dapr.Client/InvocationHandler.cs
```diff
@@ -62,7 +54,10 @@ private static string GetDefaultDaprEndpoint()
         public InvocationHandler()
         {
             this.parsedEndpoint = new Uri(GetDefaultDaprEndpoint(), UriKind.Absolute);
-            this.apiToken = GetDaprApiToken();
+            if(string.IsNullOrWhiteSpace(this.apiToken))
+            {
+                this.apiToken = Environment.GetEnvironmentVariable("DAPR_API_TOKEN");
```
rynowak, line 59: Again, this logic shouldn't appear everywhere. It's OK to just make the existing property public

test/Dapr.Actors.Test/ApiTokenTests.cs
```diff
@@ -0,0 +1,175 @@
...
+        [Fact]
+        public void CreateProxyWithRemoting_WithApiTokenEnvVarSet()
+        {
+            var actorId = new ActorId("abc");
+            var handler = new TestHttpClientHandler();
+            var options = new ActorProxyOptions
+            {
+                HttpClientMessageHandler = handler
+            };
+            Environment.SetEnvironmentVariable("DAPR_API_TOKEN", "test_token");
```
rynowak, line 52: This is the reason why I want to see all of the defaulting logic pulled out of all of the implementation code. if you have one-two places that read the environment variable, and all of the complex parts of the system just trust the data they are given then it's really easy to understand and test.

src/Dapr.Actors/Client/ActorProxyFactory.cs
```diff
@@ -42,6 +48,7 @@ public TActorInterface CreateActorProxy<TActorInterface>(ActorId actorId, string
         public ActorProxy Create(ActorId actorId, string actorType, ActorProxyOptions options = null)
         {
             var actorProxy = new ActorProxy();
+            this.daprInteractor.SetDaprApiToken(this.DefaultOptions.DaprApiToken);
```
rynowak, line 51: So the problem here is that you're updating a shared instance. I would expect this line of code to create a new `DaprHttpInteractor`. 

Think about what happens when you call Create twice with a different API token. Both proxies would share an `DaprHttpInteractor` instance, and so this line of could would update both of them to use the second API token (last one wins).

src/Dapr.Actors/DaprHttpInteractor.cs
```diff
@@ -441,25 +433,21 @@ HttpRequestMessage FinalRequestFunc()
...
-                httpClientInstance.Timeout = (TimeSpan)this.clientSettings.ClientTimeout;
+                request.Headers.Add("dapr-api-token", this.daprApiToken);
+                return;
             }
+        }
 
-            return httpClientInstance;
+        public void SetDaprApiToken(string apiToken)
+        {
+            this.daprApiToken = apiToken;
```
rynowak, line 450: We should end up not needing this.

src/Dapr.Client/DaprClientBuilder.cs
```diff
@@ -111,6 +113,21 @@ public DaprClientBuilder UseGrpcChannelOptions(GrpcChannelOptions grpcChannelOpt
...
+        /// Adds the provided <paramref name="apiToken" /> on every request to the Dapr runtime.
+        /// </summary>
+        /// <param name="apiToken">The token to be added to the request headers/>.</param>
+        /// <returns>The <see cref="DaprClientBuilder" /> instance.</returns>
+        public DaprClientBuilder UseDaprApiToken(string apiToken)
+        {
+            this.DaprApiToken = apiToken;
+            if(string.IsNullOrWhiteSpace(apiToken))
+            {
+                this.DaprApiToken = Environment.GetEnvironmentVariable(Constants.DaprApiTokenEnvironmentVariable);
```
rynowak, line 126: This isnt' great because it's impossible to 'null out' the value if you don't want what's there. We'd be better off without this logic.

rynowak: Simple and predictable works every time.

src/Dapr.Actors/Runtime/ActorRuntime.cs
```diff
@@ -139,6 +143,16 @@ internal Task FireTimerAsync(string actorTypeName, string actorId, string timerN
             }
         }
 
+        internal static IDaprInteractor GetDaprInteractor(string actorTypeName)
```
rynowak, line 146: What does this change accomplish that's better than the property we had here before? I don't see why we need a separate interactor per-actor-type.

src/Dapr.Actors/Runtime/ActorRuntime.cs
```diff
@@ -22,6 +22,8 @@ public sealed class ActorRuntime
     {
         // Map of ActorType --> ActorManager.
         private readonly Dictionary<string, ActorManager> actorManagers = new Dictionary<string, ActorManager>();
+
+        private static Dictionary<string, IDaprInteractor> daprInteractors = new Dictionary<string, IDaprInteractor>();
```
rynowak, line 26: I think we can find a solution that doesn't require a static look up. I'm not sure why this had to be added.

Some general feedback about this design:
- When you think static, think **global variable** - if you're making something mutable and static, that's a red flag. If you're making a mutable static collection, then that's a double red flag.
- Dictionary is not thread safe so this kind of pattern will lead to all kinds of wierd behaviors and you won't catch it in tests

test/Dapr.Actors.Test/ApiTokenTests.cs
```diff
@@ -0,0 +1,163 @@
...
+
+            handler.Requests.TryDequeue(out var entry).Should().BeTrue();
+            var headerValues = entry.Request.Headers.GetValues("dapr-api-token");
+            headerValues.Should().Contain("test_token");
+        }
+
+        [Fact]
+        public void CreateProxyWithNoRemoting_WithApiTokenEnvVarSet()
+        {
+            Environment.SetEnvironmentVariable(Constants.DaprApiTokenEnvironmentVariable, "test_token");
```
rynowak, line 95: I'll leave it up to you to decide, but I think doing things like this in tests makes the wrong tradeoff, and we're better off not having a test for it. Think about the implications of running tests in parallel and you'll see what I mean, it doesn't scale.

My general advice for testing the parts of the same that interface with the OS (like env-vars):

- Isolate code that touches OS features from the important logic. In a case this, it's simple things like reading the env-var in a property initializer or in the constructor. (already done)
- Don't write tests for constructors and property initializers because it's low value 
- When possible rely on e2e tests - I know we don't have them for this scenario
- Trust yourself that you did a good job with the above and it will all be ok

test/Dapr.Client.Test/InvocationHandlerTests.cs
```diff
@@ -148,6 +153,55 @@ public async Task SendAsync_RewritesUri_AndAddsApiToken()
...
+            Assert.Equal("super-duper-secure", capture.DaprApiToken);
+
+            Assert.Equal(uri, request.RequestUri?.OriginalString);
+            Assert.False(request.Headers.TryGetValues("dapr-api-token", out _));
+        }
+
+        [Fact]
+        public async Task SendAsync_RewritesUri_WithApiTokenAndEnvVarSet_EnvVarIsIgnored()
+        {
+            Environment.SetEnvironmentVariable(Constants.DaprApiTokenEnvironmentVariable, "super-duper-secure2");
```
rynowak, line 183: Again, my feedback here is that this is a test for code that *we don't have*. Nothing in the InvocationHandler reads this environment variable, and we wouldn't add it back because we have no reason to.

src/Dapr.Actors/Client/ActorProxy.cs
```diff
@@ -170,6 +172,7 @@ public Task InvokeMethodAsync(string method, CancellationToken cancellationToken
             this.ActorType = actorType;
             this.ActorMessageBodyFactory = client.GetRemotingMessageBodyFactory();
             this.JsonSerializerOptions = options?.JsonSerializerOptions ?? this.JsonSerializerOptions;
+            this.daprApiToken = options?.DaprApiToken;
```
rynowak, line 175: I don't think this field is actually used anywehere

src/Dapr.Actors/Client/ActorProxyFactory.cs
```diff
@@ -42,6 +50,7 @@ public TActorInterface CreateActorProxy<TActorInterface>(ActorId actorId, string
         public ActorProxy Create(ActorId actorId, string actorType, ActorProxyOptions options = null)
         {
             var actorProxy = new ActorProxy();
+            this.daprInteractor = new DaprHttpInteractor(this.handler, this.DefaultOptions.DaprApiToken);
```
rynowak, line 53: There's no need for a field, this will simplify the design.

rynowak (approved):

Issue #516, opened 2020-12-18
Throw DaprException for dapr errors


artursouza:
Throw DaprException for runtime errors using Dapr via the SDK. For contract violations, use the standard exceptions expected by .Net developers.

PR #561, opened 2021-01-26
Throw DaprException for errors in DaprClient

rynowak:
# Description

This changes DaprClient to throw DaprException (or a subclass), wrapping the root cause, in locations where DaprClient does networking or handles user data. This allows for simpler error handling, more topical error messages, and in the case of invoke: more information is available.

There's also work here to unify the two classes we had called `DaprException`. The new `DaprException` class is pretty plain, and used for the base class of other exception types.

## Issue reference

#516 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


src/Dapr.Actors/Client/ActorProxy.cs
```diff
@@ -169,7 +169,7 @@ public Task InvokeMethodAsync(string method, CancellationToken cancellationToken
             this.ActorId = actorId;
             this.ActorType = actorType;
             this.ActorMessageBodyFactory = client.GetRemotingMessageBodyFactory();
-            this.JsonSerializerOptions = options?.JsonSerializerOptions ?? this.JsonSerializerOptions;
+            this.JsonSerializerOptions = options?.JsonSerializerOptions ?? new JsonSerializerOptions(JsonSerializerDefaults.Web);
```
rynowak, line 172: See comments on the commit about why this change was needed.

src/Dapr.Actors/DaprHttpInteractor.cs
```diff
@@ -373,25 +373,25 @@ HttpRequestMessage FinalRequestFunc()
                 }
                 catch (Exception ex)
                 {
-                    throw new DaprException(string.Format("ServerErrorNoMeaningFulResponse", response.StatusCode), ex);
+                    throw new DaprApiException(string.Format("ServerErrorNoMeaningFulResponse", response.StatusCode), ex);
```
rynowak, line 376: Actors kinda has its own style of errors based on the HTTP interface. Some of the code changed, the functionality should be the same.

src/Dapr.Client/DaprException.cs
```diff
@@ -31,5 +33,15 @@ public DaprException(string message, Exception innerException)
             : base(message, innerException)
         {
         }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DaprException"/> class with a specified context.
+        /// </summary>
+        /// <param name="info">The <see cref="SerializationInfo" /> object that contains serialized object data of the exception being thrown.</param>
+        /// <param name="context">The <see cref="StreamingContext" /> object that contains contextual information about the source or destination. The context parameter is reserved for future use and can be null.</param>
+        protected DaprException(SerializationInfo info, StreamingContext context)
```
rynowak, line 42: Serialization constructor is needed because the actors code has it.

test/Dapr.Client.Test/PublishEventApiTest.cs
```diff
@@ -109,11 +111,10 @@ public async Task PublishEventAsync_CanPublishTopicWithNoContent_WithMetadata()
             var task = daprClient.PublishEventAsync(TestPubsubName, "test", metadata);
             httpClient.Requests.TryDequeue(out var entry).Should().BeTrue();
             var request = await GrpcUtils.GetRequestFromRequestMessageAsync<PublishEventRequest>(entry.Request);
-            var jsonFromRequest = request.Data.ToStringUtf8();
 
             request.PubsubName.Should().Be(TestPubsubName);
             request.Topic.Should().Be("test");
-            jsonFromRequest.Should().Be("\"\"");
+            request.Data.Length.Should().Be(0);
```
rynowak, line 117: This was a bug identified on inspecting the publish code. Publishing "no content" would result in `""` whereas providing `null` would result in "no content".

test/Dapr.Client.Test/PublishEventApiTest.cs
```diff
@@ -87,7 +89,7 @@ public async Task PublishEventAsync_CanPublishTopicWithNoContent()
 
             request.PubsubName.Should().Be(TestPubsubName);
             request.Topic.Should().Be("test");
-            jsonFromRequest.Should().Be("\"\"");
```
vinayada1, line 90: Do we need a UT for WithContent_ContentEqualsNull?

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -234,24 +267,100 @@ public override Task<HttpResponseMessage> InvokeMethodWithResponseAsync(HttpRequ
...
+                return await this.httpClient.SendAsync(request, cancellationToken);
+            }
+            catch (HttpRequestException ex)
+            {
+                // Our code path for creating requests places these keys in the request properties. We don't want to fail
+                // if they are not present.
+                request.Properties.TryGetValue(AppIdKey, out var appId);
+                request.Properties.TryGetValue(MethodNameKey, out var methodName);
+
+                throw new InvocationException(
```
vinayada1, line 281: question - why do we have a special InvocationException class here whereas everything else uses DaprException?

src/Dapr.Client/DaprApiException.cs
```diff
@@ -12,79 +12,79 @@ namespace Dapr.Actors
...
         /// <summary>
-        /// Initializes a new instance of the <see cref="DaprException"/> class with a specified error
+        /// Initializes a new instance of the <see cref="DaprApiException"/> class with a specified error
         /// message and a reference to the inner exception that is the cause of this exception.
         /// </summary>
         /// <param name="message">The error message that explains the reason for the exception.</param>
         /// <param name="innerException">The exception that is the cause of the current exception.</param>
-        public DaprException(string message, Exception innerException)
-            : this(message, innerException, Constants.Unknown, false)
+        public DaprApiException(string message, Exception innerException)
```
vinayada1, line 40: Is DaprApiException like DaprActorException...the naming isn't very clear

src/Dapr.Client/DaprClientGrpc.cs
rynowak: This is a subclass of DaprException - so a catch for DapException would handle this as well. The reason for the subclass is that we have more information and context to include for invocation. Also needing to handle errors in this case is *much* more common since invoke calls business logic on the side - error cases are **expected**.

Basically, it makes sense to try and optimize this experience. The errors in the other cases are fairly generic, and could be improved in the future as needed based on feedback (without breaking changes).

test/Dapr.Client.Test/PublishEventApiTest.cs
rynowak: If you mean a for passing in `null` as a value, I don't mind having one. The thing is .. this isn't testing what our code does, its mostly testing what our code DOESN'T do 😆 I don't mind having one test, but I don't want to go on an extensive journey of testing the underlying behaviors of the JSON serializer.

src/Dapr.Client/DaprApiException.cs
rynowak: What's your suggestion? I kept the structure of these two types based on the existing places they are thrown and handled in Actors.

src/Dapr.Client/DaprClientGrpc.cs
vinayada1: ok...makes sense

test/Dapr.Client.Test/PublishEventApiTest.cs
vinayada1: Yes, passing null as the value is the ut I am suggesting since that's the case we found was broken and also establishes the behavior of the API if someone is looking. I agree that writing tests that are more like json serializer tests is unnecessary.

src/Dapr.Client/DaprApiException.cs
vinayada1: DaprActorException

vinayada1 (approved):

src/Dapr.Client/DaprApiException.cs
rynowak: Discussed offline.

vinayada1 (dismissed):

vinayada1 (approved):

Issue #559, opened 2021-01-25
Make DaprClient implement IDisposable - update samples and tests


rynowak:
## Expected Behavior

DaprClient should be `IDisposable` because it holds a reference to HttpClient and a gRPC client.

## Actual Behavior

DaprClient is not disposable - resources will only be cleaned up when finalizers run. This isn't the right design approach, and it will be 

## Release Note

RELEASE NOTE: **FIX** DaprClient now implements IDisposable so that networking resources can be reclaimed deterministically.

PR #563, opened 2021-01-26
Add IDisposable to DaprClient

rynowak:
# Description

This change adds disposable support to DaprClient, and updates samples
to dispose it.

I didn't update tests because there are literally hundreds of
non-find-and-replacable cases, and we're not actually doing networking
in our tests so it won't cause an issue.

## Issue reference

Fixes: #559

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


src/Dapr.Client/DaprClient.cs
```diff
@@ -14,11 +14,21 @@ namespace Dapr.Client
...
+    /// </para>
+    /// <para>
+    /// Implementations of <see cref="DaprClient" /> implement <see cref="IDisposable" /> because the client
+    /// accesses network resources. For best performance, create a single long-lived client instance and share
+    /// it for the lifetime of the application. Avoid creating and disposing a client instance for each operation
+    /// that the application performs - this can lead to socket exhaustion and other problems.
+    /// </para>
     /// </summary>
-    public abstract class DaprClient
+    public abstract class DaprClient : IDisposable
```
vinayada1, line 28: Isn't a similar change needed in Actors code?

rynowak: No, I don't think so - actors uses a static instance of HttpClient. It's not ideal, but it's not broken either right now.

vinayada1 (approved):

Issue #573, opened 2021-02-01
Some actors unit tests have a timing bug and fail undeterministically


rynowak:
## Description

We see test failures like:

```
Failed Dapr.Actors.Test.DaprHttpInteractorTest.GetState_ValidateRequest [176 ms]
  Error Message:
   Expected handler.Requests.TryDequeue(out var entry) to be true, but found False.
  Stack Trace:
     at FluentAssertions.Execution.XUnit2TestFramework.Throw(String message)
   at FluentAssertions.Execution.TestFrameworkProvider.Throw(String message)
   at FluentAssertions.Execution.DefaultAssertionStrategy.HandleFailure(String message)
   at FluentAssertions.Execution.AssertionScope.FailWith(Func`1 failReasonFunc)
   at FluentAssertions.Execution.AssertionScope.FailWith(Func`1 failReasonFunc)
   at FluentAssertions.Execution.AssertionScope.FailWith(String message, Object[] args)
   at FluentAssertions.Primitives.BooleanAssertions.BeTrue(String because, Object[] becauseArgs)
   at Dapr.Actors.Test.DaprHttpInteractorTest.GetState_ValidateRequest() in /Users/ryan/github.com/dapr/dotnet-sdk/test/Dapr.Actors.Test/DaprHttpInteractorTest.cs:line 75
```

This is likely the case because methods like `GetStateAsync` yield control before they have enqueued the HTTP message. We should change the pattern in tests to be more robust.

rynowak:
Fixed by #589

Issue #574, opened 2021-02-01
Actors are no longer unit-testable 


rynowak:
## Description

The constructor of `ActorHost` was changed to be internal in an RC build which makes unit testing actors virtually impossible. It's intended that users can construct an `ActorHost` and pass it into an actor in a unit test.

Issue #575, opened 2021-02-01
Make it possible to unit test Actor Timers and Reminder functionality


rynowak:
## Describe the proposal

The Actor class exposes `IActorStateManager` as a public get/set property so that it can be easily mocked in tests. However the equivalent abstraction does not exist for timers and reminders, it's bolted pretty directly onto HTTP client. The current state is that its not feasible to unit test timers and reminders.

We should use the same approach to make these features testable.

PR #576, opened 2021-02-01
Make Actors unit testable

rynowak:
# Description

Changes ActorHost constructor to be public again. Now the state manager
and http interactor are set via properties. This means that code in unit
tests won't be able to cover the methods that interact with timers or
reminders - however this was already the case. Testing state management
is covered by replacing the `IActorStateManager` with a mock.

Logged an issue to improve this.

Also added validation. If you try to use something that's not fully
initialized it will result in an exception with a meaningful message
instead of a null reference.

Also fixed some sources of indeterminism in our tests. This is still
pretty fragile, but for now the tests will at least be reliable.

## Issue reference

Fixes: #573, #574

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/src/Dapr.Actors/Runtime/Actor.cs
+++ b/src/Dapr.Actors/Runtime/Actor.cs
@@ -207,6 +207,8 @@ namespace Dapr.Actors.Runtime
             TimeSpan dueTime,
             TimeSpan period)
         {
+            EnsureInteractorInitialized();
+
             var reminderInfo = new ReminderInfo(state, dueTime, period);
             var reminder = new ActorReminder(this.Id, reminderName, reminderInfo);
             var serializedReminderInfo = await reminderInfo.SerializeAsync();
@@ -223,6 +225,7 @@ namespace Dapr.Actors.Runtime
         /// </returns>
         protected Task UnregisterReminderAsync(IActorReminder reminder)
         {
+            EnsureInteractorInitialized();
             return this.Host.DaprInteractor.UnregisterReminderAsync(this.actorTypeName, this.Id.ToString(), reminder.Name);
         }
 
@@ -235,6 +238,7 @@ namespace Dapr.Actors.Runtime
         /// </returns>
         protected Task UnregisterReminderAsync(string reminderName)
         {
+            EnsureInteractorInitialized();
             return this.Host.DaprInteractor.UnregisterReminderAsync(this.actorTypeName, this.Id.ToString(), reminderName);
         }
 
@@ -263,6 +267,8 @@ namespace Dapr.Actors.Runtime
             TimeSpan dueTime,
             TimeSpan period)
         {
+            EnsureInteractorInitialized();
+
             // Validate that the timer callback specified meets all the required criteria for a valid callback method
             this.ValidateTimerCallback(this.Host, callback);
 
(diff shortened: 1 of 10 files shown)


src/Dapr.Actors/Runtime/Actor.cs
```diff
@@ -340,5 +348,15 @@ internal void ValidateTimerCallback(ActorHost host, string callback)
...
             }
         }
+
+        private void EnsureInteractorInitialized()
+        {
+            if (this.Host.DaprInteractor == null)
+            {
+                throw new InvalidOperationException(
+                    "The actor was initialized without an HTTP client, and so cannot interact with timers or reminders. " +
+                    "This is likely to happen inside a unit test.");
```
rynowak, line 358: As described in the issue, this was already impossible to unit test. Now it just fails with a reasonable error message. I logged #575 to track improving this.

src/Dapr.Actors/Runtime/ActorHost.cs
```diff
@@ -22,21 +22,17 @@ public sealed class ActorHost
         /// <param name="jsonSerializerOptions">The <see cref="JsonSerializerOptions"/> to use for actor state persistence and message deserialization.</param>
         /// <param name="loggerFactory">The logger factory.</param>
         /// <param name="proxyFactory">The <see cref="ActorProxyFactory" />.</param>
-        /// <param name="daprInteractor">The <see cref="IDaprInteractor" />.</param>
-        internal ActorHost(
+        public ActorHost(
```
rynowak, line 25: Users need to be able to access this for unit testing.

src/Dapr.Actors/Runtime/ActorManager.cs
```diff
@@ -263,7 +263,11 @@ internal async Task ActivateActorAsync(ActorId actorId)
         private async Task<ActorActivatorState> CreateActorAsync(ActorId actorId)
         {
             this.logger.LogDebug("Creating Actor of type {ActorType} with ActorId {ActorId}", this.ActorTypeInfo.ImplementationType, actorId);
-            var host = new ActorHost(this.ActorTypeInfo, actorId, this.jsonSerializerOptions, this.loggerFactory, this.proxyFactory, this.daprInteractor);
+            var host = new ActorHost(this.ActorTypeInfo, actorId, this.jsonSerializerOptions, this.loggerFactory, this.proxyFactory)
+            {
+                DaprInteractor = this.daprInteractor,
+                StateProvider = new DaprStateProvider(this.daprInteractor, this.jsonSerializerOptions),
```
rynowak, line 269: The fix for #575 will be to do the same thing for reminders as we do for state. The Actor class is using the http interactor directly instead of a higher level abstraction, so it's not mockable.

src/Dapr.Actors/DaprHttpInteractor.cs
```diff
@@ -78,7 +78,7 @@ HttpRequestMessage RequestFunc()
                 return request;
             }
 
-            return this.SendAsync(RequestFunc, relativeUrl, cancellationToken);
+            await this.SendAsync(RequestFunc, relativeUrl, cancellationToken);
```
rynowak, line 81: This was the fix for the unit test determinism issue. We should ultimately change how we're doing this testing so it's less fragile, but this should at least make the tests pass.

rynowak:
The negative coverage diff is due to skipping tests that have flaky behavior. Restoring these tests is tracked separately.

src/Dapr.Actors/Runtime/ActorHost.cs
vinayada1: Thanks! My earlier change broke this.

vinayada1 (approved):

Issue #582, opened 2021-02-09
Updates for next runtime RC


artursouza:
Update the SDK for the following breaking changes:

1. Service invocation now handles queryparams as string and not Map of string->string. Get the latest proto and fix any code in the SDK that refers to that.

2. When content type is "application/octet-stream", CloudEvent's data will now be in `data_base64` and that should be exposed as `byte[]` to the application.

PR #583, opened 2021-02-10
read binary data

vinayada1:
# Description

Read data_base64 field

#582 

Please reference the issue this PR will close: #582 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/src/Dapr.AspNetCore/CloudEventsMiddleware.cs
+++ b/src/Dapr.AspNetCore/CloudEventsMiddleware.cs
@@ -7,6 +7,7 @@ namespace Dapr
 {
     using System;
     using System.IO;
+    using System.Net;
     using System.Net.Http.Headers;
     using System.Text;
     using System.Text.Json;
@@ -74,57 +75,29 @@ namespace Dapr
             var isDataSet = json.TryGetProperty("data", out var data);
             var isBinaryDataSet = json.TryGetProperty("data_base64", out var binaryData);
 
-            if(isDataSet && isBinaryDataSet)
+            if (isDataSet && isBinaryDataSet)
             {
-                throw new DaprException("Both, data and data_base64 fields set in the Cloudevents envelope. Invalid request");
+                await httpContext.Response.WriteAsync("Both, data and data_base64 fields set in the Cloudevents envelope. Invalid request");
+                httpContext.Response.StatusCode = (int)HttpStatusCode.BadRequest;
+                return;
             }
-            // Data is optional.
             else if (isDataSet)
             {
                 body = new MemoryStream();
                 await JsonSerializer.SerializeAsync<JsonElement>(body, data);
                 body.Seek(0L, SeekOrigin.Begin);
 
-                if (json.TryGetProperty("datacontenttype", out var dataContentType) &&
-                    dataContentType.ValueKind == JsonValueKind.String)
-                {
-                    contentType = dataContentType.GetString();
-
-                    // Since S.T.Json always outputs utf-8, we may need to normalize the data content type
-                    // to remove any charset information. We generally just assume utf-8 everywhere, so omitting
-                    // a charset is a safe bet.
-                    if (contentType.Contains("charset") && MediaTypeHeaderValue.TryParse(contentType, out var parsed))
-                    {
-                        parsed.CharSet = null;
-                        contentType = parsed.ToString();
-                    }
-                }
-                else
-                {
-                    // assume JSON is not specified.
-                    contentType = "application/json";
-                }
+                contentType = this.GetDataContentType(json);
             }
             else if (isBinaryDataSet)
             {
                 // As per the spec, if the implementation determines that the type of data is Binary,
                 // the value MUST be represented as a JSON string expression containing the Base64 encoded
                 // binary value, and use the member name data_base64 to store it inside the JSON object.
-                var decodedBody = Encoding.UTF8.GetString(binaryData.GetBytesFromBase64());
-                try
-                {
-                    JsonDocument.Parse(decodedBody);
-                }
-                catch(JsonException)
-                {
-                    throw new DaprException("Base64 encoded data is not in Json format.");
-                }
-                body = new MemoryStream();
-                var sw = new StreamWriter(body);
-                sw.Write(decodedBody);
-                sw.Flush();
+                var decodedBody = binaryData.GetBytesFromBase64();
+                body = new MemoryStream(decodedBody);
                 body.Seek(0L, SeekOrigin.Begin);
-                contentType = "application/json";
+                contentType = this.GetDataContentType(json);
             }
             else
             {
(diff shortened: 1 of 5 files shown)


src/Dapr.AspNetCore/CloudEventsMiddleware.cs
```diff
@@ -70,8 +70,16 @@ private async Task ProcessBodyAsync(HttpContext httpContext, string charSet)
             string originalContentType;
             string contentType;
 
+            // Check whether to use data or data_base64 as per https://github.com/cloudevents/spec/blob/v1.0.1/json-format.md#31-handling-of-data
+            var isDataSet = json.TryGetProperty("data", out var data);
+            var isBinaryDataSet = json.TryGetProperty("data_base64", out var binaryData);
+
+            if(isDataSet && isBinaryDataSet)
+            {
+                throw new DaprException("Both, data and data_base64 fields set in the Cloudevents envelope. Invalid request");
```
rynowak, line 79: This should not be an exception. This should mark the request as an HTTP 400 and return. An exception turns into a 500 - a lot of teams using ASP.NET Core in production would treat a 500 as an incident to investigate, so you don't want to let bad data sent by a client turn into a 500.

```C#
httpContext.Response.StatusCode = 400;
return
```

src/Dapr.AspNetCore/CloudEventsMiddleware.cs
```diff
@@ -70,8 +70,16 @@ private async Task ProcessBodyAsync(HttpContext httpContext, string charSet)
             string originalContentType;
             string contentType;
 
+            // Check whether to use data or data_base64 as per https://github.com/cloudevents/spec/blob/v1.0.1/json-format.md#31-handling-of-data
+            var isDataSet = json.TryGetProperty("data", out var data);
+            var isBinaryDataSet = json.TryGetProperty("data_base64", out var binaryData);
+
+            if(isDataSet && isBinaryDataSet)
```
rynowak, line 77: format

src/Dapr.AspNetCore/CloudEventsMiddleware.cs
```diff
@@ -70,8 +70,16 @@ private async Task ProcessBodyAsync(HttpContext httpContext, string charSet)
...
+            {
+                throw new DaprException("Both, data and data_base64 fields set in the Cloudevents envelope. Invalid request");
+            }
             // Data is optional.
-            if (json.TryGetProperty("data", out var data))
+            else if (isDataSet)
```
rynowak, line 82: The comment at line 81 doesn't make sense here, it should be moved or deleted.

src/Dapr.AspNetCore/CloudEventsMiddleware.cs
```diff
@@ -97,6 +105,27 @@ private async Task ProcessBodyAsync(HttpContext httpContext, string charSet)
                     contentType = "application/json";
                 }
             }
+            else if (isBinaryDataSet)
+            {
+                // As per the spec, if the implementation determines that the type of data is Binary,
+                // the value MUST be represented as a JSON string expression containing the Base64 encoded
+                // binary value, and use the member name data_base64 to store it inside the JSON object.
+                var decodedBody = Encoding.UTF8.GetString(binaryData.GetBytesFromBase64());
```
rynowak, line 113: It's not safe to do `Encoding.UTF8.GetString(...)` on the return value here, the point is that it's *just bytes* and you can't make any assumptions about the format.

src/Dapr.AspNetCore/CloudEventsMiddleware.cs
```diff
@@ -97,6 +105,27 @@ private async Task ProcessBodyAsync(HttpContext httpContext, string charSet)
...
+                catch(JsonException)
+                {
+                    throw new DaprException("Base64 encoded data is not in Json format.");
+                }
+                body = new MemoryStream();
+                var sw = new StreamWriter(body);
+                sw.Write(decodedBody);
+                sw.Flush();
+                body.Seek(0L, SeekOrigin.Begin);
+                contentType = "application/json";
```
rynowak, line 127: This doesn't seem right. Remember that the `datacontenttype` field contains a base64 encoded representation of some arbitrary content. 

Consider what happens when you take an image in `.png` format and try to use an event to send it. The cloudevent should look like this:

```json
{
  ...
  "data_base64": "aW1hZ2VfZGF0YQo=",
  "datacontenttype": "application/octet-stream",
  ...
}
```

This should retrieve the bytes (`GetBytesFromBase64`) and write them to the stream.

Also the content type should be set based on `datacontenttype`.

rynowak: You can't make any assumptions of what `aW1hZ2VfZGF0YQo=` means (in this example), that's the whole point of this feature.

src/Dapr.AspNetCore/CloudEventsMiddleware.cs
```diff
@@ -70,32 +71,33 @@ private async Task ProcessBodyAsync(HttpContext httpContext, string charSet)
...
 
-            // Data is optional.
-            if (json.TryGetProperty("data", out var data))
+            // Check whether to use data or data_base64 as per https://github.com/cloudevents/spec/blob/v1.0.1/json-format.md#31-handling-of-data
+            var isDataSet = json.TryGetProperty("data", out var data);
+            var isBinaryDataSet = json.TryGetProperty("data_base64", out var binaryData);
+
+            if (isDataSet && isBinaryDataSet)
+            {
+                await httpContext.Response.WriteAsync("Both, data and data_base64 fields set in the Cloudevents envelope. Invalid request");
```
rynowak, line 80: I would suggest not doing this. Putting some content inside the request often raises more questions than it answers - like how would a machine understand this? What's the right format, etc. It's better to just avoid having an opinion on this.

If you want to make this visible to the **developer** of the application, then logging is the right way to do that. You could do that by injecting an `ILogger<CloudEventsMiddleware>` in the constructor.

src/Dapr.AspNetCore/CloudEventsMiddleware.cs
```diff
@@ -70,32 +71,33 @@ private async Task ProcessBodyAsync(HttpContext httpContext, string charSet)
...
+            }
+            else if (isBinaryDataSet)
+            {
+                // As per the spec, if the implementation determines that the type of data is Binary,
+                // the value MUST be represented as a JSON string expression containing the Base64 encoded
+                // binary value, and use the member name data_base64 to store it inside the JSON object.
+                var decodedBody = binaryData.GetBytesFromBase64();
+                body = new MemoryStream(decodedBody);
+                body.Seek(0L, SeekOrigin.Begin);
+                contentType = this.GetDataContentType(json);
```
rynowak, line 100: Perfect 👌

rynowak (approved):

Issue #588, opened 2021-02-14
Actors DaprHttpInteractor mismanages network resources


rynowak:
## Expected Behavior

Actors requires few network resources. It should be able to use a single connection pool to communicate with the Dapr sidecar.

## Actual Behavior

Each Actor type gets its own interactor, and thus it's own pool of connections.

An extra `HttpClientHandler` is created (but not disposed) for each actor.

The constructor overrides a mutable static each time this happens, so there's no reference to all of the interactors that have already been created.

## Release Note

RELEASE NOTE: **FIX** Dapr Actors will now user fewer TCP connections to accomplish the same work.

rynowak:
Fixed by #589

PR #589, opened 2021-02-15
Fix pattern for tests that use HttpClient

rynowak:
# Description

This change introduces a better API for us to test the DaprClient or
other HttpClient based APIs.

This will resolve the flakiness problems that we're seeing with some of
the actors tests.

Additionally fixed an issue where DaprHttpInteractor was misuing
HttpClientHandler. This would result in a new handler being created when
it isn't needed.

## Issue reference

Fixes: #573
Fixes: #588

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


src/Dapr.Actors/Client/ActorProxyFactory.cs
```diff
@@ -32,7 +32,16 @@ public ActorProxyOptions DefaultOptions
...
+        [Obsolete("Use the constructor that accepts HttpMessageHandler. This will be removed in the future.")]
+        public ActorProxyFactory(ActorProxyOptions options, HttpClientHandler handler)
+            : this(options, (HttpMessageHandler)handler)
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ActorProxyFactory"/> class.
+        /// </summary>
+        public ActorProxyFactory(ActorProxyOptions options = null, HttpMessageHandler handler = null)
```
rynowak, line 44: The way that this API was design isn't very flexible/testable. `HttpClientHandler` isn't a more specific type than `HttpMessageHandler`. 

The API change in this PR is necessary to make this code testable, and is done in a way that's binary-compatible. It's safe to remove a default value for an optional parameter as long as you provide a *better* overload. Only users that were explicitly passing a `HttpClientHandler` will be impacted by this, and they could just choose to suppress the warning.

src/Dapr.Actors/DaprHttpInteractor.cs
```diff
@@ -28,19 +28,20 @@ internal class DaprHttpInteractor : IDaprInteractor
         private readonly JsonSerializerOptions jsonSerializerOptions = JsonSerializerDefaults.Web;
         private const string DaprEndpoint = Constants.DaprDefaultEndpoint;
         private readonly string daprPort;
-        private static HttpClientHandler innerHandler = new HttpClientHandler();
-        private HttpClient httpClient = null;
-        private bool disposed = false;
+        private readonly static HttpMessageHandler defaultHandler = new HttpClientHandler();
```
rynowak, line 31: This was the root cause of #588 and possibly also of #573.

rynowak: ⚠️ ⚠️ ⚠️ mutable statics and dangerous ⚠️ ⚠️ ⚠️

test/Dapr.Actors.Test/ApiTokenTests.cs
```diff
@@ -16,107 +14,96 @@ namespace Dapr.Actors.Test
...
-            handler.Requests.TryDequeue(out var entry).Should().BeTrue();
-            Action action = () => entry.Request.Headers.GetValues("dapr-api-token");
-            action.Should().Throw<InvalidOperationException>();
+
+            var request = await client.CaptureHttpRequestAsync(async handler =>
+            {
+                var factory = new ActorProxyFactory(null, handler);
+                var proxy = factory.CreateActorProxy<ITestActor>(actorId, "TestActor");
+                await proxy.SetCountAsync(1, new CancellationToken());
+            });
```
rynowak, line 53: This design approach lets us capture HTTP or gRPC traffic without timing issues. This is very robust and applies timeouts to everything that happens, so we can't end up with tests that hang indefinitely.

test/Dapr.Actors.Test/ApiTokenTests.cs
```diff
@@ -16,107 +14,96 @@ namespace Dapr.Actors.Test
...
+
+            request.Dismiss();
```
rynowak, line 55: The other cool trick here is that we'll throw if you fail to *acknowledge* a request. This way we make sure that extra stuff isn't happening that's going untested.

rynowak:
@vinayada1 - suggestion here is that we hold this for a 1.0.1 patch of the SDK and release it as soon as the dust settles from 1.0.0.

vinayada1 (approved):

PR #591, opened 2021-02-15
Add better test reporting

rynowak:
Our experience right now is pretty bad when tests fail, you get a
generic `failed with exit code 1`.

This change adds more full-featured integration with GitHub Actions as
well as fixes a few minor issues:

- We were missing the codecov package on one project
- Updates to test SDK
- Annotations on failing tests
- Each test project as its own status check
- Separating and parallelizing steps of the build


rynowak:
Example of an annotation:

image

If you want to see an example of the "checks" look no further than this checks for this run.

rynowak:
@vinayada1

rynowak:
Note that we'll have some cleanup to do if we merge this:

image

We'll need to remove `Build Projects` as well as mark a few other things as required.

.github/workflows/sdk_build.yml
```diff
@@ -15,61 +15,119 @@ on:
...
+    steps:
+    - uses: actions/checkout@v1
+    - name: Parse release version
+      run: python ./.github/scripts/get_release_version.py
+    - name: Setup .NET Core 3.1
+      uses: actions/setup-dotnet@v1
+      if: matrix.install-3
+      with:
+        dotnet-version: 3.1.x
+    - name: Setup .NET 5.0
```
mukundansundar, line 66: shouldn't this be checked against `install-3` being false?

vinayada1 (approved):

.github/workflows/sdk_build.yml
rynowak: No. We need 5.0 regardless to build. The project here multi-target (both 3 and 5). 5 can build 3, but 3 can't build 5.

PR #602, opened 2021-02-20
End to end test for service invocation

vinayada1:
# Description

Add e2e test for service invocation

#389 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


test/Dapr.E2E.Test.App/Controllers/TestController.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        private readonly ILogger<TestController> logger;
+
+        /// <summary>
+        /// Echoes the input string
+        /// </summary>
+        /// <param name="name">string to echo.</param>
+        /// <returns>string with the input name in it.</returns>
+        [HttpGet("hello/{name}")]
+        public ActionResult<string> Get(string name)
```
rynowak, line 38: Bigtime suggestion - use something other than a string. 

Users won't really use a primitive type like a string as a top-level object. By including a real class, you expose more the JSON serializer for testing.

test/Dapr.E2E.Test.App/Dapr.E2E.Test.App.csproj
```diff
@@ -0,0 +1,15 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Folder Include="Controllers\" />
+  </ItemGroup>
```
rynowak, line 9: can remove this ItemGroup

test/Dapr.E2E.Test.App/Dapr.E2E.Test.App.csproj
```diff
@@ -0,0 +1,15 @@
...
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Folder Include="Controllers\" />
+  </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\Dapr.AspNetCore\Dapr.AspNetCore.csproj">
+      <GlobalPropertiesToRemove></GlobalPropertiesToRemove>
```
rynowak, line 12: Can remove `GlobalPropertiesToRemove` this doesn't need to be here.

test/Dapr.E2E.Test/Dapr.E2E.Test.csproj
```diff
@@ -0,0 +1,28 @@
...
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.4.1">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
+  </ItemGroup>
+
+  <ItemGroup>
+    <Compile Remove="..\Shared\GrpcUtils.cs" />
+    <Compile Remove="..\Shared\TestHttpClient.cs" />
+  </ItemGroup>
```
rynowak, line 19: Do you know why this is here? You shouldn't need to remove these if they aren't included in the first place.

test/Dapr.E2E.Test/Dapr.E2E.Test.csproj
```diff
@@ -0,0 +1,28 @@
...
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\Dapr.Extensions.Configuration\Dapr.Extensions.Configuration.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.Configuration" Version="3.1.2" />
+  </ItemGroup>
```
rynowak, line 26: You also shouldn't need this, it should be included by the project reference.

test/Dapr.E2E.Test/DaprApp.cs
```diff
@@ -0,0 +1,99 @@
...
+    public class DaprApp
+    {
+        // private string pathSeparator;
+        static string shellExeName = SetShell();
+        private string appId;
+        private int appPort;
+        const string outputToMatchOnStart = "You're up and running";
+        const string outputToMatchOnStop = "app stopped successfully";
+
+        static char pathSeparator = System.IO.Path.DirectorySeparatorChar;
```
rynowak, line 23: btw you can just do `using static System.IO.Path.DirectorySeparatorChar` and then reference `DirectorySeparatorChar`

test/Dapr.E2E.Test/DaprApp.cs
```diff
@@ -0,0 +1,99 @@
...
+        public DaprApp(string appId, int appPort = 0)
+        {
+            this.appId = appId;
+            this.appPort = appPort;
+        }
+
+        public (string, string) Start()
+        {
+            var (httpPort, grpcPort, metricsPort) = GetFreePorts();
+            var componentsPath = string.Format($"..{DaprApp.pathSeparator}..{DaprApp.pathSeparator}..{DaprApp.pathSeparator}..{DaprApp.pathSeparator}..{DaprApp.pathSeparator}test{DaprApp.pathSeparator}Dapr.E2E.Test{DaprApp.pathSeparator}components");
```
rynowak, line 43: Use `Path.Combine` for this.

test/Dapr.E2E.Test/DaprApp.cs
```diff
@@ -0,0 +1,99 @@
...
+            var daprStartCommand = string.Format($"dapr run --app-id {appId} --dapr-http-port {httpPort} --dapr-grpc-port {grpcPort} --metrics-port {metricsPort} --components-path .{DaprApp.pathSeparator}{componentsPath}");
```
rynowak, line 44: `.{DaprApp.pathSeparator}{componentsPath}` is a little confusing, why not just resolve to a full path? https://docs.microsoft.com/en-us/dotnet/api/system.io.path?view=net-5.0 would help

test/Dapr.E2E.Test/DaprApp.cs
```diff
@@ -0,0 +1,99 @@
...
+            if(this.appPort != 0)
```
rynowak, line 45: please auto-format this (`cmd+shift+p`)-> `Format Document`

test/Dapr.E2E.Test/DaprApp.cs
```diff
@@ -0,0 +1,99 @@
...
+            {
+                daprStartCommand += string.Format($" --app-port {appPort}");
```
rynowak, line 47: So you don't need both `string.Format` and `$""` 😄 - they both do string formatting, just use `$""`

test/Dapr.E2E.Test/DaprApp.cs
```diff
@@ -0,0 +1,99 @@
...
+            var projectPath = string.Format($"..{DaprApp.pathSeparator}..{DaprApp.pathSeparator}..{DaprApp.pathSeparator}..{DaprApp.pathSeparator}..{DaprApp.pathSeparator}test{DaprApp.pathSeparator}Dapr.E2E.Test.App{DaprApp.pathSeparator}Dapr.E2E.Test.App.csproj");
+            daprStartCommand += string.Format($" -- dotnet run --project {projectPath}");
+            var daprStart = new ShellCommand()
+            {
+                ShellExeName = DaprApp.shellExeName,
+                Command = daprStartCommand,
+                OutputToMatch = outputToMatchOnStart,
+                Timeout = 5000
+            };
+            daprStart.Run();
```
rynowak, line 58: I'd suggest using `System.Diagnostics.Process` unless you specifically need shell functionality. https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process?view=net-5.0

rynowak: Or: https://github.com/dotnet/command-line-api/blob/main/src/System.CommandLine/Invocation/Process.cs

This is a little more full-featured and we could just add a dependency on `System.CommandLine`.

test/Dapr.E2E.Test/DaprApp.cs
```diff
@@ -0,0 +1,99 @@
...
+            daprStop.Run();
+            Console.WriteLine("Dapr App stopped successfully");
+        }
+
+        private static (int, int, int) GetFreePorts()
+        {
+            IPAddress ip = IPAddress.Loopback;
+            var listeners = new TcpListener[3];
+            var ports = new int[3];
+            foreach (int i in Enumerable.Range(0, 3))
```
rynowak, line 84: you can also use a `for` loop 😆

test/Dapr.E2E.Test/ServiceInvocation/InvokeMethodTest.cs
```diff
@@ -0,0 +1,30 @@
...
+    using System.Threading.Tasks;
+    using Dapr.Client;
+    using Xunit;
+
+    public class Main
+    {
+        [Fact]
+        public async void TestServiceInvocation()
+        {
+            var testApp = new DaprApp("testapp", 5000);
```
rynowak, line 16: Make this implement `IDisposable` and do `using var testapp ...`

test/Dapr.E2E.Test/ServiceInvocation/InvokeMethodTest.cs
```diff
@@ -0,0 +1,30 @@
...
+    using System.Text.Json;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Dapr.Client;
+    using Xunit;
+
+    public class Main
+    {
+        [Fact]
+        public async void TestServiceInvocation()
```
rynowak, line 14: `async void` - this should be `async Task`.

test/Dapr.E2E.Test/ServiceInvocation/InvokeMethodTest.cs
```diff
@@ -0,0 +1,30 @@
+namespace Dapr.E2E.Test
```
rynowak, line 1: Add Copyright

test/Dapr.E2E.Test/DaprApp.cs
```diff
@@ -0,0 +1,99 @@
...
+            var daprStopCommand = string.Format($"dapr stop --app-id {appId}");
+            var daprStop = new ShellCommand()
+            {
+                ShellExeName = DaprApp.shellExeName,
+                Command = daprStopCommand,
+                OutputToMatch = outputToMatchOnStop,
+                Timeout = 10000
+            };
+            daprStop.Run();
+            Console.WriteLine("Dapr App stopped successfully");
```
rynowak, line 76: Generally in tests writing to the console does nothing. Since we're using xUnit - if you **actually** want to output something you need to use `ITestOutputHelper`. https://xunit.net/docs/capturing-output

test/Dapr.E2E.Test/DaprApp.cs
```diff
@@ -0,0 +1,99 @@
...
+            foreach (int i in Enumerable.Range(0, 3))
+            {
+                listeners[i] = new TcpListener(ip, 0);
+                listeners[i].Start();
+                ports[i] = ((IPEndPoint)listeners[i].LocalEndpoint).Port;
+            }
+
+            foreach (int i in Enumerable.Range(0, 3))
+            {
+                listeners[i].Stop();
```
rynowak, line 93: Is it important to have all 3 listeners alive at once? 

- If not, can we simplify this code
- If it is important, add a comment

test/Dapr.E2E.Test/ShellCommand.cs
```diff
@@ -0,0 +1,52 @@
+namespace Dapr.E2E.Test
+{
+    using System;
+    using System.Diagnostics;
+    using System.Threading;
+
+    public class ShellCommand
```
rynowak, line 7: I should we should just use https://github.com/dotnet/command-line-api/blob/main/src/System.CommandLine/Invocation/Process.cs

Add the dependency, don't copy the code 😆

test/Dapr.E2E.Test/ServiceInvocation/InvokeMethodTest.cs
```diff
@@ -0,0 +1,30 @@
...
+{
+    using System;
+    using System.Net.Http.Json;
+    using System.Text.Json;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Dapr.Client;
+    using Xunit;
+
+    public class Main
```
rynowak, line 11: This should be something other than `Main`

test/Dapr.E2E.Test/ServiceInvocation/InvokeMethodTest.cs
rynowak: It might be better to auto-assign the port for the APP too. You can configure the listening address for ASP.NET Core like `--urls http://localhost:5000`

test/Dapr.E2E.Test/ServiceInvocation/InvokeMethodTest.cs
```diff
@@ -0,0 +1,30 @@
...
+    {
+        [Fact]
+        public async void TestServiceInvocation()
+        {
+            var testApp = new DaprApp("testapp", 5000);
+            try
+            {
+                var (daprHttpEndpoint, _) = testApp.Start();
+                var client = DaprClient.CreateInvokeHttpClient(appId: "testapp", daprEndpoint: daprHttpEndpoint);
```
rynowak, line 20: Should there be a dapr side process for the test as well? This is going to use a single sidecar, but in the product scenarios there would be two right? How does Java approach this?

rynowak (changes requested):
This is a good start. Have you looked in how the CI runs this? This seems to rely on a Dapr on the path and already initialized. I'd expect to see the logic as well to acquire a specific build of Dapr (can copy from Java).

test/Dapr.E2E.Test/Dapr.E2E.Test.csproj
vinayada1: I used VS for mac to create these projects and it added a bunch of stuff...

test/Dapr.E2E.Test/DaprApp.cs
vinayada1: We need to start them at once. Otherwise, the code might end up returning the same port for all three values. Added a comment

test/Dapr.E2E.Test/ServiceInvocation/InvokeMethodTest.cs
vinayada1: I thought the tests should focus on how sdk interacts with the dapr side car. Java tests use this same approach.

vinayada1:
> This is a good start. Have you looked in how the CI runs this? This seems to rely on a Dapr on the path and already initialized. I'd expect to see the logic as well to acquire a specific build of Dapr (can copy from Java).
> 
Java tests have some workflow actions setup to install dapr cli and dapr with a specific version. We can reuse tha

test/Dapr.E2E.Test/ServiceInvocation/InvokeMethodTest.cs
rynowak: If this is the approach that java uses I think that's fine. I dont' want to reinvent the wheel.

test/Dapr.E2E.Test.App/Dapr.E2E.Test.App.csproj
```diff
@@ -0,0 +1,11 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
```
rynowak, line 4: Can this multitarget?

`<TargetFrameworks>netcoreapp3.1;net5</TargetFrameworks>

test/Dapr.E2E.Test/ServiceInvocation/InvokeMethodTest.cs
```diff
@@ -0,0 +1,56 @@
...
+    using System.Net.Http.Json;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Dapr.Client;
+    using Xunit;
+
+    public class ServiceInvocationTests : IDisposable
+    {
+        private DaprApp testApp;
+        public ServiceInvocationTests()
```
rynowak, line 17: I'd suggest accepting `ITestOutputHelper` here and passing it to `DaprApp`. This will allow you to capture all of the console output and pipe it to the test logs.

rynowak: https://xunit.net/docs/capturing-output

test/Dapr.E2E.Test/DaprTestApp.cs
```diff
@@ -0,0 +1,107 @@
...
+            var projectPath = Combine(".", "..", "..", "..", "..", "..", "test", "Dapr.E2E.Test.App", "Dapr.E2E.Test.App.csproj");
+            var daprDotnetCommand = $" -- dotnet run --project {projectPath} --framework {GetTargetFrameworkName()}";
+            if (this.useAppPort)
+            {
+                daprStartCommand += $" --app-port {appPort}";
+                daprDotnetCommand += $" --urls http://localhost:{appPort}";
+            }
+            daprStartCommand += daprDotnetCommand;
+
+            var daprStart = new DaprCommand()
```
rynowak, line 48: Can we pass in `testOutput` here and capture the process output as well?

test/Dapr.E2E.Test/ServiceInvocation/AVeryCoolWorkAroundTest.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace Dapr.E2E.WorkAround
+{
+    using Xunit;
+
+    public class WorkAroundTests
+    {
+        [Fact]
+        public void AVeryCoolWorkAroundTest()
+        {
+            Assert.True(true);
```
rynowak, line 17: very cool 😎

rynowak (approved):

test/Dapr.E2E.Test/ServiceInvocation/AVeryCoolWorkAroundTest.cs
vinayada1: Thanks for the suggestion!

Issue #615, opened 2021-03-03
Usage of nested configuration in Kubernetes Secret


Fabian-Schmidt:
I am trying to use the secret management in a Kubernetes cluster with a nested configuration object.

I am using the [`DaprSecretStoreConfigurationProvider`](https://github.com/dapr/dotnet-sdk/blob/master/src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationProvider.cs) with bulk secret reading.

A secret in Kubernetes config key must consist of alphanumeric characters, '-', '_' or '.' (e.g. 'key.name',  or 'KEY_NAME',  or 'key-name', regex used for validation is '[-._a-zA-Z0-9]+')

A secret value must be a flat string.

How is the intended usage with a nested .NET config in this scenario? As the normally used `:` is not allowed for Kubernetes secrets.

rynowak:
This seems like a fine improvement. Since this is a change in behavior we have to add an opt-out. Using `__` is unusual enough and specific enough to ASP.NET Core that I'm comfortable with an opt-out fix.

PR #616, opened 2021-03-04
Separator character for Dapr Sercet Store Configuration Provider

Fabian-Schmidt:
# Description

Support for nested configuration when secret store does not support configuration key delimiter `:`. The only GA secret store Kubernetes does not support `:` in secret key name.

Adopt same pattern as Environment Variable configuration uses. Replace `__` by `:`.

## Issue reference

Please reference the issue this PR will close: #615

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


rynowak (changes requested):
This is a good start, and looks fine overall. 

Can you add a setting to `DaprSecretStoreConfigurationSource` that disables this behavior? This is a behavior change, so if we're going to accept it there has to be a solution for someone who relies on the old behavior.

rynowak (approved):
Awesome! Thanks @Fabian-Schmidt

PR #620, opened 2021-03-05
Suppress TRX parser on PR from fork

rynowak:


.github/workflows/itests.yml
```diff
@@ -135,7 +135,7 @@ jobs:
       - name: Parse Trx files
         uses: NasAmin/trx-parser@v0.1.0
         id: trx-parser
-        if: github.repository == 'dapr/dotnet-sdk'
```
rynowak, line 138: apparently this would always refer to this repo, and thus always be true.

vinayada1 (dismissed):

vinayada1 (approved):

PR #622, opened 2021-03-07
Unit testability for timers and reminders

rynowak:
# Description

This change adds abstractions for Actors timer and reminder
functionality, making them possible to unit test. Previously these
features were wired from the Actor class directly into the client
functionality, and there was no way for a user of actors to write their
own tests when they include this functionality in their code.

Making this work required a lot of munging of basic data types because
we need to *extract* a layer of functionality into a testable API.

I also added a helper method for tests to create an actor host with
mocks attached. We'll likely need to do more in the future to support
testability for actors. I started this process by adding some "unit
test" unit tests. The only way to test that you can test is by testing
:).

----

Additionally I added E2E tests for actors timers and reminders and
improved some of the existing test functionality.

The big investment here is to change the layering of tests to ensure
that we only need to start one instance of the test app at a time.
Previously we'd start and stop the test app as part of the only test we
have. Adding more tests means we want to avoid this overhead when
possible.

This is specifically a challenge for actors since the assocation of an
actor type to an app id is **global state**. We want to keep the app
process we're using for tests long-lived to avoid classic distributed
systems problems (like the host going away) inside our tests.

## Issue reference

Fixes: #575

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


src/Dapr.Actors/Client/ActorProxyFactory.cs
```diff
@@ -55,22 +55,26 @@ public TActorInterface CreateActorProxy<TActorInterface>(ActorId actorId, string
         /// <inheritdoc/>
         public ActorProxy Create(ActorId actorId, string actorType, ActorProxyOptions options = null)
         {
+            options ??= this.DefaultOptions;
```
rynowak, line 58: This was just plain wrong. Passing in a new "options" is supposed to win over the defaults.

src/Dapr.Actors.AspNetCore/ActorsServiceCollectionExtensions.cs
```diff
@@ -52,6 +52,7 @@ public static void AddActors(this IServiceCollection services, Action<ActorRunti
                     {
                         JsonSerializerOptions = options.JsonSerializerOptions,
                         DaprApiToken = options.DaprApiToken,
+                        HttpEndpoint = options.HttpEndpoint,
```
rynowak, line 55: One of the impactful changes I had to make here is making the HTTP endpoint used by the actor client configurable. This was not configurable outside of the env-var.

src/Dapr.Actors/Client/ActorProxyOptions.cs
```diff
@@ -34,14 +41,20 @@ public JsonSerializerOptions JsonSerializerOptions
...
+        /// The URI endpoint to use for HTTP calls to the Dapr runtime. The default value will be 
+        /// <c>http://127.0.0.1:DAPR_HTTP_PORT</c> where <c>DAPR_HTTP_PORT</c> represents the value of the 
+        /// <c>DAPR_HTTP_PORT</c> environment variable.
+        /// </remarks>
+        /// <value></value>
+        public string HttpEndpoint { get; set; } = GetDefaultHttpEndpoint();
+
+        private static string GetDefaultHttpEndpoint()
         {
-            if(string.IsNullOrWhiteSpace(this.DaprApiToken))
```
rynowak, line 41: I'm not sure why this check was here. Inside the constructor this will always be null.

src/Dapr.Actors/Runtime/Actor.cs
```diff
@@ -225,8 +221,8 @@ protected virtual Task OnPostActorMethodAsync(ActorMethodContext actorMethodCont
         /// </returns>
         protected Task UnregisterReminderAsync(IActorReminder reminder)
         {
-            EnsureInteractorInitialized();
-            return this.Host.DaprInteractor.UnregisterReminderAsync(this.actorTypeName, this.Id.ToString(), reminder.Name);
+            var token = reminder is ActorReminderToken ? (ActorReminderToken)reminder : new ActorReminderToken(this.actorTypeName, this.Id, reminder.Name);
```
rynowak, line 224: This gets a little confusing, hopefully this will clear it up. For both actors and reminders there are a few existing data types in use. None of them really fit the bill for abstracting this functionality. However we're stuck supporting the types that are part the public API.

Ex:

To register a timer you need:

- actor type
- actor id
- name
- callback
- state
- due time
- period

To unregister a timer you need:

- actor type
- actor id
- name

None of the existing types has all of those things for either case 😱 so none of them are useful for defining a new abstraction. I was able to fix it by making `ActorTimer` and `ActorReminder` more beefy, and extracting base classes `ActorTimerToken`, `ActorReminderToken`.

All of the other types like `IActorReminder`, `TimerInfo`, `ReminderInfo` aren't that interesting for users. This is an example of how challenging it can be to carve out an abstraction after the fact 😉

src/Dapr.Actors/Runtime/Actor.cs
```diff
@@ -348,15 +338,5 @@ internal void ValidateTimerCallback(ActorHost host, string callback)
...
         }
-
-        private void EnsureInteractorInitialized()
-        {
-            if (this.Host.DaprInteractor == null)
-            {
-                throw new InvalidOperationException(
-                    "The actor was initialized without an HTTP client, and so cannot interact with timers or reminders. " +
-                    "This is likely to happen inside a unit test.");
-            }
```
rynowak, line 359: Prior to this change we'd just tell you explicitly that this functionality can't be tested.

src/Dapr.Actors/Runtime/ActorHost.cs
```diff
@@ -14,6 +15,42 @@ namespace Dapr.Actors.Runtime
     /// </summary>
     public sealed class ActorHost
     {
+        /// <summary>
+        /// Creates an instance of <see cref="ActorHost" /> for unit testing an actor instance.
+        /// </summary>
+        /// <param name="options">The <see cref="ActorTestOptions" /> for configuring the host.</param>
+        /// <typeparam name="TActor">The actor type.</typeparam>
+        /// <returns>An <see cref="ActorHost" /> instance.</returns>
+        public static ActorHost CreateForTest<TActor>(ActorTestOptions options = null)
```
rynowak, line 24: This is where you can plug in mocks. Actors needs an `ActorHost`.

src/Dapr.Actors/Runtime/ActorHost.cs
```diff
@@ -60,8 +122,11 @@ public sealed class ActorHost
         /// </summary>
         public IActorProxyFactory ProxyFactory { get; }
 
-        internal DaprStateProvider StateProvider { get; set; }
+        /// <summary>
+        /// Gets the <see cref="ActorTimerManager" />.
+        /// </summary>
+        public ActorTimerManager TimerManager { get; }
 
-        internal IDaprInteractor DaprInteractor { get; set; }
```
rynowak, line 65: begone found beast

src/Dapr.Actors/Runtime/ActorHost.cs
```diff
@@ -60,8 +122,11 @@ public sealed class ActorHost
...
+        internal DaprStateProvider StateProvider { get; set; }
```
rynowak, line 130: The state provider needs to get the same treatment. It's *somewhat* possible to mock the state provider today.

src/Dapr.Actors/Runtime/ActorReminder.cs
```diff
@@ -3,41 +3,74 @@
...
         }
 
-        public TimeSpan Period
-        {
-            get { return this.ReminderInfo.Period; }
-        }
+        /// <summary>
+        /// Gets the reminder state.
+        /// </summary>
+        public byte[] State { get; }
```
rynowak, line 64: Unfortunately the **existing** public API has inconsistent terminology (state vs data). Timers call this `Data`

src/Dapr.Actors/Runtime/ActorTestOptions.cs
```diff
@@ -0,0 +1,97 @@
...
+        /// Gets or sets the <see cref="IActorProxyFactory" />.
+        /// </summary>
+        public IActorProxyFactory ProxyFactory { get; set; } = new InvalidProxyFactory();
+
+        /// <summary>
+        /// Gets or sets the <see cref="ActorTimerManager" />.
+        /// </summary>
+        public ActorTimerManager TimerManager { get; set; } = new InvalidTimerManager();
+
+        private class InvalidProxyFactory : IActorProxyFactory
```
rynowak, line 48: Note, it's now also possible to mock the proxy factory. So you can mock calls to other actors.

src/Dapr.Actors/Runtime/ActorTimerManager.cs
```diff
@@ -0,0 +1,43 @@
...
+// ------------------------------------------------------------
+
+using System.Threading.Tasks;
+
+namespace Dapr.Actors.Runtime
+{
+    /// <summary>
+    /// Provides timer and reminder functionality to an actor instance.
+    /// </summary>
+    public abstract class ActorTimerManager
```
rynowak, line 13: New abstraction for mocking/testing. This isolates the user-level APIs from the implementation (facade).

src/Dapr.Actors/Runtime/TimerInfo.cs
```diff
@@ -8,16 +8,15 @@ namespace Dapr.Actors.Runtime
...
     using System.Text.Json;
     using System.Text.Json.Serialization;
     using System.Threading;
-    using System.Threading.Tasks;
     using Dapr.Actors.Resources;
 
     /// <summary>
     /// Represents the details of the timer set on an Actor.
     /// </summary>
+    [Obsolete("This class is an implementation detail of the framework and will be made internal in a future release.")]
```
rynowak, line 19: Unfortunately this was public, despite there not really being anything useful to do with it.

test/Dapr.Actors.Test/ActorUnitTestTests.cs
```diff
@@ -0,0 +1,159 @@
...
+using Dapr.Actors.Client;
+using Dapr.Actors.Runtime;
+using Microsoft.Extensions.Logging.Abstractions;
+using Moq;
+using Xunit;
+
+namespace Dapr.Actors
+{
+    // Tests that test that we can test... hmmmm...
+    public class ActorUnitTestTests
```
rynowak, line 19: Here's the good stuff.

rynowak: This is literally just an example test to provide that it's possible to write tests. We probably need 3-4 more of these for other features.

test/Dapr.Actors.Test/DaprFormatTimeSpanTests.cs
```diff
@@ -3,6 +3,7 @@
 // Licensed under the MIT License.
 // ------------------------------------------------------------
 
+#pragma warning disable 0618
```
rynowak, line 6: `TimerInfo` is obsolete. These are dotted around. When we do a major release we'll make it internal and clean this up.

test/Dapr.Actors.Test/Runtime/ActorManagerTests.cs
```diff
@@ -17,7 +17,8 @@ public sealed class ActorManagerTests
         private ActorManager CreateActorManager(Type type, ActorActivator activator = null)
         {
             var registration = new ActorRegistration(ActorTypeInformation.Get(type));
-            return new ActorManager(registration, activator ?? new DefaultActorActivator(), JsonSerializerDefaults.Web, NullLoggerFactory.Instance, ActorProxy.DefaultProxyFactory, new DaprHttpInteractor());
+            var interactor = new DaprHttpInteractor(clientHandler: null, "http://localhost:3500", apiToken: null);
```
rynowak, line 20: Eventually we'll get the client APIs out of all of these tests. We need to create a similar abstraction for the state manager.

test/Dapr.E2E.Test.Actors/IPingActor.cs
```diff
@@ -0,0 +1,15 @@
...
+
+using System.Threading.Tasks;
+using Dapr.Actors;
+
+namespace Dapr.E2E.Test.Actors
+{
+    public interface IPingActor : IActor
+    {
+        Task Ping();
+    }
```
rynowak, line 14: This is the best way I could find that verify that the actors runtime is healthy.

test/Dapr.E2E.Test/DaprTestApp.cs
```diff
@@ -19,8 +21,8 @@ public class DaprTestApp
         static string daprBinaryName = "dapr";
         private string appId;
         private bool useAppPort;
-        const string outputToMatchOnStart = "dapr initialized. Status: Running.";
-        const string outputToMatchOnStop = "app stopped successfully";
+        private readonly string[] outputToMatchOnStart = new string[]{ "You're up and running", };
+        private readonly string[] outputToMatchOnStop = new string[]{ "app stopped successfully", "failed to stop app id", };
```
rynowak, line 25: Need to handle the case where the app didn't properly start.

The existing code was leaking the process if something went wrong during startup. Ideally we'd also keep track of the PID and kill it if its still running when the tests finish.

test/Dapr.E2E.Test/DaprTestApp.cs
```diff
@@ -31,28 +33,63 @@ public DaprTestApp(ITestOutputHelper output, string appId, bool useAppPort = fal
...
+
+                // `dotnet run` args
+                "dotnet", "run",
+                "--project", projectPath,
+                "--framework", GetTargetFrameworkName(),
+
+                // Prevent building again, it slows things down as restore can be slow or unreliable in CI environments
+                "--no-build",
+                "--no-restore",
+            });
```
rynowak, line 75: I find this approach really preferrable for commandline args. it's easier to read and easy to add things like quoting if we need it.

test/Dapr.E2E.Test/DaprTestApp.cs
```diff
@@ -31,28 +33,63 @@ public DaprTestApp(ITestOutputHelper output, string appId, bool useAppPort = fal
...
-            var daprStart = new DaprCommand()
+            // TODO: we don't do any quoting right now because our paths are guaranteed not to contain spaces
+            var daprStart = new DaprCommand(this.testOutput)
             {
                 DaprBinaryName = DaprTestApp.daprBinaryName,
-                Command = daprStartCommand,
+                Command = string.Join(" ", arguments),
                 OutputToMatch = outputToMatchOnStart,
-                Timeout = 10000
+                Timeout = TimeSpan.FromSeconds(30),
```
rynowak, line 88: I found that some of these timeouts weren't long enough. it's best to err on the side of really generous timeouts. You only want to hit a timeout if it's really truly not working.

rynowak: Ex: the Java E2E tests use a 60s timeout

test/Dapr.E2E.Test/E2ETests.cs
```diff
@@ -0,0 +1,97 @@
...
+                var response = await client.GetAsync($"{HttpEndpoint}/v1.0/healthz");
+                if (response.IsSuccessStatusCode)
+                {
+                    return;
+                }
+
+                await Task.Delay(TimeSpan.FromMilliseconds(250));
+            }
+
+            throw new TimeoutException("Timed out waiting for daprd health check");
```
rynowak, line 69: I found that this was needed for testing service invocation. This was unreliable for me locally without the health check.

test/Dapr.E2E.Test/ServiceInvocation/E2ETests.ServiceInvocationTests.cs
```diff
@@ -10,35 +10,24 @@ namespace Dapr.E2E.Test
...
+
+            using var client = DaprClient.CreateInvokeHttpClient(appId: this.AppId, daprEndpoint: this.HttpEndpoint);
             var transaction = new Transaction()
             {
                 Id = "1",
                 Amount = 50
             };
+
             var response = await client.PostAsJsonAsync<Transaction>("/accountDetails", transaction, cts.Token);
-            var account = await response.Content.ReadFromJsonAsync<Account>();
```
rynowak, line 41: This test had a bug when the response was a 500 (runtime not ready).

Validating HTTP requests is one of those areas where you really want a helper method that's going to log everything important on failure. We want all of the failures to lead directly to a fix, so they need to capture a lot of info.

test/Dapr.E2E.Test/DaprTestApp.cs
```diff
@@ -19,8 +21,8 @@ public class DaprTestApp
         static string daprBinaryName = "dapr";
         private string appId;
         private bool useAppPort;
-        const string outputToMatchOnStart = "dapr initialized. Status: Running.";
-        const string outputToMatchOnStop = "app stopped successfully";
+        private readonly string[] outputToMatchOnStart = new string[]{ "You're up and running", };
```
vinayada1, line 24: The issue with this string was that this output appears only when the sidecar starts...the app has not necessarily started and been discovered. The service invocation calls might fail at that point. "dapr initialized. Status: Running." is shown after dapr and the app are fully up

src/Dapr.Actors/Runtime/ActorRuntimeOptions.cs
```diff
@@ -161,5 +161,22 @@ public string DaprApiToken
...
+        /// </summary>
+        /// <remarks>
+        /// The URI endpoint to use for HTTP calls to the Dapr runtime. The default value will be 
+        /// <c>http://127.0.0.1:DAPR_HTTP_PORT</c> where <c>DAPR_HTTP_PORT</c> represents the value of the 
+        /// <c>DAPR_HTTP_PORT</c> environment variable.
+        /// </remarks>
+        /// <value></value>
+        public string HttpEndpoint { get; set; } = GetDefaultHttpEndpoint();
+
+        private static string GetDefaultHttpEndpoint()
```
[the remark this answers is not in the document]
rynowak: Deduped this everywhere

test/Dapr.E2E.Test/DaprTestApp.cs
rynowak: OK cool. Will revert this. Ultimately I think it's the health check that made a difference.

rynowak:
@vinayada1 updated

vinayada1 (approved):

Issue #627, opened 2021-03-11
Dapr Configuration Normalization for Azure Key Vault


SamuelMcAravey:
## Describe the proposal
Currently, the `Dapr.Extensions.Configuration` NuGet package includes functionality to normalize a double underscore '__' into a colon to allow for hierarchical configuration. When using the Azure Key Vault secret store we run into the limitation described here:

> Azure Key Vault secret names are limited to alphanumeric characters and dashes. Hierarchical values (configuration sections) use -- (two dashes) as a separator. Colons, which are normally used to delimit a section from a subkey in ASP.NET Core configuration, aren't allowed in key vault secret names. Therefore, two dashes are used and swapped for a colon when the secrets are loaded into the app's configuration.

https://docs.microsoft.com/en-us/aspnet/core/security/key-vault-configuration?view=aspnetcore-5.0#secret-storage-in-the-production-environment-with-azure-key-vault

Since the Azure Key Vault secret store is officially supported by Dapr I believe it would be a good idea to either:
1) Add normalization specifically for Key Vault (easy)
2) Allow for customizable normalization or modification of the configuration keys via code (hard)

https://github.com/dapr/dotnet-sdk/blob/7890bc7977e68af8bab85066d2e8213497cda1bd/src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationProvider.cs#L72-L75

The first option could be changed like this:
```csharp
private static string NormalizeKey(string key)
{
    return key.Replace("__", ConfigurationPath.KeyDelimiter).Replace("--", ConfigurationPath.KeyDelimiter);
}
```

Is this something feasible?

rynowak:
We already added this I think: https://github.com/dapr/dotnet-sdk/issues/615 - it will be part of an upcoming 1.1 release of the SDK.

Does this address what you're looking for?

rynowak:
I think we could make the delimiter configurable - this is what other config providers for `Microsoft.Extensions.Configuration` do. That way the core logic would live inside the SDK, and users would just be responsible for providing delimiters.

We're not going to add support for additional delimiters as built-in defaults. It's always a breaking change to add one, and there are already two built-in options.. `:` and `__`.

rynowak:
I'm open to suggestions about how to make this easy, but my reservation is that we can't add all possible delimiters for all possible configuration stores as defaults. That becomes to hard to work with in the other direction.

SamuelMcAravey:
[...]

PR #629, opened 2021-03-11
[WIP] Added support for custom delimiters in secret keys.

SamuelMcAravey:
# Description

This PR proposes the addition of a `CustomDelimiters` property. The custom delimiters can be used in two ways:

```csharp
.AddDaprSecretStore("store", daprClient, new[] { "--", "≡" }) // Example of multiple delimiters
```

or

```csharp
.AddDaprSecretStore((conf) =>
{
    conf.Store = "store";
    conf.Client = daprClient;
    conf.CustomDelimiters = new[] { "--", "≡" }; // Example of multiple delimiters
})
```

This preserves the double underscore replacement regardless of the developer defining any custom delimiters or not. Documentation will be added after getting some initial opinion on this.

## Issue reference

This PR closes issue: #627 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationExtensions.cs
+++ b/src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationExtensions.cs
@@ -83,7 +83,7 @@ namespace Dapr.Extensions.Configuration
             this IConfigurationBuilder configurationBuilder,
             string store,
             DaprClient client,
-            IList<string> keyDelimiters)
+            IEnumerable<string> keyDelimiters)
         {
             ArgumentVerifier.ThrowIfNullOrEmpty(store, nameof(store));
             ArgumentVerifier.ThrowIfNull(client, nameof(client));
--- a/src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationExtensions.cs
+++ b/src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationExtensions.cs
@@ -83,7 +83,7 @@ namespace Dapr.Extensions.Configuration
             this IConfigurationBuilder configurationBuilder,
             string store,
             DaprClient client,
-            IEnumerable<string> keyDelimiters)
+            IList<string> keyDelimiters)
         {
             ArgumentVerifier.ThrowIfNullOrEmpty(store, nameof(store));
             ArgumentVerifier.ThrowIfNull(client, nameof(client));
--- a/src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationExtensions.cs
+++ b/src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationExtensions.cs
@@ -76,14 +76,14 @@ namespace Dapr.Extensions.Configuration
         /// </summary>
         /// <param name="configurationBuilder">The <see cref="IConfigurationBuilder"/> to add to.</param>
         /// <param name="store">Dapr secret store name.</param>
-        /// <param name="customDelimiters">A collection of delimiters that will be replaced by ':' in the key of every secret.</param>
+        /// <param name="keyDelimiters">A collection of delimiters that will be replaced by ':' in the key of every secret.</param>
         /// <param name="client">The Dapr client</param>
         /// <returns>The <see cref="IConfigurationBuilder"/>.</returns>
         public static IConfigurationBuilder AddDaprSecretStore(
             this IConfigurationBuilder configurationBuilder,
             string store,
             DaprClient client,
-            IEnumerable<string> customDelimiters)
+            IEnumerable<string> keyDelimiters)
         {
             ArgumentVerifier.ThrowIfNullOrEmpty(store, nameof(store));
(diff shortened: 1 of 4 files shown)


SamuelMcAravey:
There are two overloads for the DaprSecretStoreConfigurationProvider constructor that no longer being used, but they were left in since they are marked as `public`. I assume that is why code coverage is down.

src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationProvider.cs
```diff
@@ -69,9 +97,39 @@ public DaprSecretStoreConfigurationProvider(string store, bool normalizeKey, IRe
...
+            key = key.Replace("__", ConfigurationPath.KeyDelimiter);
+
+            if (this.customDelimiters?.Any() == true)
+            {
+                foreach (var keyDelimiter in this.customDelimiters)
+                {
+                    key = key.Replace(keyDelimiter, ConfigurationPath.KeyDelimiter);
+                }
+            }
+
```
rynowak, line 131: I think this is mostly right, I have a few suggestions after seeing this code... It might be better if we could make `__` removable rather than something that's **truly** built in. 

Here's what I'd suggest:

Make the property on `DaprSecretStoreConfigurationSource` look like:

```
public IList<string> KeyDelimiters { get; set; } = new List<string>(){ "--", };
```

Turn your parameter into `IEnumerable<string> `keyDelimiters` and in your new overload, you replace `source.KeyDelimiters`. This way we don't actually special case `--`, and it's something users can remove if they want. This seems a little more composable.

Your thoughts?

SamuelMcAravey:
I agree with you. Making the double underscore optional was in the initial code, but keeping it "built-in" was the way I chose in order to keep the functionality as close as possible to the current behavior.

I did want to clarify the updated property definition on the `DaprSecretStoreConfigurationSource ` class. On the one hand, an `IList<>` is convenient for the developer to add to the existing values, and on the other hand, the `SecretDescriptors` and `Metadata` properties are read-only. For consistency, I'd vote to keep it as `IEnumerable`. For convenience, I would vote for `ICollection` or `IList`. Thoughts?

I'll get another commit with the changes you've suggested so far. Thanks!

src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationSource.cs
```diff
@@ -21,15 +21,15 @@ public class DaprSecretStoreConfigurationSource : IConfigurationSource
...
         /// Default value true.
         /// </summary>
         public bool NormalizeKey { get; set; } = true;
 
         /// <summary>
-        /// Gets or sets the custom key delimiters.
+        /// Gets or sets the custom key delimiters. Contains the '__' delimiter by default.
         /// </summary>
-        public IEnumerable<string> CustomDelimiters { get; set; }
+        public IEnumerable<string> KeyDelimiters { get; set; } = new List<string> { "__" };
```
rynowak, line 32: Can you make this `IList<>` it's better if it can be updated in place.

src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationProvider.cs
```diff
@@ -101,29 +101,27 @@ public DaprSecretStoreConfigurationProvider(string store, bool normalizeKey, IRe
...
             this.metadata = metadata;
             this.client = client;
         }
 
         private string NormalizeKey(string key)
         {
-            key = key.Replace("__", ConfigurationPath.KeyDelimiter);
-
-            if (this.customDelimiters?.Any() == true)
+            if (this.keyDelimiters?.Any() == true)
```
rynowak, line 122: If you can change this to `IList` then you can get rid of `.Any()` and and use the `.Count` property

SamuelMcAravey:
That last commit should have resolved your last two asks. Anything else that catches your eye?

src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationExtensions.cs
```diff
@@ -71,6 +71,33 @@ public static class DaprSecretStoreConfigurationExtensions
...
+        /// <param name="configurationBuilder">The <see cref="IConfigurationBuilder"/> to add to.</param>
+        /// <param name="store">Dapr secret store name.</param>
+        /// <param name="keyDelimiters">A collection of delimiters that will be replaced by ':' in the key of every secret.</param>
+        /// <param name="client">The Dapr client</param>
+        /// <returns>The <see cref="IConfigurationBuilder"/>.</returns>
+        public static IConfigurationBuilder AddDaprSecretStore(
+            this IConfigurationBuilder configurationBuilder,
+            string store,
+            DaprClient client,
+            IList<string> keyDelimiters)
```
rynowak, line 86: Can you keep this as `IEnumerable<>` and just convert to a list on L94?

rynowak (commented):
Just one minor comment to address

rynowak (approved):
Looks great! Thanks for helping with this.

Issue #523, opened 2021-01-02
Add nullability annotations to DaprClient


rynowak:
## Describe the proposal

Once we ingest the .NET 5.0 SDK we we'll be able to add nullability annotations to most of the methods on `DaprClient`. The reason we didn't do this because is because prior to the C#9 compiler, a core use case didn't work with nullability:

```C#
public Task<TValue> DoSomething<TValue>();
```

There is no supported way (pre C# 9) to say *this method returns a task that might return a null* without also constraining `TValue : class`. We didn't want to block the use of value types at a fundamental level especially when records and structs are about to hit.

Now that this barrier has been removed there will be a benefit to adding NNRT annotations.

dlemstra:
I gave this a try and changed almost everything in the `Dapr.Client` project. But `DaprClientGrpc` fails to compile. Details are in the image below. Not sure if this is a bug or that I should implement it differently because the method is abstract? Any ideas @rynowak?

![image](https://user-images.githubusercontent.com/10426229/111083373-92450d00-850d-11eb-924b-1611ecb82610.png)
![afbeelding](https://user-images.githubusercontent.com/10426229/111083412-c1f41500-850d-11eb-8177-ffb2c6486ad8.png)

Small reproducable test case that fails to compile:
```C
using System;
using System.Collections.Generic;
using System.Text;
using System.Threading.Tasks;

#nullable enable

namespace Dapr.Client
{
    abstract class Class1
    {
        public abstract Task<T?> Get<T>();
    }
    class Class2 : Class1
    {
        public override Task<T?> Get<T>()
        {
            throw new NotImplementedException();
        }
    }
}
```

rynowak:
Thanks @dlemstra - I would have expected this to work as well.  I was able to track down the documentation for this. https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/unconstrained-type-parameter-annotations

I wonder if we need to use `where T : default`

If you are interested in taking this on, it might be best to do it in chunks and review each stage separately. Maybe start with something simple like bindings. I'd hate for you to take the trouble of changing everything before we agree on design.

rynowak:
Note for myself in the future - when we do start rolling this out, we'll want make an announcement about how we're positioning it:

- We expect this to be a *source* breaking change for users with nullabilty enabled
- We expect that we will get some things wrong and will change over time
[...]

PR #636, opened 2021-03-20
Enabled nullable reference types in Dapr.Extensions.Configuration.

dlemstra:
# Description

This PR enables nullable reference types in the Dapr.Extensions.Configuration library. 

It also includes a change to create a copy of the `ArgumentVerifier` class so that each library has it's own version of this file. That is debatable but this was the only library that referenced if from another. Another idea would be to move the `ArgumentVerifier,cs` file to `properties` folder change the namespace to just `Dapr` and reference it in all the projects. I don't mind including that in this pull request but I can also open a new one once this has been merged.

This also uses `!` when constructing the class `DaprSecretStoreConfigurationProvider`. This was done because it did not make sense to me to allow null in the constructor of that class. Was wondering if an exception should be raised in the `Build` method instead when the values are null. Maybe it would be even better to add a constructor to `DaprSecretStoreConfigurationSource` that has a `store` and `client` argument and raise an exception when either of these are null?

## Issue reference

Implements part of #523.

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationSource.cs
```diff
@@ -29,22 +29,22 @@ public class DaprSecretStoreConfigurationSource : IConfigurationSource
...
         /// A collection of metadata key-value pairs that will be provided to the secret store. The valid metadata keys and values are determined by the type of secret store used.
         /// </summary>
-        public IReadOnlyDictionary<string, string> Metadata { get; set; }
+        public IReadOnlyDictionary<string, string>? Metadata { get; set; }
 
         /// <summary>
         /// Gets or sets the http client.
         /// </summary>
-        public DaprClient Client { get; set; }
+        public DaprClient? Client { get; set; }
```
rynowak, line 47: Most of the properties in this class fall into the category of *should not be null in common usage*. For instance users should not encounter a case where `Client` is null, nor should they set it to null. 

I'd suggest using `{ get; set; } = default!;` For this one.

src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationSource.cs
```diff
@@ -29,22 +29,22 @@ public class DaprSecretStoreConfigurationSource : IConfigurationSource
...
         /// Gets or sets the custom key delimiters. Contains the '__' delimiter by default.
         /// </summary>
-        public IList<string> KeyDelimiters { get; set; } = new List<string> { "__" };
+        public IList<string>? KeyDelimiters { get; set; } = new List<string> { "__" };
 
         /// <summary>
         /// Gets or sets the secret descriptors.
         /// </summary>
-        public IEnumerable<DaprSecretDescriptor> SecretDescriptors { get; set; }
+        public IEnumerable<DaprSecretDescriptor>? SecretDescriptors { get; set; }
```
rynowak, line 37: Does it work to set this to `{ get; set; } = Array.Empty<DaprSecretDescriptors>()`

src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationSource.cs
```diff
@@ -29,22 +29,22 @@ public class DaprSecretStoreConfigurationSource : IConfigurationSource
         /// <summary>
         /// Gets or sets the custom key delimiters. Contains the '__' delimiter by default.
         /// </summary>
-        public IList<string> KeyDelimiters { get; set; } = new List<string> { "__" };
+        public IList<string>? KeyDelimiters { get; set; } = new List<string> { "__" };
```
rynowak, line 32: Would prefer that this not be made nullable

src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationSource.cs
```diff
@@ -18,7 +18,7 @@ public class DaprSecretStoreConfigurationSource : IConfigurationSource
         /// <summary>
         /// Gets or sets the store name.
         /// </summary>
-        public string Store { get; set; }
+        public string? Store { get; set; }
```
rynowak, line 21: Would prefer that this not be made nullable

src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationSource.cs
rynowak: Looks like I have the answer - this should be nullable after all. https://github.com/dapr/dotnet-sdk/pull/636/files#diff-e03608cfa6c16ee7656cce2f6bbe55da1dea01c3994fd6822a7adb20fa44d8cfL139

Unrelated to your PR we made a mistake here... it's not good to have a semantic meaning between *null collection* and *empty collection*.

src/Dapr.Extensions.Configuration/Dapr.Extensions.Configuration.csproj
```diff
@@ -13,10 +14,6 @@
     <ProjectReference Include="..\Dapr.Client\Dapr.Client.csproj" />
   </ItemGroup>
 
-  <ItemGroup>
-    <Compile Include="..\Dapr.Client\ArgumentVerifier.cs" />
-  </ItemGroup>
```
rynowak, line 18: Instead of copying this into the project can you create `..\Shared\ArgumentVerifierNullable.cs`? Then we can limit the amount of duplication to N == 2.

src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationSource.cs
dlemstra: Do you also want me to to change `DaprSecretStoreConfigurationProvider` to not allow a null value for this? Or only make it non nullable in this path?

src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationSource.cs
dlemstra: Used an empty string as the default value here. This will raise an `ArgumentException` in the constructor of `DaprSecretStoreConfigurationProvider`.

src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationSource.cs
```diff
@@ -18,7 +18,7 @@ public class DaprSecretStoreConfigurationSource : IConfigurationSource
         /// <summary>
         /// Gets or sets the store name.
         /// </summary>
-        public string Store { get; set; }
+        public string Store { get; set; } = string.Empty;
```
rynowak, line 21: This should also be initialized with `default!`

rynowak (changes requested):
Sorry for the long delay on this. One more small suggestion and then this is ready to go.

src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationSource.cs
dlemstra: Changed the default value but I wonder if an empty string would not be better than making it "secretly" null.

rynowak: Part of it is personal preference (based on experience). Using `null` as a sentinel value is a far better practice than `string.Empty`. You actually **want** to get an exception when you access an uninitialized value.

The challenge here is that it's hard to retrofit nullability onto existing APIs. The design of this class is not a good fit, but unfortunately can't be easily changed.

rynowak (dismissed):

dlemstra:
Had to do another push @rynowak to update this branch.

rynowak (approved):

rynowak:
Thanks @dlemstra

Issue #655, opened 2021-04-14
DAPR_API_TOKEN environment variable is not read correctly in 1.1.0


saber-wang:
https://github.com/dapr/dotnet-sdk/blob/5e5a42c96e5b6de1374d50432ae3b8442b908964/src/Shared/DaprDefaults.cs#L26

a-elsheikh:
we've also just bumped into this one - fyi @rynowak

rynowak:
Yes, we'll fix this and release a patch.

rynowak:
Can you provide some more info here for testing? What's the actual problem you encountered (just looking for confirmation).

saber-wang:
@rynowak  I just looked at the source code and found this problem

a-elsheikh:
my specific scenario was initializing an ActorProxyFactory and using it to create an actor proxy, calls to the actor began throwing an AuthenticationException with message "Invalid client credentials". Stepping through the code led me to the line mentioned above

rynowak:
Thanks @saber-wang and @a-elsheikh - I am able to confirm this is a bug. I also found an issue with propagating API tokens for the `CreateHttpMethodRequest` family of methods.

rynowak:
We will fix this and release a `1.1.1` patch

PR #658, opened 2021-04-16
Fix handling of DAPR_API_TOKEN

rynowak:
# Description

This fixes a functional bug in how we initialize the default settings of
based on the `DAPI_API_TOKEN` environment variable.

Additional, the `CreateInvokeMethodRequest` family of methods were
missing API token support. This has been added.

## Issue reference

Fixes: #655

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


src/Dapr.Client/DaprClient.cs
```diff
@@ -34,7 +34,7 @@ public abstract class DaprClient : IDisposable
         /// Gets the <see cref="JsonSerializerOptions" /> used for JSON serialization operations.
         /// </summary>
         public abstract JsonSerializerOptions JsonSerializerOptions { get; }
-        
+
```
rynowak, line 37: just autoformat doing its thing

src/Dapr.Client/DaprClient.cs
```diff
@@ -92,14 +92,14 @@ public static HttpClient CreateInvokeHttpClient(string appId = null, string dapr
             return httpClient;
         }
 
-        internal static KeyValuePair<string, string> GetDaprApiTokenHeader(string apiToken)
+        internal static KeyValuePair<string, string>? GetDaprApiTokenHeader(string apiToken)
```
rynowak, line 95: this is clearer to model it as a nullable since the api token is optional.

Internal api so there's no breaking change impact

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -263,6 +263,11 @@ public override HttpRequestMessage CreateInvokeMethodRequest(HttpMethod httpMeth
                 }
             };
 
+            if (this.apiTokenHeader is not null)
+            {
+                request.Headers.TryAddWithoutValidation(this.apiTokenHeader.Value.Key, this.apiTokenHeader.Value.Value);
+            }
```
rynowak, line 269: This was a bug that no one had reported yet, but worth patching. Found this while reproducing the bug.

vinayada1 (approved):

Issue #640, opened 2021-03-24
Add method for health check in dapr client


dmytrohridin:
## Proposal
https://docs.dapr.io/reference/api/health_api/
It would be useful to have build-in method (for example `Ping`) in `DaprClient` that checks health of sidecar

rynowak:
Yes, this is something we should add. Happy to accept a PR from anyone who wants to help with this.

dmytrohridin:
will try

artursouza:
This issue is related to https://github.com/dapr/dotnet-sdk/issues/452

PR #662, opened 2021-04-21
Implement dapr sidecar healthcheck method

dmytrohridin:
# Description

Implement a method that performs health check of Dapr sidecar

## Issue reference

Please reference the issue this PR will close: #640  

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation (summary)


src/Dapr.Client/DaprClient.cs
```diff
@@ -233,6 +233,12 @@ public HttpRequestMessage CreateInvokeMethodRequest(string appId, string methodN
         /// <returns>An <see cref="HttpRequestMessage" /> for use with <c>SendInvokeMethodRequestAsync</c>.</returns>
         public abstract HttpRequestMessage CreateInvokeMethodRequest(HttpMethod httpMethod, string appId, string methodName);
 
+        /// <summary>
+        /// Creates an <see cref="HttpRequestMessage" /> that can be used to perform health-check of Dapr sidecar
+        /// </summary>
+        /// <returns>An <see cref="HttpRequestMessage" /> for use with <c>InvokeHealthMethodAsync</c>.</returns>
+        public abstract HttpRequestMessage CreateInvokeHealthMethodRequest();
```
rynowak, line 240: What's the use case of this separate from `InvokeHealthMethodAsync`? If there isn't a concrete scenario where this is necessary, just leave it out - it could be added latter if needed.

dmytrohridin: currently no scenario, just try to be consistent with request methods. Will simplify

rynowak (changes requested):
Sorry, it turns out I had a bunch of pending comments that never got sent through. I meant to leave all of these at the same time.

dmytrohridin:
@rynowak thank you for the comments, all make sense. Will fix them this days

rynowak (approved):
Awesome. Thanks for doing this. We're in a hold right now to ship 1.2 so I merge this as soon as we're open again.

rynowak:
thanks @dmytrohridin

Issue #677, opened 2021-05-13
Add support to provide Actor TypeName during runtime at startup.


bjorkstromm:
## Describe the proposal

Currently, since Actor Placement Service is global within the Kubernetes cluster, it makes it difficult to run multiple versions (e.g. DEV and TEST) of the application within the same cluster. This is due to the fact that Actor names are either automatically assigned from the name of the implementing class or by using the `ActorAttribute` and specifying the `TypeName` property. Bot of these approaches are compile-time constants which can't be adjusted during runtime.

Here's two proposals:

1. Make it possible to resolve Actor TypeName from an environment variable when specified in the `ActorAttribute`. This is e.g. possible in Azure Functions with most of the attributes there. This would allow us to use the attribute as follows:
   ```csharp
   [Actor(TypeName = "%MyActorName%")]
   public class MyActorName : Dapr.Actors.Runtime.Actor, IMyActor
   {
   }
   ```
   The `%MyActorName%` can then be resolved from an environment variable using e.g. [Environment.ExpandEnvironmentVariables](https://docs.microsoft.com/en-us/dotnet/api/system.environment.expandenvironmentvariables?view=net-5.0), or using a more sophisticated approach as Azure Functions SDK (see [NameResolverExtensions.cs](https://github.com/Azure/azure-webjobs-sdk/blob/892f6278f449c468391ce76aa086cf78aa065a5f/src/Microsoft.Azure.WebJobs.Host/NameResolverExtensions.cs#L39)).

2. Make it possible to provide ActorTypeName when calling `RegisterActor<T>()` on the `ActorRegistrationCollection`. This would currently be possible if `ActorTypeName` on `ActorTypeInformation` wasn't readonly. (I do it today using reflection :)). Propsal here is to add a overload to `RegisterActor<T>()` which takes a `string` parameter for the Actor TypeName.

rynowak:
I'd prefer approach 2 - the functions approach would be surprising, given that nothing else we do works that way. 

Feel free to send a PR for option 2 :)

bjorkstromm:
@rynowak consider it done!

Regarding Approach 1, I actually wanted this also for the `TopicAttribute`. See https://github.com/dapr/dotnet-sdk/issues/678. Don't really know what the best approach would be there (beside the Azure Function approach). But, but... Let's continue that discussion on that issue.

devlead:
Must say I really like option one, would make it really straightforward to have the same set of "artifacts" flow and be reused through environments.

PR #680, opened 2021-05-16
Adds support for providing Actor type name during runtime.

bjorkstromm:
# Description

Adds support for providing Actor type name during runtime.

## Issue reference

Please reference the issue this PR will close: #677

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/src/Dapr.Actors/Runtime/ActorRegistrationCollection.cs
+++ b/src/Dapr.Actors/Runtime/ActorRegistrationCollection.cs
@@ -1,4 +1,4 @@
-// ------------------------------------------------------------
+﻿// ------------------------------------------------------------
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 // ------------------------------------------------------------
@@ -32,6 +32,22 @@ namespace Dapr.Actors.Runtime
             where TActor : Actor
         {
             var actorTypeInfo = ActorTypeInformation.Get(typeof(TActor));
+            var registration = new ActorRegistration(actorTypeInfo);
+            configure?.Invoke(registration);
+            this.Add(registration);
+        }
+
+        /// <summary>
+        /// Registers an actor type in the collection.
+        /// </summary>
+        /// <typeparam name="TActor">Type of actor.</typeparam>
+        /// <param name="actorTypeName">The name of the actor type represented by the actor.</param>
+        /// <param name="configure">An optional delegate used to configure the actor registration.</param>
+        /// <remarks>The value of <paramref name="actorTypeName"/> will have precedence over the default actor type name derived from the actor implementation type or any type name set via <see cref="ActorAttribute"/>.</remarks>
+        public void RegisterActor<TActor>(string actorTypeName, Action<ActorRegistration> configure = null)
+            where TActor : Actor
+        {
+            var actorTypeInfo = ActorTypeInformation.Get(typeof(TActor), actorTypeName);
             var registration = new ActorRegistration(actorTypeInfo);
             configure?.Invoke(registration);
             this.Add(registration);
--- a/src/Dapr.Actors/Runtime/ActorTypeInformation.cs
+++ b/src/Dapr.Actors/Runtime/ActorTypeInformation.cs
@@ -1,4 +1,4 @@
-// ------------------------------------------------------------
+﻿// ------------------------------------------------------------
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 // ------------------------------------------------------------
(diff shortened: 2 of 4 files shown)


src/Dapr.Actors/Runtime/ActorRegistrationCollection.cs
```diff
@@ -36,5 +36,21 @@ public void RegisterActor<TActor>(Action<ActorRegistration> configure = null)
...
+        /// Registers an actor type in the collection.
+        /// </summary>
+        /// <typeparam name="TActor">Type of actor.</typeparam>
+        /// <param name="actorTypeName">The name of the actor type represented by the actor.</param>
+        /// <param name="configure">An optional delegate used to configure the actor registration.</param>
+        /// <remarks>The value of <paramref name="actorTypeName"/> will have precedence over the default actor type name derived from the actor implementation type or any type name set via <see cref="ActorAttribute"/>.</remarks>
+        public void RegisterActor<TActor>(string actorTypeName, Action<ActorRegistration> configure = null)
+            where TActor : Actor
+        {
+            var actorTypeInfo = ActorTypeInformation.Get(typeof(TActor), actorTypeName);
```
rynowak, line 50: It's going to take me a bit to review this and rationalize the change. I have to check whether there are other places where we call `ActorTypeInformation.Get`.

bjorkstromm: Ok. To clarify, this should be a non-breaking change. I added an overload for `ActorTypeInformation.Get` which takes an string parameter for specifying the actor type name.

```csharp
public static ActorTypeInformation Get(Type actorType)
``` 

The above still exists, but will call the below with `actorTypeName: null`.
```csharp
public static ActorTypeInformation Get(Type actorType, string actorTypeName)
```

And then in the above method I've just modified it to check is `actorTypeName` is provided.

src/Dapr.Actors/Runtime/ActorTypeInformation.cs
```diff
@@ -121,7 +138,7 @@ public static ActorTypeInformation Get(Type actorType)
 
             var actorAttribute = actorType.GetCustomAttribute<ActorAttribute>();
 
-            string actorTypeName = actorAttribute?.TypeName ?? actorType.Name;
+            actorTypeName ??= actorAttribute?.TypeName ?? actorType.Name;
```
amanbha, line 141: I would prefer getting it from attribute and follow the approach which is being done for pubsub here https://github.com/dapr/dotnet-sdk/pull/681

bjorkstromm: Sounds ok, that approach would however lock configuring actor type name to statics.

amanbha: adding @rynowak as well as he preferred the option1 which is implemented int he PR whereas I am suggesting option 2

rynowak: I think switching this to use the same technique as #681 would be fine as well. I realize now that we were talking about similar features across two different areas and I gave you different answers.

bjorkstromm: The main difference between #681 and this is that in #681 it's working on endpoint metadata, which doesn't need to be an attribute. In this case we don't have a dependency on ASP.NET Core, thus this can't easily be built using a similar technique.

I can create a new PR, which creates a new interface `IActorMetadata`, let `ActorAttribute` implement that and then replace all references to `ActorAttribute` with `IActorMetadata`. But that will lock this implementation to only be able to use statics for overriding Actor type name. Statics should be fine for this scenario, but if one would e.g. like to use `IConfiguration` for specifying Actor type name there's a little extra (relatively ugly) work that needs to be done (e.g. create a static class and set some values prior to calling `services.AddActors(...)`).

I'll let you decide 😄

amanbha: @bjorkstromm in that case current approach seems fine, thanks for clarifying.

src/Dapr.Actors/Runtime/ActorRegistrationCollection.cs
rynowak: > Ok. To clarify, this should be a non-breaking change. I added an overload for ActorTypeInformation.Get which takes an string parameter for specifying the actor type name.

What I'm concerned about isn't the addition of a new method it's a the introduction of a 'split brain' problem. We need to make sure it's not possible for two parts of the system to disagree about the type name.

rynowak: Can you please do the following:

- On ActorTypeInformation - mark `TryGet(Type)` and `Get(Type)` as `[Obsolete]`- we want to make sure we don't have code calling those since they don't accept a type name override.
- On ActorHost we need another overload of `CreateForTest` that accepts a type name.
- Update the remaining references to `Get(Type)` to call `GetType(Type, null)` - these will be unit tests

bjorkstromm:
@rynowak it seems like this was not closed by intention... Can you reopen, and I'll address feedback mentioned here: https://github.com/dapr/dotnet-sdk/pull/680#discussion_r646191590

src/Dapr.Actors/Runtime/ActorRegistrationCollection.cs
bjorkstromm: Fixed this here https://github.com/bjorkstromm/dotnet-sdk/commit/e1b62c9ddcde129c269a77bd477bcaac717f0820

But since this PR was unintentionally closed, changes won't appear here. Could you please reopen, or should I open a new PR?

PR #684, opened 2021-05-27
Improve Actor descriptions

eNeRGy164:
I wanted to improve the code coverage for the SDK.
Picked a place to start where there was no coverage yet: `Dapr.Actors.Descriptions`.

While adding tests, I found some interesting things (and small typo's).
I added every change as a separate commit for convenience so they can be inspected individually.

# Description

1. c73dbb922d803780fd98f843fe350678b9a18072 If other flags were set (for example `HasThis`) it would fail to detect the `VarArgs`. Also fixed a missing 'r' in resource identifier.
2. 3f657c6042052fe557ffa83347cc3e2e87bec141 Fixed a missing 'r' in resource identifier.
3. ce159b840473b6a2f9afced451b50983d3f17d72 Fixed a double space in an error message
4. c6732d51155784aecb8e0500c6eea53350dcf97e This sentence did not reflect the actual error.
   For example:
    ```csharp
    internal interface ITestActor : IActor
    {
        Task Method(CancellationToken cancellationToken, bool additionalArgument);
    }
    ```
    resulted in the exception:
    `Method 'Method' of actor interface '...ITestActor' has '...CancellationToken' parameter 'additionalArgument', and it is not the last parameter. If a method of the actor interface has parameter of type '...CancellationToken' it must be the last parameter.`

   but `additionalArgument` is not of type `CancellationToken`.

   Changed to:
    `Method 'Method' of actor interface '...ITestActor' has has a '...CancellationToken' parameter that is not the last parameter. If an actor method accepts a '...CancellationToken' parameter, it must be the last parameter.`
5. 6b83e43d97a2be914f7b174be8f6f9bcafd42fc9 The incorrect error resource identifier was choosen, resulting in an incorrect error with misalligned variables:
    For example:
    ```csharp
    internal interface ITestActor : ICloneable
    {
    }
    ```
    resulted in the exception:
    `The type '...ITestActor' is not an actor interface as it does not derive from the interface '...ICloneable'.`

    Changed to:

    `The type '...ITestActor'is not an actor interface as it derive from a non actor interface '...ICloneable'. All actor interfaces must derive from '...IActor'.`
6. d3504cbc521bd84b0055d762df70762b2ababf25 All unit test that describe and cover all code in the `Dapr.Actors.Descriptions` namespace.

## Issue reference

Contributes to #437

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


src/Dapr.Actors/Resources/SR.resx
```diff
@@ -179,7 +179,7 @@
     <value>The {0} interface '{1}' is using generics. Generic interfaces cannot be remoted.</value>
   </data>
   <data name="ErrorRemotedMethodCancellationTokenOutOfOrder" xml:space="preserve">
-    <value>Method '{1}' of {0} interface '{2}' has '{4}' parameter '{3}', and it is not the last parameter. If a method of the {0} interface has parameter of type '{4}' it must be the last parameter.</value>
+    <value>Method '{1}' of {0} interface '{2}' has parameter '{3}', while another argument was already of type '{4}'. If a method of the {0} interface has parameter of type '{4}', it must be the last parameter.</value>
```
rynowak, line 182: I don't think this is clearer - especially the `while another argument was already of type...` part.

Based on your example:

For example:

```C#
internal interface ITestActor : IActor
{
    Task Method(CancellationToken cancellationToken, bool additionalArgument);
}
```

I'd suggest the following:

> Method 'Method' of actor interface '...ITestActor' has a '...CancellationToken' parameter that is not the last parameter. If an actor method accepts a  '...CancellationToken' parameter then it must be the last parameter.

src/Dapr.Actors/Description/InterfaceDescription.cs
```diff
@@ -90,7 +91,7 @@ public MethodDescription[] Methods
             bool useCRCIdGeneration)
         {
             EnsureValidMethods(remotedInterfaceKindName, remotedInterfaceType, methodReturnCheck);
-            var methods = remotedInterfaceType.GetMethods();
+            var methods = remotedInterfaceType.GetInterfaceMethods();
```
rynowak, line 94: @eNeRGy164 - can you please separate this commit from this PR and open an issue to discuss. Any behavior changes need to be really well understood. 

I suspect that this is a breaking change we would not accept until the next major release as it *tightens* validation we're performing. We need to talk through all of the possible cases that would be impacted.

rynowak: For example, you said:

> ChildActor has now the description of 3 methods instead of 1 and IOverloadedActor will no longer be seen as valid as GetString is overloaded.

If I understand you correctly, this means that something that is working today will stop working. Even if it's a mistake thats not something we'd just change - these kinds of breaks are very hard for people to reason about when it happens to them.

rynowak:
Thanks for adding the new tests. This is great!

src/Dapr.Actors/Resources/SR.resx
eNeRGy164: Updated

src/Dapr.Actors/Description/InterfaceDescription.cs
eNeRGy164: I will create a separate issue for this.

eNeRGy164: Issue #694 created

eNeRGy164:
@rynowak Any chance to review this after the changes I made 12 days ago?

rynowak (approved):

Issue #685, opened 2021-05-29
Solution does not support Live Unit Testing in Visual Studio Enterprise


eNeRGy164:
This is only for developer comfort, not an issue using the SDK itself.

## Expected Behavior

When enabling Live Unit Testing in Visual Studio Enterprise, it should work.

## Actual Behavior

Fails with the message

> [15:40:32.665 Error] System.IO.FileNotFoundException: Could not find file 'D:\github\eNeRGy164\dapr\dotnet-sdk\.vs\all.sln\v16\lut\7304\i\Dapr.Client\netcoreapp3.1\dapr\proto\common\v1\CommonGrpc.cs'.

A normal build generates an empty `CommonGrpc.cs`, but a LUT build apparently does not.
I'm not sure if this is a Visual Studio LUT or a Grpc.Tools bug. Or maybe just the combination of both.

A workaround is to generate the empty file if it is a LUT build.

After this LUT works:
![image](https://user-images.githubusercontent.com/10671831/120072625-0f0b7000-c095-11eb-9291-26011553a542.png)

## Steps to Reproduce the Problem

Press _play_ in a Visual Studio Enterprise.

## Release Note

RELEASE NOTE:
**FIX** Support Live Unit Testing in Visual Studio Enterprise.

rynowak:
@jamesnk - are you familiar with this problem?

JamesNK:
No. I don’t have live unit tests enabled

rynowak:
Can you try it on an empty grpc project? I don't have Windows/VS easily available. If there's a bug here its in Grpc.Tools - we're not doing anything fancy in this repo.

eNeRGy164:
I couldn't reproduce it on an empty solution, even using the Dapr proto files, including shared files and targets.

PR #686, opened 2021-05-29
Create empty Grpc files on LUT build

eNeRGy164:
# Description

Create a workaround that generates an empty file if it is a Live Unit Testing build.

## Issue reference

Will close: #685

## Checklist

* [x] Code compiles correctly


rynowak:
hi @eNeRGy164 thanks for digging into this. I've pinged the .NET gRPC maintainer to see if I can understand this better. Please sit tight a bit while I do research.

halspang:
@eNeRGy164 - This fell off of our radar, but is it still something you'd like to see? I can see if any progress was made originally/if I can follow-up myself.

WhitWaldo:
@eNeRGy164 Thank you for your contribution!

I'm going to close this PR as we haven't heard anything from you in the last year and with v1.15, the shape of the solution is changing a bit. The various Dapr.* projects do not implement from the protos directly and instead reference them from `Dapr.Protos`, so creating the empty prototype here for testing no longer feels quite relevant. That said, if a the scenario comes up in which we wind up needing to add this to the `Dapr.Protos` file, I'll make sure you assign you for attribution on the PR.

PR #688, opened 2021-05-31
Adds Dapr authentication handler

### LINKED ISSUE #682: Create functionality which eases securing application using Dapr-Api-Token
## Describe the proposal

When reserching guidance for securing PUBSUB, https://github.com/dapr/dapr/issues/3178 I found out that we can instruct Dapr to send a `Dapr-Api-Token` header with each request when communication with the app.

I noticed there are no convenience methods in the .NET SDK for e.g. securing pubsub endpoints using this. What I did instead was to create a simple `AuthenticationHandler` which reads the `Dapr-Api-Token` header and matches it against a pre-configured value (in my case, the `DAPR_API_TOKEN` environment variable).

I'd suggest we'd add something to the .NET SDK for easily securing endpoints that Dapr will communicate with. I'm far from a ASP.NET Core expert, so if @rynowak could chime in with any suggestions I'm all ears. The custom `AuthenticationHandler` is a simple solution, but if you'd like to create something around already existing middleware we could do that as well.

bjorkstromm:
# Description

- This includes functionality which simplifies securing endpoints using Dapr API Token authentication
- Fixes #682

## Issue reference

Please reference the issue this PR will close: #682 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


bjorkstromm:
I'm opening a draft PR to discuss the implementation. Is this how you imagined it? 

// cc @rynowak @Tratcher

src/Dapr.AspNetCore/DaprAuthenticationBuilderExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+        /// </param>
+        /// <returns>A reference to <paramref name="builder"/> after the operation has completed.</returns>
+        public static AuthenticationBuilder AddDapr(this AuthenticationBuilder builder, Func<string> tokenFactory = null)
+        {
+            tokenFactory ??= DaprDefaults.GetDefaultApiToken;
+
+            return builder
+                .AddScheme<DaprAuthenticationHandlerOptions, DaprAuthenticationHandler>(
+                    DaprAuthenticationHandlerOptions.DefaultScheme,
+                    options => options.TokenFactory = tokenFactory);
```
Tratcher, line 34: I'd expect the token value to be read from the IConfiguration here and assigned to the options. Why resolve it per request?

Tratcher:
That's the right outline for implementing an auth handler. It is a strange use of an auth handler though, only accepting a single, hard-coded 'user' credential that doesn't give you any meaningful claims.

examples/AspNetCore/ControllerSample/Controllers/SampleController.cs
```diff
@@ -58,6 +59,7 @@ public ActionResult<Account> Get([FromState(StoreName)] StateEntry<Account> acco
         ///  "pubsub", the first parameter into the Topic attribute, is name of the default pub/sub configured by the Dapr CLI.
         [Topic("pubsub", "deposit")]
         [HttpPost("deposit")]
+        [Authorize("Dapr")]
```
rynowak, line 62: I'd prefer not to change this sample and just add to docs if it's going to make running the samples harder.

src/Dapr.AspNetCore/DaprAuthenticationBuilderExtensions.cs
```diff
@@ -0,0 +1,37 @@
+﻿// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+// ------------------------------------------------------------
+
+namespace Microsoft.AspNetCore.Authentication
+{
+    using System;
+    using Dapr;
+    using Dapr.AspNetCore;
```
rynowak, line 10: Outside the `namespace` please. That's the preferred style and the repo still has a mix unfortunately.

src/Dapr.AspNetCore/DaprAuthenticationHandler.cs
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        private AuthenticateResult HandleAuthenticate()
+        {
+            if (!Request.Headers.TryGetValue(DaprApiToken, out var token))
+            {
+                return AuthenticateResult.NoResult();
+            }
+
+            var expectedToken = Options.TokenFactory?.Invoke();
```
rynowak, line 40: This is an interesting behavioral question - hopefully @Tratcher can answer as well.

So if I configure the new auth handler, but fail to set the expected today via `DAPR_API_TOKEN` it looks like this would always return `NoResult()`. So does that mean that the request is considered unauthenticated?

src/Dapr.AspNetCore/DaprAuthenticationHandler.cs
```diff
@@ -0,0 +1,63 @@
...
+            if (string.IsNullOrWhiteSpace(expectedToken))
+            {
+                return AuthenticateResult.NoResult();
+            }
+
+            if (!string.Equals(token, expectedToken))
+            {
+                return AuthenticateResult.NoResult();
```
rynowak, line 48: Should this be `.Fail()`? The request supplied a value that was not accepted.

src/Dapr.AspNetCore/DaprAuthenticationHandler.cs
```diff
@@ -0,0 +1,63 @@
...
+            }
+
+            var claims = new List<Claim>
+            {
+                new Claim(ClaimTypes.Name, "Dapr")
+            };
+            var identity = new ClaimsIdentity(claims, Options.Scheme);
+            var identities = new List<ClaimsIdentity> { identity };
+            var principal = new ClaimsPrincipal(identities);
+            var ticket = new AuthenticationTicket(principal, Options.Scheme);
```
rynowak, line 58: @Tratcher - any design issues or constraints we should be aware of here?

src/Dapr.AspNetCore/DaprAuthorizationOptionsExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+    /// Provides extension methods for <see cref="AuthorizationOptions" />.
+    /// </summary>
+    public static class DaprAuthorizationOptionsExtensions
+    {
+        /// <summary>
+        /// Adds Dapr authorization policy.
+        /// </summary>
+        /// <param name="options">The <see cref="AuthorizationOptions"/>.</param>
+        /// <param name="name">The name of the policy.</param>
+        public static void AddDapr(this AuthorizationOptions options, string name = "Dapr")
```
rynowak, line 20: Is this policy name just descriptive?

rynowak (commented):
This looks pretty awesome! will need some integration tests for sure.

bjorkstromm:
> 
> 
> This looks pretty awesome! will need some integration tests for sure.

Yes! Will do. Just wanted to make sure implementation details was on right track before adding tests.

src/Dapr.AspNetCore/DaprAuthenticationHandler.cs
Tratcher: Elsewhere I recommended moving the env var resolution to Options and setting it from IConfiguration at startup. That would also be the place to add validation for a missing value. E.g. crash on startup for bad config.

examples/AspNetCore/ControllerSample/Controllers/SampleController.cs
bjorkstromm: Removed.

src/Dapr.AspNetCore/DaprAuthenticationBuilderExtensions.cs
bjorkstromm: I changed the signature to use `Action<DaprAuthenticationOptions> configureOptions` instead, just like when configuring other authentication handlers. Or did you specifically want this from `IConfiguration`?

src/Dapr.AspNetCore/DaprAuthenticationBuilderExtensions.cs
bjorkstromm: Fixed

src/Dapr.AspNetCore/DaprAuthenticationHandler.cs
bjorkstromm: Token is now a string property on `Options`. I changed to `Fail` with a error message.

src/Dapr.AspNetCore/DaprAuthenticationHandler.cs
bjorkstromm: Changed.

src/Dapr.AspNetCore/DaprAuthorizationOptionsExtensions.cs
bjorkstromm: The name needs to match the policy used when adding `AuthorizeAttribute`. E.g. `[Authorize("Dapr")]`

xavidram:
Will this be getting some traction? This pull request is very helpful when securing service controllers. I had to do this manually as well, so having this in the Dapr dotnet-sdk would help out a lot in the log run.

mpalmer78:
Please implement this.

rynowak (approved):

rynowak:
Thanks @bjorkstromm

xavidram:
Thank you for completing this merge. Much appreciated!

Issue #585, opened 2021-02-12
F# Invoke Timer doesn't work


LukePammant:
## Expected Behavior

I should be able to register a Timer in an F# Actor and have it call the appropriate method.

## Actual Behavior

The dotnet sdk throws an error on `ValidateTimerCallback`. 

The first error occurs on [`Actor.cs ln:335`](https://github.com/dapr/dotnet-sdk/blob/master/src/Dapr.Actors/Runtime/Actor.cs#L335) is:
`Timer callback method: {callback} does not exist in the Actor class: {actorTypeName}` 
This can be worked around by prefixing the method name (callback argument) with "get_". For example the `Tick` method must be registered to the timer as `get_Tick`. 

The second error is when validating that the return type is a `Task`. Even if the method just returns `Task.CompletedTask` [this line ](https://github.com/dapr/dotnet-sdk/blob/master/src/Dapr.Actors/Runtime/Actor.cs#L348) will throw an error saying that the return type is not a task. 

When I inspect the ReturnType while debugging I see that the F# `Task` is: `Microsoft.FSharp.Core.FSharpFunc``2[[Microsoft.FSharp.Core.Unit, FSharp.Core, Version=5.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a],[System.Threading.Tasks.Task, System.Private.CoreLib, Version=5.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]], FSharp.Core, Version=5.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a`

## Steps to Reproduce the Problem

From an F# actor call the Start method:
(POST {{daprUrl}}/v1.0/actors/MyActor/myactorid/method/Start)
`
member this.Start() =
                task {
                    do!
                        this.RegisterTimerAsync(
                            "Tick",
                            "get_Tick",
                            null,
                            TimeSpan.FromMilliseconds(float 1000),
                            TimeSpan.FromMilliseconds(float 1000)
                        )
                        |> Async.AwaitTask
                        |> Async.Ignore
                }  :> Task
`

The tick method can be:
`member this.Tick() = Task.CompletedTask`

The tick method can be called via Dapr without issues via:
(POST {{daprUrl}}/v1.0/actors/MyActor/myactorid/method/Tick)

Maybe the dotnet sdk can fall back to calling the method the same way the InvokeMethod does without trying to cast the method to a `Task`?
[...]

PR #689, opened 2021-05-31
F# actor timer support

LukePammant:
# Description

This PR add some minor tweaks to the reflection code used by the SDK to validate that an actor method is callable by the actor runtime. 

## Issue reference

#585

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/src/Dapr.Actors/Runtime/Actor.cs
+++ b/src/Dapr.Actors/Runtime/Actor.cs
@@ -300,7 +300,8 @@ namespace Dapr.Actors.Runtime
 
         internal MethodInfo GetMethodInfoUsingReflection(Type actorType, string callback)
         {
-            return actorType.GetMethod(callback, BindingFlags.Public | BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.Static);
+            return actorType.GetMethod(callback, BindingFlags.Public | BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.Static)
+                ?? actorType.GetMethod($"get_{callback}", BindingFlags.Public | BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.Static); // F# support
         }
 
         internal void ValidateTimerCallback(ActorHost host, string callback)
@@ -333,7 +334,8 @@ namespace Dapr.Actors.Runtime
             }
 
             // The timer callback should return only Task return type
-            if (methodInfo.ReturnType != typeof(Task))
+            if (methodInfo.ReturnType != typeof(Task) 
+                && methodInfo.ReturnType.GetMethod("Invoke")?.ReturnType != typeof(Task)) // Support for F# 'task' computation expressions
             {
                 throw new ArgumentException("Timer callback can only return type Task");
             }
--- a/src/Dapr.Actors/Runtime/ActorManager.cs
+++ b/src/Dapr.Actors/Runtime/ActorManager.cs
@@ -209,7 +209,7 @@ namespace Dapr.Actors.Runtime
         internal async Task FireTimerAsync(ActorId actorId, Stream requestBodyStream, CancellationToken cancellationToken = default)
         {
             #pragma warning disable 0618
-            var timerData = await JsonSerializer.DeserializeAsync<TimerInfo>(requestBodyStream);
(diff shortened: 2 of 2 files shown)


src/Dapr.Actors/Runtime/Actor.cs
```diff
@@ -300,7 +300,8 @@ protected async Task UnregisterTimerAsync(string timerName)
 
         internal MethodInfo GetMethodInfoUsingReflection(Type actorType, string callback)
         {
-            return actorType.GetMethod(callback, BindingFlags.Public | BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.Static);
+            return actorType.GetMethod(callback, BindingFlags.Public | BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.Static)
+                ?? actorType.GetMethod($"get_{callback}", BindingFlags.Public | BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.Static); // F# support
```
rynowak, line 304: Can you add a link to https://github.com/fsharp/fsharp/issues/840 or a more descriptive comment here? I worry that the details about *why* this needs to be done this way will be lost to time.

src/Dapr.Actors/Runtime/Actor.cs
```diff
@@ -333,7 +334,8 @@ internal void ValidateTimerCallback(ActorHost host, string callback)
             }
 
             // The timer callback should return only Task return type
-            if (methodInfo.ReturnType != typeof(Task))
+            if (methodInfo.ReturnType != typeof(Task) 
+                && methodInfo.ReturnType.GetMethod("Invoke")?.ReturnType != typeof(Task)) // Support for F# 'task' computation expressions
```
rynowak, line 338: It feels like this check should be more robust. Like we should only look for `Invoke()` based on the type of `methodInfo.ReturnType`. 

What's the type that we're looking at here? Is it this? https://fsharp.github.io/fsharp-core-docs/reference/fsharp-core-fsharpfunc-2.html

src/Dapr.Actors/Runtime/ActorManager.cs
```diff
@@ -209,16 +209,13 @@ async Task<byte[]> RequestFunc(Actor actor, CancellationToken ct)
         internal async Task FireTimerAsync(ActorId actorId, Stream requestBodyStream, CancellationToken cancellationToken = default)
         {
             #pragma warning disable 0618
-            var timerData = await JsonSerializer.DeserializeAsync<TimerInfo>(requestBodyStream);
+            var timerData = await JsonSerializer.DeserializeAsync<TimerInfo>(requestBodyStream, jsonSerializerOptions);
```
rynowak, line 212: Can you revert this change and add a comment instead? 

TimerInfo and other types provided by the Actor SDK are written to have no behavior that changes based on the options.

src/Dapr.Actors/Runtime/ActorManager.cs
```diff
@@ -209,16 +209,13 @@ async Task<byte[]> RequestFunc(Actor actor, CancellationToken ct)
...
             #pragma warning restore 0618
 
             // Create a Func to be invoked by common method.
             async Task<byte[]> RequestFunc(Actor actor, CancellationToken ct)
             {
-                var actorTypeName = actor.Host.ActorTypeInfo.ActorTypeName;
-                var actorType = actor.Host.ActorTypeInfo.ImplementationType;
-                var methodInfo = actor.GetMethodInfoUsingReflection(actorType, timerData.Callback);
-
+                var methodInfo = this.actorMethodInfoMap.LookupActorMethodInfo(timerData.Callback);
```
rynowak, line 218: What functionally changes here?

src/Dapr.Actors/Runtime/Actor.cs
```diff
@@ -333,7 +334,8 @@ internal void ValidateTimerCallback(ActorHost host, string callback)
             }
 
             // The timer callback should return only Task return type
-            if (methodInfo.ReturnType != typeof(Task))
+            if (methodInfo.ReturnType != typeof(Task)
```
rynowak, line 337: Do we need to do the same for reminders? or is this shared code?

rynowak (changes requested):
Looks like this isn't quite complete yet. At a minimum it needs tests for the new support.

Issue #695, opened 2021-06-08
Support .net 6 preview 4 new Minimal APIs


doddgu:
Hi, is there anyone working on Minimal APIs support?

doddgu:
Maybe adding extensions for IEndpointRouteBuilder will work?

doddgu:
in DaprEndpointRouteBuilderExtensions.cs
I added an extension method. It's looks like worked, but I only test invocation.
```csharp
        /// <summary>
        /// Adds Dapr integration for Minimal APIs to the provided <see cref="IEndpointRouteBuilder" />.
        /// </summary>
        /// <param name="builder">The <see cref="IEndpointRouteBuilder" />.</param>
        /// <param name="services">The <see cref="IServiceCollection" />.</param>
        /// <param name="configureClient">The (optional) <see cref="DaprClientBuilder" /> to use for configuring the DaprClient.</param>
        /// <returns>The <see cref="IEndpointRouteBuilder" /> builder.</returns>
        public static IEndpointRouteBuilder AddDapr(this IEndpointRouteBuilder builder, IServiceCollection services, Action<DaprClientBuilder> configureClient = null)
        {
            if (builder is null)
            {
                throw new ArgumentNullException(nameof(builder));
            }

            // This pattern prevents registering services multiple times in the case AddDapr is called
            // by non-user-code.
            if (services.Any(s => s.ImplementationType == typeof(DaprMinimalApisMarkerService)))
            {
                return builder;
            }

            services.AddDaprClient(configureClient);

            services.AddSingleton<DaprMinimalApisMarkerService>();
            services.AddSingleton<IApplicationModelProvider, StateEntryApplicationModelProvider>();
            services.Configure<MvcOptions>(options =>
            {
                options.ModelBinderProviders.Insert(0, new StateEntryModelBinderProvider());
            });

            return builder;
        }
```

doddgu:
We found sub isn't support too.

doddgu:
I found another exception, result in the actor function does not work.
see #https://github.com/dotnet/aspnetcore/issues/33964

And MapActorsHandlers function throw exception too.
Because endpoints.ServiceProvider.GetService<ActorRuntime>() return null. I change it to builder.Services.BuildServiceProvider().GetService<ActorRuntime>() , it's working, but health check broken down.

davidfowl:
Explicit support shouldn't need to be added for minimal APIs. There are a few things decoupled from MVC won't be supported:
[...]

PR #696, opened 2021-06-08
Minimal APIs support

doddgu:
# Description

added AddDapr extension method for DaprEndpointRouteBuilderExtensions to support minimal APIs. And added tests for this, but I only tested invocation manually, I'm not sure the other features are working.

## Issue reference

Please reference the issue this PR will close: #695

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/src/Dapr.AspNetCore/DaprEndpointRouteBuilderExtensions.cs
+++ b/src/Dapr.AspNetCore/DaprEndpointRouteBuilderExtensions.cs
@@ -11,6 +11,10 @@ namespace Microsoft.AspNetCore.Builder
     using System.Threading;
     using System.Threading.Tasks;
     using Dapr;
+    using Dapr.AspNetCore;
+    using Dapr.Client;
+    using Microsoft.AspNetCore.Mvc;
+    using Microsoft.AspNetCore.Mvc.ApplicationModels;
     using Microsoft.AspNetCore.Routing;
     using Microsoft.AspNetCore.Routing.Patterns;
     using Microsoft.Extensions.DependencyInjection;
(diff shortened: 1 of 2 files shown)


davidfowl:
See my comment, I don't think any of this is required.

src/Dapr.AspNetCore/DaprEndpointRouteBuilderExtensions.cs
```diff
@@ -21,6 +25,39 @@ namespace Microsoft.AspNetCore.Builder
...
     public static class DaprEndpointRouteBuilderExtensions
     {
+        /// <summary>
+        /// Adds Dapr integration for Minimal APIs to the provided <see cref="IEndpointRouteBuilder" />.
+        /// </summary>
+        /// <param name="builder">The <see cref="IEndpointRouteBuilder" />.</param>
+        /// <param name="services">The <see cref="IServiceCollection" />.</param>
+        /// <param name="configureClient">The (optional) <see cref="DaprClientBuilder" /> to use for configuring the DaprClient.</param>
+        /// <returns>The <see cref="IEndpointRouteBuilder" /> builder.</returns>
+        public static IEndpointRouteBuilder AddDapr(this IEndpointRouteBuilder builder, IServiceCollection services, Action<DaprClientBuilder> configureClient = null)
```
rynowak, line 35: The design of this does not really make sense to me. I would expect the **services** to already be baked by the time you have an endpoint route builder.

rynowak, line 35: The design of this does not really make sense to me. I would expect the **services** to already be baked by the time you have an endpoint route builder. 

It's still not super clear to me what (if any) changes are needed and what scenarios are being enabled. Let's use the issue to discuess.

rynowak:
Closing in favor of #738

Issue #661, opened 2021-04-20
Add support for actor reentrancy


halspang:
## Describe the feature
The dapr core runtime is planning on supporting actor reentrancy. The bulk of this logic will be handled in dapr's runtime but the SDK will be responsible for:

- Preserving the header `dapr-reentrant-id`
- Providing the new config fields required by `dapr/config`

For more context on actor reentrancy, the design proposal can be found here: [actor reentrancy](https://github.com/dapr/dapr/issues/2901)

## Release Note
RELEASE NOTE: **ADD** Support for actor reentrancy

tomkerkhove:
Are there any docs on this? I'm probably looking over it.

halspang:
Nothing SDK specific yet, but there's info on the feature itself.

https://docs.dapr.io/developing-applications/building-blocks/actors/actor-reentrancy/

tomkerkhove:
I have found those.

So if we want to use it, we just have to use the host options?

sregdorr:
Just curious... If I'm understanding the PR correctly, this is configured in the host options at the runtime level. Does that mean that you cant configure reentrancy at the actor level? Either all actors are reentrant, or none of them are?

halspang:
Yes, you have to use the host options which means all actor types registered with those options are reentrant. We do not have the fine-grained settings for individual actor reentrancy. @sregdorr - What is your use case for having an individual actor be reentrant instead of the entire type?

sregdorr:
Thanks for the info @halspang! I have an app that is currently utilizing dotnet Orleans and am interested in moving it to Dapr to allow for a polyglot system. I have a couple of grains (actors) whose sole purpose are simply to shuttle tasks to a set of other grains that actually do the legwork. For those shuttle grains, identity is not important and single-threaded operation is not desired. I always identify them with a `Guid.Empty` and In Orleans I can make them reentrant by setting the [Reentrant] attribute on the grain class. 

I do realize that this may be an anti-pattern with regard to the actor model however.  If I understand correctly, identity of an actor is pretty much the point! It seems that these shuttle grains might be better utilized as a simple gRPC service.

halspang:
@sregdorr Gotcha, thanks for explaining your use case! 

If identity/single-threaded isn't required, could you just give them a unique ID? Instead of `Guid.Empty` do a `Guid.NewGuid`?

[...]

PR #708, opened 2021-07-01
Add support for Actor Reentrancy

halspang:
# Description
This commit allows for the Dapr-Reentrancy-Id header to be
propogated through an actor's call chain which lets the dapr
runtime allow reentrant calls.

## Issue reference

Please reference the issue this PR will close: #661 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


src/Dapr.Actors/Helper.cs
```diff
@@ -26,5 +27,19 @@ public static string GetCallContext()
                 return Guid.NewGuid().ToString();
             }
         }
+
+        public static string GetReentrancyContext()
```
rynowak, line 31: These methods should be defined on `ActorLogicalCallContext`. Putting them on a class called `Helper` just obfuscates their purpose.

src/Dapr.Actors/Reentrancy/ActorLogicalReentrancyContext.cs
```diff
@@ -0,0 +1,36 @@
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+// ------------------------------------------------------------
+
+namespace Dapr.Actors.Reentrancy
+{
+    using System;
+    using System.Threading;
```
rynowak, line 9: Place `usings` outside the namespace declaration. We haven't bulk updated all of the files, but this is the preferred style and should be followed for new code.

src/Dapr.Actors/Reentrancy/ActorLogicalReentrancyContext.cs
```diff
@@ -0,0 +1,36 @@
...
+
+    internal static class ActorLogicalReentrancyContext
+    {
+        private static readonly AsyncLocal<string> fabActAsyncLocal = new AsyncLocal<string>();
```
rynowak, line 13: This could have a better variable name... maybe just something like `state`? I'm not sure what a `fabAct` is or why the type name is part of the variable.

src/Dapr.Actors.AspNetCore/ActorsEndpointRouteBuilderExtensions.cs
```diff
@@ -71,13 +72,20 @@ private static IEndpointConventionBuilder MapActorDeactivationEndpoint(this IEnd
         private static IEndpointConventionBuilder MapActorMethodEndpoint(this IEndpointRouteBuilder endpoints)
         {
             var runtime = endpoints.ServiceProvider.GetRequiredService<ActorRuntime>();
+            var logger = endpoints.ServiceProvider.GetRequiredService<ILoggerFactory>().CreateLogger<ILogger>();
```
rynowak, line 75: Is this used for anything?

src/Dapr.Actors/DaprHttpInteractor.cs
```diff
@@ -113,6 +113,12 @@ HttpRequestMessage RequestFunc()
 
                 request.Headers.Add(Constants.RequestHeaderName, Encoding.UTF8.GetString(serializedHeader, 0, serializedHeader.Length));
 
+                var reentrancyId = Helper.GetReentrancyContext();
+                if (reentrancyId != string.Empty)
```
rynowak, line 117: Use `null` as a sentinel value, not empty string. It's a better practice in general to use `null` as a sentinel for *unset* because then if you have a bug you will get a stack trace instead of wierd behavior.

src/Dapr.Actors/Helper.cs
```diff
@@ -26,5 +27,19 @@ public static string GetCallContext()
                 return Guid.NewGuid().ToString();
             }
         }
+
```
rynowak, line 30: Is `GetCallContext()` ^^^ used for anything after your changes? I have a feeling this is vestigial and can be removed.

src/Dapr.Actors/Helper.cs
```diff
@@ -26,5 +27,19 @@ public static string GetCallContext()
...
+        public static string GetReentrancyContext()
+        {
+            if (ActorLogicalReentrancyContext.TryGet(out var reentrancyContext))
+            {
+                return reentrancyContext;
+            }
+            return string.Empty;
```
rynowak, line 37:
```suggestion
            return null;
```

src/Dapr.Actors/Reentrancy/ActorLogicalReentrancyContext.cs
```diff
@@ -0,0 +1,36 @@
+// ------------------------------------------------------------
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+// ------------------------------------------------------------
+
+namespace Dapr.Actors.Reentrancy
```
rynowak, line 6: How many types are in this namespace? It seems unjustified to add this. I'd suggest just putting it in `Dapr.Actors`

src/Dapr.Actors/Runtime/ActorRuntime.cs
```diff
@@ -93,6 +93,17 @@ internal Task SerializeSettingsAndRegisteredTypes(IBufferWriter<byte> output)
                 writer.WriteBoolean("drainRebalancedActors", (this.options.DrainRebalancedActors));
             }
 
+            if (this.options.ReentrancyConfig != null) {
```
rynowak, line 96: Allman style :)

src/Dapr.Actors/Runtime/ActorRuntime.cs
```diff
@@ -93,6 +93,17 @@ internal Task SerializeSettingsAndRegisteredTypes(IBufferWriter<byte> output)
...
+                writer.WriteStartObject("reentrancy");
+                
+                writer.WriteBoolean("enabled", this.options.ReentrancyConfig.Enabled);
+                if (this.options.ReentrancyConfig.MaxStackDepth != null) {
```
rynowak, line 100: Allman style (brace on separate line)

src/Dapr.Actors/Runtime/ActorRuntimeOptions.cs
```diff
@@ -119,6 +121,25 @@ public bool DrainRebalancedActors
             }
         }
 
+        /// <summary>
+        /// A configuration that defines the Actor Reentrancy parameters. If set and enabled, Actors
+        /// will be able to perform reentrant calls to themselves/others. If not set or false, Actors
+        /// will continue to lock for every request.
+        /// See https://docs.dapr.io/developing-applications/building-blocks/actors/actor-reentrancy/
+        /// </summary>
+        public ActorReentrancyConfig ReentrancyConfig
```
rynowak, line 130: IMO this is should be initialized to a default value so you can do `options.ReentrancyConfig.Foo = bar`.

test/Dapr.Actors.Test/Runtime/ActorRuntimeOptionsTests.cs
```diff
@@ -113,5 +112,30 @@ public void SettingJsonSerializerOptionsToNull_Fails()
 
             action.Should().Throw<ArgumentNullException>();
         }
+
+        [Fact]
+        public void SettingReentrancyConfigWithoutMaxStackDepth_Succeeds()
```
rynowak, line 117: I would prefer not having tests like this that just set properties and read them back. It's testing code that we *didn't* write.

test/Dapr.Actors.Test/Runtime/ActorRuntimeTests.cs
```diff
@@ -144,6 +144,136 @@ public async Task TestActorSettings()
 
             element = root.GetProperty("drainRebalancedActors");
             Assert.True(element.GetBoolean());
+
+            try
+            {
+                element = root.GetProperty("reentrancy");
+                Assert.True(false);
+            } catch (System.Collections.Generic.KeyNotFoundException)
```
rynowak, line 152: https://docs.microsoft.com/en-us/dotnet/api/system.text.json.jsonelement.trygetproperty?view=net-5.0

test/Dapr.Actors.Test/Runtime/ActorRuntimeTests.cs
```diff
@@ -144,6 +144,136 @@ public async Task TestActorSettings()
...
+        {
+            var actorType = typeof(TestActor);
+
+            var options = new ActorRuntimeOptions();
+            options.Actors.RegisterActor<TestActor>();
+            options.ActorIdleTimeout = TimeSpan.FromSeconds(33);
+            options.ActorScanInterval = TimeSpan.FromSeconds(44);
+            options.DrainOngoingCallTimeout = TimeSpan.FromSeconds(55);
+            options.DrainRebalancedActors = true;
+            options.ReentrancyConfig = new Reentrancy.ActorReentrancyConfig {
```
rynowak, line 169: https://www.youtube.com/watch?v=jUTORC4eoGc

test/Dapr.Actors.Test/Runtime/ActorRuntimeTests.cs
```diff
@@ -144,6 +144,136 @@ public async Task TestActorSettings()
...
+            Assert.Equal(TimeSpan.FromSeconds(33), ConverterUtils.ConvertTimeSpanFromDaprFormat(element.GetString()));
+
+            element = root.GetProperty("actorScanInterval");
+            Assert.Equal(TimeSpan.FromSeconds(44), ConverterUtils.ConvertTimeSpanFromDaprFormat(element.GetString()));
+
+            element = root.GetProperty("drainOngoingCallTimeout");
+            Assert.Equal(TimeSpan.FromSeconds(55), ConverterUtils.ConvertTimeSpanFromDaprFormat(element.GetString()));
+
+            element = root.GetProperty("drainRebalancedActors");
+            Assert.True(element.GetBoolean());
```
rynowak, line 207: Seems like this is duplicating a lot of logic from an existing test... want to either fold the new checks into it? or skip the redundant stuff?

test/Dapr.E2E.Test.App/Actors/ReentrantActor.cs
```diff
@@ -0,0 +1,57 @@
...
+            : base(host)
+        {
+        }
+
+        public Task Ping()
+        {
+            return Task.CompletedTask;
+        }
+
+        public async Task ReentrantCall(ReentrantCallOptions callOptions)
```
rynowak, line 24: Can you add a description in comments (somewhere, doesn't have to be here), of what this test is testing and why it's meaningful?

test/Dapr.E2E.Test.App/Actors/ReentrantActor.cs
```diff
@@ -0,0 +1,57 @@
...
+                await UpdateState(false);
+            }
+        }
+
+        public Task<State> GetState() 
+        {
+            return this.StateManager.GetOrAddStateAsync<State>("reentrant-record", new State());
+        }
+
+        private async Task UpdateState(bool isEnter)
```
rynowak, line 50: What is `isEnter` mean/do?

rynowak: Notice that it's hardcoded to `true` below

test/Dapr.E2E.Test/Actors/E2ETests.ReentrantTests.cs
```diff
@@ -0,0 +1,42 @@
...
+    using System;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Dapr.Actors;
+    using Dapr.E2E.Test.Actors.Reentrancy;
+    using Xunit;
+
+    public partial class E2ETests : IAsyncLifetime
+    {
+        private static readonly int NUM_CALLS = 10;
```
rynowak, line 16: `NumCalls`

test/Dapr.E2E.Test/Actors/E2ETests.ReentrantTests.cs
```diff
@@ -0,0 +1,42 @@
...
+            var records = state.Records;
+
+            Assert.True(records.Count == NUM_CALLS * 2);
+            for (int i = 0; i < NUM_CALLS * 2; i++)
+            {
+                Assert.Equal(i < NUM_CALLS, records[i].IsEnter);
+
+                if (i > 0)
+                {
+                    Assert.True(records[i].Timestamp > records[i - 1].Timestamp);
```
rynowak, line 37: I'm a little worried about precision here and whether this is good enough resolution everywhere the tests will run (clock frequency varies). Would it be better to use an incrementing atomic inside the test actor?

rynowak (changes requested):
I've left my comments here based on the code changes and it's mostly minor feedback on that point.

What's less minor feedback is how these interactions work with the state store. 

- For instance if I make some *pending* changes to the state store and then make a re-entrant call what happens.
- If I update state inside a `try... finally` (like your test does) and hit an exception due to a network hiccup can I end up with multiple calls executing inside my code simultaneously? [dictionary](https://github.com/dapr/dotnet-sdk/blob/master/src/Dapr.Actors/Runtime/ActorStateManager.cs#L19) is not thread-safe!

src/Dapr.Actors/Helper.cs
halspang: Gotcha. I was following the example set by the Helper class which appeared to be handling a similar context. From what you say below though, it sounds like this may not have been the best example to go off of.

src/Dapr.Actors.AspNetCore/ActorsEndpointRouteBuilderExtensions.cs
halspang: Nope, I had debugging code and removed it but missed the logger.

src/Dapr.Actors/Helper.cs
halspang: It's referenced, but I'm not really sure what it's accomplishing. I can make an issue to remove it if you'd like, but would prefer to not have that in this PR.

test/Dapr.Actors.Test/Runtime/ActorRuntimeOptionsTests.cs
halspang: Fair enough. I included them as the other values had them. Would you like me to actually remove at least my new tests here?

test/Dapr.Actors.Test/Runtime/ActorRuntimeTests.cs
halspang: Noted for the future! With the defaults there I just check those now.

test/Dapr.Actors.Test/Runtime/ActorRuntimeTests.cs
halspang: I'll remove the redundant stuff. I like to leave the old tests as untouched as possible with new features like these that are optional.

test/Dapr.E2E.Test.App/Actors/ReentrantActor.cs
halspang: This is following the pattern for other e2e tests involving actors. Though in this case I forgot to assign it :|

Essentially what it lets us do is just track when an actor's method is entered and left. This let's us see the single threaded path.

test/Dapr.E2E.Test/Actors/E2ETests.ReentrantTests.cs
halspang: I can definitely switch to the AtomicInt, but isn't this also the point of actors? Since it's the same actor host and the execution is guaranteed single threaded, we should be OK, right? Or are you worried that it'll record the same time, not be out of order?

halspang:
Thanks for the feedback! I'll try and address your questions below. This is reminiscent of the conversation we had in the original design, which I don't think we ever concluded on.

>  For instance if I make some _pending_ changes to the state store and then make a re-entrant call what happens.
This is what we never really concluded on in the design. Currently, the way it's implemented, the pending state will remain pending until the user tells it to commit. This potentially means that different methods could commit that if you're calling into the same actor. The way I see it, we can handle this in two ways.

1. State that users are responsible for committing all their state and if they don't want other methods to commit state from prior methods, either name them independently (like a scratch space) or commit before a reentrant call. 
2. If we detect a reentrancy id on an outgoing call, commit the state.

Of those two, I'd rather do the first as customers may not want their state committed until after the reentrant call has finished. We already have them commit manually, correct? If so, this just feels like an updated guidance for using reentrancy in conjunction with this SDK.

If that goes away from the experience we want in the SDK, please let me know as I definitely want to comply with whatever we've already established.

> If I update state inside a `try... finally` (like your test does) and hit an exception due to a network hiccup can I end up with multiple calls executing inside my code simultaneously? [dictionary](https://github.com/dapr/dotnet-sdk/blob/master/src/Dapr.Actors/Runtime/ActorStateManager.cs#L19) is not thread-safe!

This is highlighting what I think is one of the fundamental dangers in reentrancy. In the case of a network interrupt which severs the connection from one side but doesn't kill it on the other, we'll be able to issue a retry where we could technically get into a state where the actor is being called twice. I feel like there are ways we could discuss to make this impossible, but it could also be documented as a thing to be resilient against when using reentrancy specifically. What do you think?

src/Dapr.Actors/Reentrancy/ActorLogicalReentrancyContext.cs
```diff
@@ -0,0 +1,36 @@
...
+{
+    using System;
+    using System.Threading;
+
+    internal static class ActorLogicalReentrancyContext
+    {
+        private static readonly AsyncLocal<string> fabActAsyncLocal = new AsyncLocal<string>();
+
+        public static bool IsPresent()
```
rynowak, line 15: The other change you should make here is to a use a class (not a string) with a property that holds the string.

This allows you to *wipe out* the reentrancy value when the invocation ends. 

Here's an example of this pattern: https://github.com/aspnet/HttpAbstractions/blob/master/src/Microsoft.AspNetCore.Http/HttpContextAccessor.cs#L12

The problem with the current approach is that it's possible (and easy) to start background tasks from inside an actor method that capture the current reentrancy state.

Ex:

```C#
public async Task MyActorThing() 
{
      _ = Task.Run(async () =>
     {
             await Task.Delay(TimeSpan.FromMinutes(5));
             
             // if you use an ActorProxy here, it would be fooled into thinking this is a  re-entrant call, even though it's been 5 minutes...
      });
}
```

test/Dapr.Actors.Test/Runtime/ActorRuntimeOptionsTests.cs
rynowak: Yes, I think it's better just to not have things like this.

src/Dapr.Actors.AspNetCore/ActorsEndpointRouteBuilderExtensions.cs
```diff
@@ -11,6 +11,7 @@ namespace Microsoft.AspNetCore.Builder
     using Microsoft.AspNetCore.Http;
     using Microsoft.AspNetCore.Routing;
     using Microsoft.Extensions.DependencyInjection;
+    using Microsoft.Extensions.Logging;
```
rynowak, line 14: This is unused now so this can be removed.

src/Dapr.Actors/ActorReentrancyConfig.cs
```diff
@@ -0,0 +1,51 @@
...
+        public int MaxStackDepth
+        {
+            get
+            {
+                return this.maxStackDepth;
+            }
+
+            set
+            {
+                this.maxStackDepth = value;
```
rynowak, line 47: What's the default/semantics of this when its value is zero? Does it make sense for this to be `int?` so it defaults to null (unset)

src/Dapr.Actors/ActorReentrancyContextAccessor.cs
```diff
@@ -0,0 +1,48 @@
...
+
+using System.Threading;
+
+namespace Dapr.Actors
+{
+    /// <summary>
+    /// Accessor for the reentrancy context. This provides the necessary ID to continue a reentrant request
+    /// across actor invocations.
+    /// </summary>
+    public class ActorReentrancyContextAccessor
```
rynowak, line 14: This class should be `internal` and `static` - we don't want application code access this magic infrastructure.

src/Dapr.Actors.AspNetCore/ActorsEndpointRouteBuilderExtensions.cs
```diff
@@ -78,6 +79,12 @@ private static IEndpointConventionBuilder MapActorMethodEndpoint(this IEndpointR
                 var actorId = (string)routeValues["actorId"];
                 var methodName = (string)routeValues["methodName"];
 
+                if (context.Request.Headers.ContainsKey(Constants.ReentrancyRequestHeaderName))
+                {
+                    var daprReentrancyHeader = context.Request.Headers[Constants.ReentrancyRequestHeaderName];
+                    ActorReentrancyContextAccessor.ReentrancyContext = daprReentrancyHeader;
```
rynowak, line 85: The other thing we need to do here is **unset** the value after the user-code executes. 

There should be a try/finally around `runtime.Dispatch....` and in the finally, you want to set the context back to null.

rynowak: The **unset** is what fixes where bugs where long-running background tasks started from inside an Actor still think they are running inside an actor.

src/Dapr.Actors/Runtime/Actor.cs
```diff
@@ -62,7 +62,7 @@ protected Actor(ActorHost host)
         /// <summary>
         /// Gets the StateManager for the actor.
         /// </summary>
-        protected IActorStateManager StateManager { get; set; }
+        public IActorStateManager StateManager { get; protected set; }
```
rynowak, line 65: 👍

src/Dapr.Actors/Runtime/ActorManager.cs
```diff
@@ -340,6 +339,13 @@ private async Task<T> DispatchInternalAsync<T>(ActorId actorId, ActorMethodConte
             T retval;
             try
             {
+                // Set the state context of the request, if possible.
+                var stateManager = state.Actor.StateManager;
+                if (typeof(IActorContextualState).IsAssignableFrom(stateManager.GetType()))
```
rynowak, line 344: You can do this a lot simpler. 

- `Type.IsAssignableFrom` is only needed when you have a two types to compare, you have a type and a value.
- Pattern matching (relatively new in C#) makes this really nice

```C#
if (state.Actor.StateManager is IActorContextualState contextualStateManager)
{
    contextualStateManager.SetStateContext(...)
}
```

src/Dapr.Actors/Runtime/ActorManager.cs
```diff
@@ -340,6 +339,13 @@ private async Task<T> DispatchInternalAsync<T>(ActorId actorId, ActorMethodConte
...
+                {
+                    await (stateManager as IActorContextualState).SetStateContext(Helper.GetCallContext());
```
rynowak, line 346: `Helper.GetCallContext()` - I want to see a different name for this. `Helper` or `Util` is a solid clue that this should be factored differently. Will make a suggestion there.

rynowak, line 346: `Helper.GetCallContext()` - I want to see a different name for this. `Helper` or `Util` is a solid clue that this should be factored differently. Will make a suggestion there.

test/Dapr.E2E.Test.App/Actors/ReentrantActor.cs
```diff
@@ -0,0 +1,65 @@
...
+                else
+                {
+                    await actor.ReentrantCall(new ReentrantCallOptions 
+                    { 
+                        CallsRemaining = callOptions.CallsRemaining - 1,
+                        CallNumber = callOptions.CallNumber + 1,
+                    });
+                }                
+            } 
+            finally
```
rynowak, line 46: Do we need the finally here? I'm asking because we discussed this pattern and said we'd tell users to avoid it.

src/Dapr.Actors/Runtime/ActorStateManager.cs
```diff
@@ -3,26 +3,31 @@
...
-    using System.Threading.Tasks;
-    using Dapr.Actors.Resources;
-
-    internal sealed class ActorStateManager : IActorStateManager
+    internal sealed class ActorStateManager : IActorStateManager, IActorContextualState
     {
         private readonly Actor actor;
         private readonly string actorTypeName;
-        private readonly Dictionary<string, StateMetadata> stateChangeTracker;
+        private readonly ConcurrentDictionary<string, Dictionary<string, StateMetadata>> contextualStateChangeTracker;
```
rynowak, line 20: This approach has a pretty serious flaw because you need someone to **remove** the entries from the dictionary. If you think about a very long running application you'd accumulate millions of entries here just sitting in memory.

I'd suggest something different. Instead of:

```C#
private static AsyncLocal<string> context = new AsyncLocal<string>();
```

```C#
private static AsyncLocal<(string, Dictionary<string, StateMetadata>)> context = new AsyncLocal<(string, Dictionary<string, StateMetadata>)>();
```

rynowak (commented):
asdf

src/Dapr.Actors/ActorReentrancyConfig.cs
halspang: If we pass a null through, it assumes a default value on the dapr sidecar. I removed the nullable cause I gave it a default value, but perhaps I was too aggressive in my defaulting?

src/Dapr.Actors/Runtime/ActorManager.cs
halspang: Such fancy language features!

test/Dapr.E2E.Test.App/Actors/ReentrantActor.cs
halspang: Nah, I can move that method out.

Out of curiosity, why do we tell customers to avoid that pattern?

src/Dapr.Actors/Runtime/ActorStateManager.cs
halspang: That's a much better way of doing it :) I was thinking about if I needed to remove the dictionary during a clear/reset but this removes that concern.

src/Dapr.Actors/Runtime/ActorManager.cs
halspang: I just got rid of the call entirely. It was just providing a way to get a new GUID so I just made a new one here.

src/Dapr.Actors/ActorReentrancyConfig.cs
rynowak: discussed offline, it's better if we can avoid building default values into the SDKs, and rely on the runtime's behavior.

src/Dapr.Actors/Runtime/ActorManager.cs
rynowak: great

test/Dapr.E2E.Test.App/Actors/ReentrantActor.cs
rynowak: That pattern means that you commit the result even when a failure happens. It leads to potential concurrency issues accessing the state manager - we largely fixed within the state manager, but user code is not likely to be thread-safe.

src/Dapr.Actors/Runtime/ActorStateManager.cs
rynowak: You'd be surprised what can be a security bug 😉

rynowak (approved):

Issue #705, opened 2021-07-01
Call actor throw InvalidProgamException


doddgu:
## Expected Behavior

Call actor success.

## Actual Behavior

![image](https://user-images.githubusercontent.com/7299249/124070536-8ce5e100-da70-11eb-992e-59febca6536c.png)

## Steps to Reproduce the Problem

Use Minimal APIs.
.Net 6 preview 5
OS: Win 11
IDE: VS 2022 Preview

Github Repo: https://github.com/doddgu/MinimalApiActor.git
Steps: 
1. Run docker compose.
2. Visit https://localhost:8002
3. Click the "Test Actor" button.
Original Exception: The reason is when using MinimalApi, MapActorsHandlers method is using endpoints.ServiceProvider. The ServiceProvider can not get the service.
![image](https://user-images.githubusercontent.com/7299249/124410619-7516a180-dd7d-11eb-8d4d-7beb779ddf5c.png)

Now Exception: I try to add extension, please see #695 , I resolve with builder.Services. But, it throw another InvalidProgramException.

## Release Note

RELEASE NOTE:

davidfowl:
Steps to reproduce the problem are missing…

doddgu:
## Steps to Reproduce the Problem：
Github Repo: https://github.com/doddgu/MinimalApiActor.git
Original Exception: The reason is when using MinimalApi, MapActorsHandlers method is using endpoints.ServiceProvider. The ServiceProvider can not get the service.
![image](https://user-images.githubusercontent.com/7299249/124410619-7516a180-dd7d-11eb-8d4d-7beb779ddf5c.png)
Now Exception: I try to add extension, please see #695 , I resolve with builder.Services. But, it throw another exception in the topic.

doddgu:
Based on web api, it's success.

Issue #699, opened 2021-06-10
Couldn't call actor via using ActorProxy 


thangchung:
## Expected Behavior

I have written the application which emulated the booking system. The code is as below

```csharp
var actorId = new ActorId(request.ShowId.ToString());
var proxy = ActorProxy.Create<IBookingMovieActor>(actorId, nameof(BookingMovieActor));
await proxy.Book(request);
```

I expected that it could call to the `BookingMovieActor` with the code as

```csharp
public class BookingMovieActor : Actor, IBookingMovieActor
{
    private readonly IBookingService _bookingService;
    private readonly ILogger<BookingMovieActor> _logger;
    private readonly StorageState _state;

    public BookingMovieActor(ActorHost host, IBookingService bookingService, ILogger<BookingMovieActor> logger) : base(host)
    {
        _bookingService = bookingService;
        _logger = logger;
        _state = new StorageState();
    }

    public async Task Book(BookRequest request)
    {
        _logger.LogInformation($"Trying to book SeatNumber={request.SeatNumber}");

        await _bookingService.Book(_state, request, _logger);
    }
}
```

## Actual Behavior

But when I run `dapr` extension on `tye` with `tye run`

```yaml
name: dapr-actor
extensions:
- name: dapr
  log-level: debug
  config: appconfig
  components-path: "components/"
services:
- name: server
  project: src/Server/Server.csproj
  bindings:
  - port: 5000
```

And access to the route `POST: http://localhost:5000/Weather`

```csharp
[HttpPost]
public async Task Post([FromBody] BookRequest request)
{
    var actorId = new ActorId(request.ShowId.ToString());
    var proxy = ActorProxy.Create<IBookingMovieActor>(actorId, nameof(BookingMovieActor));
    await proxy.Book(request);
}
```

I always got the error

```
Dapr.Actors.ActorMethodInvocationException: Remote Actor Method Exception
 ---> Dapr.Actors.ActorInvokeException: Common Language Runtime detected an invalid program.
   --- End of inner exception stack trace ---
   at Dapr.Actors.DaprHttpInteractor.InvokeActorMethodWithRemotingAsync(ActorMessageSerializersManager serializersManager, IActorRequestMessage remotingRequestRequestMessage, CancellationToken cancellationToken) in Dapr.Actors.dll:token 0x6000047+0x238
[...]

PR #711, opened 2021-07-07
fix: InvalidProgramException under ASP.Net 6.0

jcdickinson:
# Description

Call `Type.GetTypeFromHandle` when calling `IActorRequestMessageBody.GetParameter`

The code that generates the `IActorRequestMessageBody` parameter setup for the dispatched call is technically incorrect, as it passes a `RuntimeTypeHandle` to a method expecting `Type`. The correct thing to do is to call `Type.GetTypeHandle` first to perform the conversion.

It looks like RyuJIT is now validating this, where it wasn't before. It is possible the 5.0 was automatically doing the coercion, or would have crashed if the branch was ever executed (it's pretty likely that this branch has never been executed).

The test will only fail when the test project and global.json are configured for the 6.0 SDK.

## Issue reference

Please reference the issue this PR will close: #705, #699

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation - N/A

### DIFF
--- a/src/Dapr.Actors/Builder/MethodDispatcherBuilder.cs
+++ b/src/Dapr.Actors/Builder/MethodDispatcherBuilder.cs
@@ -1,4 +1,4 @@
-// ------------------------------------------------------------
+﻿// ------------------------------------------------------------
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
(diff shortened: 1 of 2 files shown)


rynowak (approved):
Thanks!

PR #713, opened 2021-07-09
fix: MapActorsHandlers support minimal APIs

doddgu:
# Description

I'm using minimal APIs, I found `endpoints.ServiceProvider.GetService<ActorRuntime>()` always return null. I fix this.
But I found another exception #705 , thanks @jcdickinson fix it. 
Now, my fix and #711 can make actor run success.

## Issue reference

Please reference the issue this PR will close: #695 #705 #711

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


rynowak:
@doddgu - is there a change needed here? The issue was closed without an explanation so I'm working if there is a fix needed in the SDK or not.

doddgu:
> @doddgu - is there a change needed here? The issue was closed without an explanation so I'm working if there is a fix needed in the SDK or not.

It works on preview 7. I think this pr can close.

Issue #709, opened 2021-07-05
Add support for subscribing to raw messages


sapinderpalsingh:
## Describe the proposal
Dapr 1.2 supports subscribing to raw messages which are non cloud events. We can either change the current `MapSubscribeHandler` by adding a boolean flag to it or we can create a new extension method to support subscribing to raw messages.
Also third option is configuring it at the topic level by adding one more boolean argument for processing raw messages, which can provide more flexibility where we can have one topic configured for cloud events and other for raw messages.

```
[Topic("pubsub", "topic",true)]
[HttpPost("onEvent")]
public async Task<IActionResult> OnEvent([FromBody] Event data)
{
}
```

Currently i have created a separate extension method to support this in our recent client engagement.

tanvigour:
@rynowak Any thoughts on this?

rynowak:
Hi @sapinderpalsingh 


> We can either change the current MapSubscribeHandler by adding a boolean flag to it or we can create a new extension method to support subscribing to raw messages.

I would suggest adding a new overload that takes an options class. This is preferrable to a boolean argument for scannability and makes it easy to add other settings in the future without changing the signature. Something like:

```C#
public static IEndpointConventionBuilder MapSubscribeHandler(this IEndpointRouteBuilder endpoints, SubscribeOptions options)
{
}

public class SubscribeOptions
{
    public bool EnableRawPayload { get; set; }
}
```

> Also third option is configuring it at the topic level by adding one more boolean argument for processing raw messages, which can provide more flexibility where we can have one topic configured for cloud events and other for raw messages.

It seems really useful to have both of these options. You have the ability to configure globally, as well as per-topic for the app.

I'd suggest something like:

```C#
public class TopicAttribute : Attribute, ITopicMetadata, IRawTopicMetadata
{
    ...

    public bool? EnableRawPayload { get; set; }
}

public interface IRawTopicMetadata
{
    bool? EnableRawPayload { get; set; }
}
```

The extra interface is needed because we can't add properties to `ITopicMetadata` without a breaking change.

codeputer:
I often wonder why method signatures like this don't do more of what HTTP did with the payload .  Don't get me wrong, I think Http is so loosely opinionated that it results in so many bad practices.

That said, why not have that overloaded parameter do what cloud events did, in that the interface describes the contract as well as the payload?

With three properties, your done
Contract Name,
Contract Version,
[...]

PR #717, opened 2021-07-15
Added support for processing raw messages.

sapinderpalsingh:
# Description

This PR adds support for subscribing to raw messages.

## Issue reference

This PR will close issue #709 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


sapinderpalsingh:
@rynowak , should we change the existing `MapSubscribeHandler` as well? This will be useful if the `RawPayLoad` is enabled at the `TopicAttribute` level and user is calling `endpoints.MapSubscribeHandler()` in startup.cs.

src/Dapr.AspNetCore/DaprEndpointRouteBuilderExtensions.cs
```diff
@@ -78,5 +77,67 @@ public static IEndpointConventionBuilder MapSubscribeHandler(this IEndpointRoute
...
+
+        /// <summary>
+        /// Maps an endpoint that will respond to requests to <c>/dapr/subscribe</c> from the
+        /// Dapr runtime.
+        /// </summary>
+        /// <param name="endpoints">The <see cref="IEndpointRouteBuilder" />.</param>
+        /// <param name="options">Configuration options</param>
+        /// <returns>The <see cref="IEndpointConventionBuilder" />.</returns>
+        /// <seealso cref="MapSubscribeHandler(IEndpointRouteBuilder)"/>
+        public static IEndpointConventionBuilder MapSubscribeHandler(this IEndpointRouteBuilder endpoints, SubscribeOptions options)
```
rynowak, line 89: This duplicates all of the existing functionality. Can you make the existing method call into this one?

src/Dapr.AspNetCore/DaprEndpointRouteBuilderExtensions.cs
```diff
@@ -78,5 +77,67 @@ public static IEndpointConventionBuilder MapSubscribeHandler(this IEndpointRoute
...
+                var entries = dataSource.Endpoints
+                    .OfType<RouteEndpoint>()
+                    .Where(e => e.Metadata.GetMetadata<ITopicMetadata>()?.Name != null)   // only endpoints which have  TopicAttribute with not null Name.
+                    .Distinct()
+                    .Select(e => (e.Metadata.GetMetadata<ITopicMetadata>().PubsubName, e.Metadata.GetMetadata<ITopicMetadata>().Name, e.Metadata.GetMetadata<IRawTopicMetadata>()?.EnableRawPayload, e.RoutePattern));
+
+
+                var logger = context.RequestServices.GetService<ILoggerFactory>().CreateLogger("DaprTopicSubscription");
+
+                ArrayList subscriptions = new();
```
rynowak, line 108: We should not use `ArrayList`. I'd suggest either using a strongly-typed collection like `List<>` or doing the approach used in the other method (just loop directly with no intermediate list).

src/Dapr.AspNetCore/DaprEndpointRouteBuilderExtensions.cs
```diff
@@ -78,5 +77,67 @@ public static IEndpointConventionBuilder MapSubscribeHandler(this IEndpointRoute
...
+                        .Select(part => part.Content))));
+
+                    subscriptions.Add(new
+                    {
+                        topic = entry.Name,
+                        pubsubName = entry.PubsubName,
+                        route,
+                        metadata = new
+                        {
+                            rawPayload = (entry.EnableRawPayload == true) ? "true" : (entry.EnableRawPayload == false) ? "false" : options.EnableRawPayload.ToString().ToLower()
```
rynowak, line 134: Is `rawPayload` supposed to be a string in the response or a boolean? It looks like this is going to serialize as: `{ "rawPayload": "true" }` instead of `{ "rawPayload": true }`.

You can probably simplify this to `rawPayload = entry.EnableRawPayload ?? false`

src/Dapr.AspNetCore/TopicAttribute.cs
```diff
@@ -26,10 +26,29 @@ public TopicAttribute(string pubsubName, string name)
             this.PubsubName = pubsubName;
         }
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="TopicAttribute" /> class.
+        /// </summary>
+        /// <param name="pubsubName">The name of the pubsub component to use.</param>
+        /// <param name="name">The topic name.</param>
+        /// <param name="enableRawPayload">The enable/disable raw pay load flag.</param>
+        public TopicAttribute(string pubsubName, string name, bool enableRawPayload)
```
rynowak, line 35: I'd suggest skipping the optional constructor and making the property get/set instead. This is slightly more idiomatic for optional values in an attribute.

rynowak (changes requested):
A few points of feedback about the implementation but overall looks on-track. 

This definitely needs a unit test. You should be able to find an existing test that you can update or add a new one.

src/Dapr.AspNetCore/DaprEndpointRouteBuilderExtensions.cs
sapinderpalsingh: This is supposed to be string and not boolean.
https://docs.dapr.io/developing-applications/building-blocks/pubsub/pubsub-raw

src/Dapr.AspNetCore/TopicAttribute.cs
sapinderpalsingh: I created a separate constructor to take care of following scenarios:
true => override global setting with true
false => override global setting with false
null => use global setting
If I use the existing constructor then we won't be able to take care of the last one, since we can't have nullable arguments to attribute constructor.

sapinderpalsingh: Also, if we add a new argument to the existing constructor then that will become a breaking change as users will need to update code after they update the SDK.

src/Dapr.AspNetCore/DaprEndpointRouteBuilderExtensions.cs
sapinderpalsingh: Refactored code.

src/Dapr.AspNetCore/DaprEndpointRouteBuilderExtensions.cs
rynowak: Ah you're right!

src/Dapr.AspNetCore/Subscription.cs
```diff
@@ -0,0 +1,44 @@
...
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+// ------------------------------------------------------------
+
+namespace Dapr
+{
+    /// <summary>
+    /// This class defines subscribe endpoint response
+    /// </summary>
+    public class Subscription
```
rynowak, line 11: These types need to be internal, they are not interesting for users of the SDK. Or if you like, convert back to manual JSON which I kinda prefer.

rynowak (changes requested):
New types for Subscription and metadata need to be internal

src/Dapr.AspNetCore/Subscription.cs
sapinderpalsingh: IMO, strongly typed subscription can be easily managed in future when the metadata object grows with more properties. So marking it as internal as you suggested.

rynowak (approved):

PR #735, opened 2021-08-12
Update pubsub example.

JhouXerox:
# Description
I add an example about how to pub and how to sub in .NET SDK.

## What's Include
An ASP.NET WebAPI project named `Publisher` shows how to publish with custom model.
An ASP.NET WebAPI project named `DeclarativeSubscriber` shows how to subscribe topic by yaml file(Declarative) and deal the event.
An ASP.NET WebAPI project named `ProgrammaticSubscriber` shows how to subscribe topic by coding(Programmatic) and deal the event.


examples/PubSub/ProgrammaticSubscriberDotNet/Controllers/Order.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Threading.Tasks;
+using Dapr;
+using Dapr.Client;
+using Microsoft.AspNetCore.Mvc;
+using PubSub.Domain;
+
+namespace ProgrammaticSubscriberDotNet.Controllers
+{
+    [ApiController]
+    [Route("Order")]
```
halspang, line 15: Is the `Route` required here?

examples/PubSub/DeclarativeSubscriber/Controllers/Order.cs
```diff
@@ -0,0 +1,33 @@
...
+    [ApiController]
+    [Route("Order")]
+    public class Order : ControllerBase
+    {
+        private readonly string reporter = "====================\r\n" +
+                                           "OrderId:\t\t{0}\r\n" +
+                                           "OldValue:\t\t{1}\r\n" +
+                                           "NewValue:\t\t{2}\r\n" +
+                                           "OrderUpdateTime:\t{3}\r\n" +
+                                           "====================";
```
halspang, line 20: Can you also make this a verbatim string?

examples/PubSub/DeclarativeSubscriber/Controllers/Order.cs
```diff
@@ -0,0 +1,33 @@
...
+        public async Task<OkResult> OnOrderUpdateAsync(PubSubEto<DemoOrderETO> meta)
+        {
+            return await Task.Run(() =>
+            {
+                Console.WriteLine(string.Format(reporter,meta.data.OrderId,meta.data.OldValue,meta.data.NewValue,meta.data.DateTime));
+                
+                //to tell Dapr message ACK.
+                return Ok();
+            });
+        }
```
halspang, line 31: I'd recommend returning an `ActionResult` or `IActionResult` as it's more generic (allows for failure status codes) and it keeps in line with other examples.

This feels a little forced as a Task/async function. Did you have a specific reason for making it this way?

examples/PubSub/ProgrammaticSubscriber/Controllers/Order.cs
```diff
@@ -0,0 +1,42 @@
...
+    [ApiController]
+    [Route("Order")]
+    public class Order : ControllerBase
+    {
+        private readonly string reporter = "====================\r\n" +
+                                           "OrderId:\t\t{0}\r\n" +
+                                           "OldValue:\t\t{1}\r\n" +
+                                           "NewValue:\t\t{2}\r\n" +
+                                           "OrderUpdateTime:\t{3}\r\n" +
+                                           "====================";
```
halspang, line 21: Could this also be a verbatim string?

examples/PubSub/ProgrammaticSubscriber/Controllers/Order.cs
```diff
@@ -0,0 +1,42 @@
...
+        public async Task<OkResult> OnOrderUpdateAsync(PubSubEto<DemoOrderETO> meta)
+        {
+            return await Task.Run(() =>
+            {
+                Console.WriteLine(string.Format(reporter,meta.data.OrderId,meta.data.OldValue,meta.data.NewValue,meta.data.DateTime));
+                
+                //to tell Dapr message ACK.
+                return Ok();
+            });
+        }
```
halspang, line 40: Same comment about needing to be async.

examples/PubSub/ProgrammaticSubscriberDotNet/Controllers/Order.cs
```diff
@@ -0,0 +1,40 @@
...
+        public async Task<OkResult> OnOrderUpdateAsync(DemoOrderETO data,[FromServices] DaprClient daprClient)
+        {
+            return await Task.Run(() =>
+            {
+                Console.WriteLine(string.Format(reporter,data.OrderId,data.Type,data.OldValue,data.NewValue,data.DateTime));
+                
+                //to tell Dapr message ACK.
+                return Ok();
+            });
+        }
```
halspang, line 38: Same comment for async.

examples/PubSub/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+```sh
+curl -X POST "http://localhost:5000/Order?newOwnerId=EE3845EB-B734-44D6-AB5A-1956A05B9E95" -d ''
+```
+
+On Windows:
+
+```sh
+curl -X POST "http://localhost:5000/Order?newOwnerId=EE3845EB-B734-44D6-AB5A-1956A05B9E95" -d ''
+```
```
halspang, line 78: Do you think we could also get an example of how to publish with the dapr cli just for completeness?

`dapr publish -p pubsub -t order.update -d ...`

halspang (commented):
Thanks for addressing the feedback! I have a few more questions.

1. Do you think we can combine any of projects? I'm thinking specifically the programmatic ones. They don't need to be 2 projects but rather 2 controllers in the same project. This will cut down on a lot of boiler plate but still makes sense organizationally to me.
2. I think the topics should be different for each method. When you have multiple subscribers to the same topic, you can get weird behavior depending on the underlying pubsub component.

examples/PubSub/DeclarativeSubscriber/Controllers/Order.cs
JhouXerox: nope, no specific reason.

It's ready to use await in the future.And It can be a general func.

JhouXerox:
> Thanks for addressing the feedback! I have a few more questions.
> 
> 1. Do you think we can combine any of projects? I'm thinking specifically the programmatic ones. They don't need to be 2 projects but rather 2 controllers in the same project. This will cut down on a lot of boiler plate but still makes sense organizationally to me.
> 2. I think the topics should be different for each method. When you have multiple subscribers to the same topic, you can get weird behavior depending on the underlying pubsub component.

About question 1:
I try to combine these as one project, and it created a new problem.
In the General way controller,I exposed an HTTPGet API routed to "/dapr/subscribe".
And in the dotnet way controller, I used TopicAttribute.

But it seems that the TopicAttribute is also exposed an API routed to "/dapr/subscribe".
these 2 controllers can only work with one sub behavior.
I had to comment the sub behavior code of one of them.
I can't find a nice way.

About question 2:
In this example,I let them be different.Maybe it should has a fully test in all pubsub brokers.

JhouXerox:
> Thanks for addressing the feedback! I have a few more questions.
> 
> 1. Do you think we can combine any of projects? I'm thinking specifically the programmatic ones. They don't need to be 2 projects but rather 2 controllers in the same project. This will cut down on a lot of boiler plate but still makes sense organizationally to me.
> 2. I think the topics should be different for each method. When you have multiple subscribers to the same topic, you can get weird behavior depending on the underlying pubsub component.

Yes, of cause. You can combine any projects.I've combined subscriber projects to 1. see the latest commit.

rynowak:
Is there an issue describing this? This is a really large PR and I'm not really sure what this is for. We are pretty intentional with accepting new sample content because it's something that have to document and maintain.

yash-nisar:
@JhouXerox Is there an update on this ?

JhouXerox:
nope, maybe this example is too old to compatibility with the latest version

PR #746, opened 2021-09-01
Adding match and priority to TopicAttribute for PubSub routing.

pkedy:
# Description

Add PubSub routing per dapr/dapr#3406.

Example usage

```C#
[Topic("pubsub", "transactions", "event.type == \"withdraw.v2\"", 1)]
[HttpPost("withdraw")]
public async Task<ActionResult> Withdraw(...)
```

## Issue reference

Part of dapr/dapr#2582

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/examples/AspNetCore/ControllerSample/CustomTopicAttribute.cs
+++ b/examples/AspNetCore/ControllerSample/CustomTopicAttribute.cs
@@ -18,7 +18,6 @@ namespace ControllerSample
         {
             this.PubsubName = Environment.ExpandEnvironmentVariables(pubsubName);
             this.Name = Environment.ExpandEnvironmentVariables(name);
-            this.Priority = int.MaxValue;
         }
 
         /// <inheritdoc/>
--- a/examples/AspNetCore/ControllerSample/CustomTopicAttribute.cs
+++ b/examples/AspNetCore/ControllerSample/CustomTopicAttribute.cs
@@ -18,12 +18,19 @@ namespace ControllerSample
         {
             this.PubsubName = Environment.ExpandEnvironmentVariables(pubsubName);
             this.Name = Environment.ExpandEnvironmentVariables(name);
+            this.Priority = int.MaxValue;
         }
+
+        /// <inheritdoc/>
+        public string PubsubName { get; }
 
         /// <inheritdoc/>
         public string Name { get; }
 
         /// <inheritdoc/>
-        public string PubsubName { get; }
+        public new string Match { get; }
+
+        /// <inheritdoc/>
+        public int Priority { get; }
     }
 }
--- a/src/Dapr.AspNetCore/DaprEndpointRouteBuilderExtensions.cs
+++ b/src/Dapr.AspNetCore/DaprEndpointRouteBuilderExtensions.cs
@@ -57,14 +57,16 @@ namespace Microsoft.AspNetCore.Builder
                 var dataSource = context.RequestServices.GetRequiredService<EndpointDataSource>();
                 var entries = dataSource.Endpoints
                     .OfType<RouteEndpoint>()
-                    .Where(e => e.Metadata.GetMetadata<ITopicMetadata>()?.Name != null)   // only endpoints which have  TopicAttribute with not null Name.
+                    .Where(e => e.Metadata.GetMetadata<ITopicMetadata>()?.Name != null)   // only endpoints which have TopicAttribute with not null Name.
                     .Distinct()
-                    .Select(e => (e.Metadata.GetMetadata<ITopicMetadata>().PubsubName, e.Metadata.GetMetadata<ITopicMetadata>().Name, e.Metadata.GetMetadata<IRawTopicMetadata>()?.EnableRawPayload, e.RoutePattern));
-
-
+                    .Select(e => (e.Metadata.GetMetadata<ITopicMetadata>().PubsubName, e.Metadata.GetMetadata<ITopicMetadata>().Name, e.Metadata.GetMetadata<IRawTopicMetadata>()?.EnableRawPayload, e.Metadata.GetMetadata<ITopicMetadata>().Match, e.Metadata.GetMetadata<ITopicMetadata>().Priority, e.RoutePattern))
+                    .GroupBy(e => new { e.PubsubName, e.Name })
+                    .Select(e => e.OrderBy(e => e.Priority))
+                    .SelectMany(e => e);
                 var logger = context.RequestServices.GetService<ILoggerFactory>().CreateLogger("DaprTopicSubscription");
 
                 List<Subscription> subscriptions = new();
+                Dictionary<string, Subscription> subscriptionMap = new();
                 foreach (var entry in entries)
                 {
                     // only return topics which have routes without parameters.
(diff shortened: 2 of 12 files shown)


src/Dapr.AspNetCore/DaprEndpointRouteBuilderExtensions.cs
```diff
@@ -84,12 +86,52 @@ private static IEndpointConventionBuilder CreateSubscribeEndPoint(IEndpointRoute
                         .Select(part => part.Content))));
 
                     var rawPayload = entry.EnableRawPayload ?? options?.EnableRawPayload;
-                    var subscription = new Subscription
+                    var subscriptionKey = entry.PubsubName + ":" + entry.Name;
```
rynowak, line 89: You can use a tuple for this, and it's a little more clear that it's a compound key. 

```
var subscriptionKey = (entry.PubsubName, entry.Name);
```

Make the dictionary a `Dictionary<(string, string), Subscription>`

src/Dapr.AspNetCore/DaprEndpointRouteBuilderExtensions.cs
```diff
@@ -84,12 +86,52 @@ private static IEndpointConventionBuilder CreateSubscribeEndPoint(IEndpointRoute
...
+                    Subscription subscription;
+                    try
                     {
-                        Topic = entry.Name,
-                        PubsubName = entry.PubsubName,
-                        Route = route
-                    };
+                        subscription = subscriptionMap[subscriptionKey];
+                    } catch (KeyNotFoundException) {
```
rynowak, line 94: Use the pattern:

```C#
if (subscriptionMap.TryGetValue(subscriptionKey, out var subscription))
{
    subscription = new Subscription(...);
    subscriptionMap.Add(subscriptionKey, subscription);
}
```

src/Dapr.AspNetCore/DaprEndpointRouteBuilderExtensions.cs
```diff
@@ -57,14 +57,16 @@ private static IEndpointConventionBuilder CreateSubscribeEndPoint(IEndpointRoute
...
-
-
+                    .Select(e => (e.Metadata.GetMetadata<ITopicMetadata>().PubsubName, e.Metadata.GetMetadata<ITopicMetadata>().Name, e.Metadata.GetMetadata<IRawTopicMetadata>()?.EnableRawPayload, e.Metadata.GetMetadata<ITopicMetadata>().Match, e.Metadata.GetMetadata<ITopicMetadata>().Priority, e.RoutePattern))
+                    .GroupBy(e => new { e.PubsubName, e.Name })
+                    .Select(e => e.OrderBy(e => e.Priority))
+                    .SelectMany(e => e);
                 var logger = context.RequestServices.GetService<ILoggerFactory>().CreateLogger("DaprTopicSubscription");
 
                 List<Subscription> subscriptions = new();
+                Dictionary<string, Subscription> subscriptionMap = new();
```
rynowak, line 69: We shouldn't need both a list and dictionary. Just use a dictionary and you can access the values using `subscriptionMap.Values.OrderBy( ... )`. 

You want the explicit ordering for testability since dictionary does not guarantee iteration order.

src/Dapr.AspNetCore/ITopicMetadata.cs
```diff
@@ -19,5 +19,15 @@ public interface ITopicMetadata
...
         /// </summary>
         string PubsubName { get; }
+
+        /// <summary>
+        /// The CEL expression to use to match events for this handler.
+        /// </summary>
+        string Match { get; }
+
+        /// <summary>
+        /// The priority in which this rule should be evaluated (lower to higher).
```
rynowak, line 29: Is lower better?

examples/AspNetCore/ControllerSample/CustomTopicAttribute.cs
```diff
@@ -18,12 +18,19 @@ public CustomTopicAttribute(string pubsubName, string name)
         {
             this.PubsubName = Environment.ExpandEnvironmentVariables(pubsubName);
             this.Name = Environment.ExpandEnvironmentVariables(name);
+            this.Priority = int.MaxValue;
```
rynowak, line 21: Suggestion - just let this be zero by default. This lets people increase and decrease priority relative to default. It's also a little less surprising that an int field has a default value of zero.

src/Dapr.AspNetCore/DaprEndpointRouteBuilderExtensions.cs
```diff
@@ -84,12 +86,52 @@ private static IEndpointConventionBuilder CreateSubscribeEndPoint(IEndpointRoute
...
+                            PubsubName = entry.PubsubName,
+                        };
+                        subscriptions.Add(subscription);
+                        subscriptionMap[subscriptionKey] = subscription;
+                    }
+
+                    if (string.IsNullOrEmpty(entry.Match))
+                    {
+                        if (subscription.Routes != null &&
+                            subscription.Routes.Rules.Count > 0)
```
rynowak, line 107: Can simplify to `subscription.Routes.Rules?.Count > 0`

src/Dapr.AspNetCore/DaprEndpointRouteBuilderExtensions.cs
```diff
@@ -84,12 +86,52 @@ private static IEndpointConventionBuilder CreateSubscribeEndPoint(IEndpointRoute
...
+                        {
+                            subscription.Routes.Default = route;
```
rynowak, line 109: What happens if a subscription has **multiple routes** with no rules? it seems like this is an authoring mistake we should report as an error.

src/Dapr.AspNetCore/DaprEndpointRouteBuilderExtensions.cs
```diff
@@ -84,12 +86,52 @@ private static IEndpointConventionBuilder CreateSubscribeEndPoint(IEndpointRoute
...
+                        } else
+                        {
+                            subscription.Route = route;
+                        }
+                    } else
```
rynowak, line 114: Fix formatting on this (you can just autoformat in VS). if/else and the braces go on their own line.

src/Dapr.AspNetCore/DaprEndpointRouteBuilderExtensions.cs
```diff
@@ -84,12 +86,52 @@ private static IEndpointConventionBuilder CreateSubscribeEndPoint(IEndpointRoute
...
+                    {
+                        if (subscription.Routes == null)
+                        {
+                            subscription.Routes = new Routes
+                            {
+                                Rules = new(),
+                            };
+                        }
```
rynowak, line 122: Can simplify this block to `subscription.Routes ?? = new();`

src/Dapr.AspNetCore/DaprEndpointRouteBuilderExtensions.cs
```diff
@@ -84,12 +86,52 @@ private static IEndpointConventionBuilder CreateSubscribeEndPoint(IEndpointRoute
...
+
+                    if (string.IsNullOrEmpty(entry.Match))
+                    {
+                        if (subscription.Routes != null &&
+                            subscription.Routes.Rules.Count > 0)
+                        {
+                            subscription.Routes.Default = route;
+                        } else
+                        {
+                            subscription.Route = route;
```
rynowak, line 112: It seems like the logic here depends on a lot on what order we see the attributes. Would it simplify things to remove the `.SelectMany` on line 65 and process a group at a time. 

I'm particularly passionate about doing a good job reporting invalid configurations. Users frequently ask for help troubleshooting pub/sub configurations.

src/Dapr.AspNetCore/TopicAttribute.cs
```diff
@@ -24,6 +24,7 @@ public TopicAttribute(string pubsubName, string name)
 
             this.Name = name;
             this.PubsubName = pubsubName;
+            this.Priority = int.MaxValue;
```
rynowak, line 27: I would prefer this to be initialized to zero unless there's a reason why there can't be. ASP.NET core's use of priorities/orders works this way.

rynowak: - lower is better
- zero is the default

src/Dapr.AspNetCore/TopicAttribute.cs
```diff
@@ -50,5 +90,11 @@ public TopicAttribute(string pubsubName, string name, bool enableRawPayload)
 
         /// <inheritdoc/>
         public bool? EnableRawPayload { get; }
+
+        /// <inheritdoc/>
+        public new string Match { get; }
```
rynowak, line 95: Is there a reason why this is `new`? If not then take it out :)

rynowak: Ah, I see it collides with a method on `Attribute. Carry on then!

src/Dapr.AspNetCore/TopicAttribute.cs
pkedy: I have it initialize to MaxValue so that it comes after the routed attributes in the OrberBy I do in DaprEndpointRouteBuilderExtensions.cs, but the way the code is written now, it doesn't really matter that much. If it gets ordered after routed attributes, then it will be assigned to `Routes.Default` instead of `Route` (which then has be be moved to `Routes.Default` when a routed attribute is encountered. TLDR is a tiny optimization.

fbridger:
@rynowak @pkedy I believe this PR introduced a Regression related to multiple endpoint subscribing to the same Topic.

Please check this issue: #752

Issue #728, opened 2021-07-29
Expose HttpClient on ActorProxy


clintsinger:
The `ActorProxy` doesn't allow access to the `HttpClient`.   This is useful so that one would be able to add additional configuration such as being able to define the `HttpClient.Timeout`.   

As an example use case, being able to set the `Timeout` to `InfiniteTimespan`  when doing service debugging would help in scenarios when stepping through the code takes longer than the default timeout.

doddgu:
I agree, I think it could be more open.  For example, response processing.
The scene is I want throw user friendly exception in actor, and I don't want to cover it with error code and error message in response.

client
```csharp
var balance = await bank.GetAccountBalance();
```
client side global exception catch
```csharp
ActorProxy.ResponseProcessing = (resp, e) =>
{
     if (resp.StatusCode == 299)
    {
        // to do something
        e.Handle = true;  // dapr actor 
    }
};
```

actor method, global exception catch will catch the user friendly exception and set status code = 299, response body is "Just Test".
```csharp
public async Task<AccountBalance> GetAccountBalance()
{
    throw new UserFriendlyException("Just Test");
}
```
----
Now, my solution is
Common actor response
```csharp
public class ActorResponse<T> where T : class
{
    public int ErrorCode { get; set; }
    public string ErrorMessage { get; set; }
    public T Result { get; set; }
}
```
client
```csharp
var balanceResponse = await bank.GetAccountBalance();
if (balanceResponse.ErrorCode == 200)
{
    // to do something
}
else if(balanceResponse.ErrorCode == 299)
{
    throw new UserFriendlyException(balanceResponse.ErrorMessage);
}
else
{
    throw new Exception(balanceResponse.ErrorMessage);
}
```
actor method
```csharp
public async Task<ActorResponse<AccountBalance>> GetAccountBalance()
{
    return new ActorResponse()
    {
        ErrorCode = 299,
        ErrorMessage = "JustTest",
    }
}
```
----
It's too complicated. I think c# could be a little more elegant.

PR #748, opened 2021-09-02
Add an option to set Timeout for ActorProxy

halspang:
# Description
Actors, by there nature, may have to wait for a long period before
being allowed to execute. This could lead to long request times. The
default timeout for the HttpClient being used is fairly low. This
commit allows for the timeout to be set when constructing the
proxy.

## Issue reference

Please reference the issue this PR will close: #728

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


src/Dapr.Actors/DaprHttpInteractor.cs
```diff
@@ -44,6 +44,15 @@ internal class DaprHttpInteractor : IDaprInteractor
...
         }
 
+        public DaprHttpInteractor(
+            HttpMessageHandler clientHandler,
+            string httpEndpoint,
+            string apiToken,
+            TimeSpan? requestTimeout) : this(clientHandler, httpEndpoint, apiToken)
+        {
+            this.httpClient.Timeout = requestTimeout ?? this.httpClient.Timeout;
+        }
```
rynowak, line 54: Does this need to be its own constructor? The existing one is internal.

halspang: Removed it. Didn't think about the fact that since it's internal it can't be referenced and thus won't break anyone's code when I changed the signature.

rynowak (approved):

PR #758, opened 2021-10-01
Improve Actor sample

RicardoNiepel:
# Description

- using die ActorClient also as Dapr application
- adding sample.http to Actor sample

## Issue reference

No - only small changes

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/examples/Actor/README.md
+++ b/examples/Actor/README.md
@@ -33,14 +33,8 @@ The `ActorClient` project shows how to make client calls for actor using Remotin
 Run the client project from `ActorClient` directory as:
 
 ```sh
- dotnet run
- ```
-
- *Note: If you started the actor service with dapr port other than 3500, then set the environment variable DAPR_HTTP_PORT to the value of --dapr-http-port specified while starting the actor service before running the client in terminal.*
- ```
- On Windows: set DAPR_HTTP_PORT=<port>
- On Linux, MacOS: export DAPR_HTTP_PORT=<port>
- ```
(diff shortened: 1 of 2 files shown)


examples/Actor/README.md
```diff
@@ -33,14 +33,8 @@ The `ActorClient` project shows how to make client calls for actor using Remotin
...
 ```sh
- dotnet run
- ```
-
- *Note: If you started the actor service with dapr port other than 3500, then set the environment variable DAPR_HTTP_PORT to the value of --dapr-http-port specified while starting the actor service before running the client in terminal.*
- ```
- On Windows: set DAPR_HTTP_PORT=<port>
- On Linux, MacOS: export DAPR_HTTP_PORT=<port>
- ```
+ dapr run --app-id actor_client dotnet run
```
halspang, line 36: What's the reason for doing this? You don't need to run this app with a dapr sidecar to invoke the actors. I think showing that you can interact with actors from outside of dapr is an important point.

RicardoNiepel: Then both options should be added to the docs to explain the two ways.

In real scenarios, even if you are communicating with actors from the outside - this "outside" application will not run in the actors cluster - thus expecting dapr sidecar of the actors app is running on the same machine will not be possible.

WhitWaldo:
Thank you for your contribution! Given the age of this PR, I'm going to go ahead and close it, but please don't hesitate to open any other PRs against this project.

Issue #715, opened 2021-07-12
Unable to subscribe an enpoint to multiple pub-subs


mmisztal1980:
## Expected Behavior

Having registered multiple different pubsubs (redis streams & rabbit) to an endpoint, I expected that the endpoint would subscribe to all of them. This was interesting to me because I was testing if I could have HA on the pubsub(s) by having multiple instances of it (for instance 2x separate rabbitmq instances)

## Actual Behavior

When using `.WithTopic` multiple times, it appears that the last registered pubsub wins.

I've tested it by :
- commenting out the 2nd pubsub, to verify that the 1st pubsub works as intended
- inverting the order in which pubsubs were registered to verify that "last one wins" theory - the test confirmed it.

## Steps to Reproduce the Problem

Sample C# code to reproduce the problem:

```csharp
endpoints
   .MapPost("/orders", async context =>
  {
    var order = await JsonSerializer.DeserializeAsync<Order>(context.Request.Body);
    this.Service.Logger.LogInformation($"Order {order.Id} received");                        
  })
  .WithTopic(Constants.PubSub.RedisStreams, Constants.OrdersTopicName)
  .WithTopic(Constants.PubSub.RabbitMq, Constants.OrdersTopicName);
```

## Release Note

RELEASE NOTE:
Fixed an issue which prevented registering multiple pubsub subscriptions on to an endpoint.

rynowak:
@halspang - can you look into this?

PR #763, opened 2021-10-08
Allow multiple topics to call the same endpoint

halspang:
# Description

Using the WithTopic builder pattern allows you to specify multiple
topics that route to a single endpoint. Given the topics are unique,
this should be a valid case.

This change allows for a single endpoint to bind to multiple topics.
Note that this can only be done via the EndpointBuilder and not the
TopicAttribute.

## Issue reference

Please reference the issue this PR will close: #715 
## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


rynowak:
Can you update this to add an `[AttributeUsage(...)]` to `[Topic(...)]` - we need to be able to set AllowMultiple. And we should add a test where multiple topic attributes are applied.

Issue #777, opened 2021-11-09
add support for statestore query API


dmitsh:
This feature is implemented in dapr runtime: dapr/dapr#3662

It should be available via the SDK

PR #810, opened 2022-01-06
Add support for Query API

halspang:
# Description

This commit adds methods to the DaprClient to call the new Query API.
Currently, a raw query string is provided to the API. The API is
still in the alpha stage which may cause shifts in how the API works.
This should not effect how the SDK exposes the call, but may change
the underlying semantics.

https://github.com/dapr/dotnet-sdk/issues/777

## Issue reference

Please reference the issue this PR will close: #777 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


src/Dapr.Client/DaprClient.cs
```diff
@@ -756,6 +756,44 @@ public async Task<StateEntry<TValue>> GetStateEntryAsync<TValue>(string storeNam
...
+        /// <param name="storeName">The name of the statestore.</param>
+        /// <param name="jsonQuery">A JSON formatted query string.</param>
+        /// <param name="metadata">Metadata to send to the statestore.</param>
+        /// <param name="cancellationToken">A <see cref="CancellationToken"/> that can be used to cancel the operation.</param>
+        /// <typeparam name="TValue">The data type of the value to read.</typeparam>
+        /// <returns>A <see cref="StateQueryResponse{TValue}"/> that may be paginated, use <see cref="StateQueryResponse{TValue}.Token"/> to continue the query.</returns>
+        public virtual async Task<StateQueryResponse<TValue>> QueryStateAsync<TValue>(
+            string storeName,
+            string jsonQuery,
+            IReadOnlyDictionary<string, string> metadata,
```
rynowak, line 784: Should metadata be optional?

src/Dapr.Client/DaprClient.cs
```diff
@@ -756,6 +756,44 @@ public async Task<StateEntry<TValue>> GetStateEntryAsync<TValue>(string storeNam
...
+        /// <summary>
+        /// Queries a statestore with the given request. The request contains a JSON representation of the query as described by the Dapr QueryState API.
+        /// Note that the underlying statestore must support queries.
+        /// </summary>
+        /// <param name="request">A <see cref="StateQueryRequest"/></param>
+        /// <param name="cancellationToken">A <see cref="CancellationToken"/> that can be used to cancel the operation.</param>
+        /// <typeparam name="TValue">The data type of the value to read.</typeparam>
+        /// <returns>A <see cref="StateQueryResponse{TValue}"/> that may be paginated, use <see cref="StateQueryResponse{TValue}.Token"/> to continue the query.</returns>
+        public abstract Task<StateQueryResponse<TValue>> QueryStateAsync<TValue>(
+            StateQueryRequest request,
```
rynowak, line 768: Should metadata be its own parameter here?

rynowak: After looking at this in more detail, it seems best to just get rid of this overload and have a single one. It's just 3 parameters and one of them is optional anyway.

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -1112,6 +1162,7 @@ private CallOptions CreateCallOptions(Metadata headers, CancellationToken cancel
             {
                 options.Headers.Add(this.apiTokenHeader.Value.Key, this.apiTokenHeader.Value.Value);
             }
+            options.Headers.Add("Content-Type", "application/x-www-form-urlencoded");
```
rynowak, line 1165: This seems like an oops.

src/Dapr.Client/StateQueryRequest.cs
```diff
@@ -0,0 +1,38 @@
...
+*/
+
+using System.Collections.Generic;
+
+namespace Dapr.Client
+{
+    /// <summary>
+    /// Represents a query request to a statestore.
+    /// </summary>
+    public class StateQueryRequest
```
rynowak, line 21: As discussed lets remove this.

src/Dapr.Client/StateQueryResponse.cs
```diff
@@ -0,0 +1,64 @@
...
+*/
+
+using System.Collections.Generic;
+
+namespace Dapr.Client
+{
+    /// <summary>
+    /// Represents the response from a state query.
+    /// </summary>
+    public class StateQueryResponse<TValue>
```
rynowak, line 21: Can we make StateQueryResponse and related types immutable?

src/Dapr.Client/StateQueryResponse.cs
```diff
@@ -0,0 +1,64 @@
...
+    {
+        /// <summary>
+        /// The results of the query.
+        /// </summary>
+        public IReadOnlyList<StateQueryItem<TValue>> Results { get; set; }
+
+        /// <summary>
+        /// The pagination token to continue the query.
+        /// </summary>
+        public string Token { get; set; }
```
rynowak, line 31: The API in this PR defines no way to send this back as part of a request. How does this work?

src/Dapr.Client/StateQueryResponse.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        /// <summary>
+        /// The ETag for the key from the matched query.
+        /// </summary>
+        public string ETag { get; set; }
+
+        /// <summary>
+        /// The error from the query, if one occurred.
+        /// </summary>
+        public string Error { get; set; }
```
rynowak, line 62: What do users do with this?

rynowak: This seems like a bonkers API design no? Would selected items ever fail to decrypt or would everything fail uniformly.

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -1015,6 +1015,56 @@ private Autogenerated.StateItem ToAutogeneratedStateItem(StateTransactionRequest
...
+            try
+            {
+                var queryResponse = await client.QueryStateAlpha1Async(queryRequest, options);
+                var items = queryResponse.Results.Select(res =>
+                    new StateQueryItem<TValue>()
+                    {
+                        Key = res.Key,
+                        Data = TypeConverters.FromJsonByteString<TValue>(res.Data, this.JsonSerializerOptions),
+                        ETag = res.Etag,
+                        Error = res.Error,
```
rynowak, line 1047: The fact that the error property exists means that we can't **just** try to deserialize every item, we have to check for an error first.

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -1015,6 +1015,56 @@ private Autogenerated.StateItem ToAutogeneratedStateItem(StateTransactionRequest
                 throw new DaprException("State operation failed: the Dapr endpoint indicated a failure. See InnerException for details.", ex);
             }
         }
+
+        /// <inheritdoc/>
+        public async override Task<StateQueryResponse<TValue>> QueryStateAsync<TValue>(StateQueryRequest request, CancellationToken cancellationToken = default)
```
rynowak, line 1020: Let's think about failure modes. If an item fails to decrypt what would a user want to do with partial data? Is there anything useful to be done?

Some options:

- Throw if any items failed and no way to get partial data
- Throw if any items failed and include partial data as part of the exception
- Allow the choice as part of the API (flag)
- Current option: return the data and don't throw on decryption error

test/Dapr.Client.Test/StateApiTest.cs
```diff
@@ -999,6 +999,49 @@ public async Task DeleteBulkStateAsync_ValidateRequest()
...
+            {
+                StoreName= "testStore",
+                Query = queryJson,
+            };
+            var request = await client.CaptureGrpcRequestAsync(async daprClient =>
+            {
+                await daprClient.QueryStateAsync<string>(stateRequest);
+            });
+
+            request.Dismiss();
```
rynowak, line 1018: See feedback on configuration PR - we can test more of the surface area by including a response.

src/Dapr.Client/DaprClientGrpc.cs
halspang: Sure was, thanks for catching it!

src/Dapr.Client/StateQueryResponse.cs
halspang: You send it back in the query itself actually. It's part of the `pagination` section. I'm not sure how to relay that in the SDK without just linking to the docs. Or is this where we should put in a simple example?

https://docs.dapr.io/developing-applications/building-blocks/state-management/howto-state-query-api/#example-3

src/Dapr.Client/StateQueryResponse.cs
halspang: As we discussed, everything decrypts individually,  hence the individual errors.

src/Dapr.Client/DaprClientGrpc.cs
halspang: Yeah I'll add that check in as part of the solution we discussed below.

src/Dapr.Client/DaprClientGrpc.cs
halspang: I think #2 is a good option. I'll implement that.

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -1015,6 +1015,57 @@ private Autogenerated.StateItem ToAutogeneratedStateItem(StateTransactionRequest
...
+            {
+                var items = new List<StateQueryItem<TValue>>();
+                var queryResponse = await client.QueryStateAlpha1Async(queryRequest, options);
+                foreach (var item in queryResponse.Results)
+                {
+                    if (!string.IsNullOrEmpty(item.Error))
+                    {
+                        // When we encounter an error, we throw an exception containing the results thus far.
+                        var partialResults = new StateQueryResponse<TValue>(items, string.Empty, queryResponse.Metadata);
+                        throw new StateQueryException<TValue>($"Encountered an error while processing state query results: {item.Error}", partialResults);
```
rynowak, line 1052: You should do this in two passes. The problem with this is you get different data depending on order of the documents returned.

So first read the results and build `items`. Then loop again and check for errors. That way you always include all of the data.

src/Dapr.Client/DaprClientGrpc.cs
rynowak: Can you remove this for real?

src/Dapr.Client/DaprClientGrpc.cs
halspang: Done. I also included a list of the keys that failed so it's easier for a user to identify where things are going wrong/deal with the problematic keys.

src/Dapr.Client/DaprClientGrpc.cs
rynowak: Can you remove this for real?

src/Dapr.Client/StateQueryResponse.cs
rynowak: Docs are fine, I it's a little unfortunate building queries as strings **and** building them with interpolation.

rynowak (approved):

Issue #787, opened 2021-11-23
add support for Configuration API


dmitsh:
The Configuration API feature has been implemented in Dapr runtime: https://github.com/dapr/dapr/pull/3606
It needs to be accessible via the SDK.

PR #812, opened 2022-01-11
Add support for the GetConfiguration API

halspang:
# Description

This commit adds support for the GetConfiguration API. This is an
alpha API and may be subject to change. This commit also does not
cover the Subscribe API for Configurations.

Included is a simple example showing the usage of the Configuration
API.

## Issue reference

Please reference the issue this PR will close: #787 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


examples/Client/ConfigurationApi/ConfigurationApi.csproj
```diff
@@ -0,0 +1,19 @@
...
+    <ProjectReference Include="..\..\..\src\Dapr.Client\Dapr.Client.csproj">
+      <GlobalPropertiesToRemove></GlobalPropertiesToRemove>
+    </ProjectReference>
+  </ItemGroup>
+  <ItemGroup>
+    <None Remove="Components\" />
+  </ItemGroup>
+  <ItemGroup>
+    <Folder Include="Components\" />
+  </ItemGroup>
```
rynowak, line 18: This can be simplified, we just need the project reference. You can safely remove `GlobalPropertiesToRemove`

src/Dapr.Client/ConfigurationItem.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System.Collections.Generic;
+
+namespace Dapr.Client
+{
+    /// <summary>
+    /// Class that represents an item fetched from the Dapr Configuration API.
+    /// </summary>
+    public class ConfigurationItem
+    {
+        /// <summary>
```
rynowak, line 10: Feedback: consider making this immutable after construction, to more clearly communicate intent. Not really a big decision.

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -1092,6 +1092,47 @@ private Autogenerated.StateItem ToAutogeneratedStateItem(StateTransactionRequest
...
 
+        #region Configuration API
+        /// <inheritdoc/>
+        public override Task<IReadOnlyList<ConfigurationItem>> GetConfiguration(
+            string storeName,
+            IReadOnlyList<string> keys,
+            IReadOnlyDictionary<string, string> metadata = default,
+            CancellationToken token = default)
+        {
+            var request = new Autogenerated.GetConfigurationRequest()
```
rynowak, line 1103: Add null check for `storeName`, would be good to tolerate `keys` being null since we said it's optional.

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -1092,6 +1092,47 @@ private Autogenerated.StateItem ToAutogeneratedStateItem(StateTransactionRequest
...
+            {
+                StoreName = storeName
+            };
+            request.Keys.AddRange(keys);
+
+            if (metadata != null)
+            {
+                foreach (var kvp in metadata)
+                {
+                    request.Metadata.Add(kvp.Key, kvp.Value);
```
rynowak, line 1113: Is request.metadata initialized by default? or no?

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -1092,6 +1092,47 @@ private Autogenerated.StateItem ToAutogeneratedStateItem(StateTransactionRequest
         }
         #endregion
 
+        #region Configuration API
+        /// <inheritdoc/>
+        public override Task<IReadOnlyList<ConfigurationItem>> GetConfiguration(
```
rynowak, line 1097: Suggestion: let's change this to `ConfigurationResponse` where it has a property with the items. This way if we add new data to the wire format, we can avoid a bronking change.

src/Dapr.Client/DaprClient.cs
```diff
@@ -782,6 +783,20 @@ public async Task<StateEntry<TValue>> GetStateEntryAsync<TValue>(string storeNam
...
 
+        /// <summary>
+        /// Get a list of configuration items based on keys from the given statestore. 
+        /// </summary>
+        /// <param name="storeName">The name of the statestore to be queried.</param>
+        /// <param name="keys">An optional list of keys to query for. If provided, the result will only contain those keys. An empty list indicates all keys should be fetched.</param>
+        /// <param name="metadata">Optional metadata that will be sent to the statestore being queried.</param>
+        /// <param name="cancellationToken">A <see cref="CancellationToken" /> that can be used to cancel the operation.</param>
+        /// <returns></returns>
+        public abstract Task<IReadOnlyList<ConfigurationItem>> GetConfiguration(
```
rynowak, line 794: let's figure out how to communciate to users taht this is experimental and may change in the future.

test/Dapr.Client.Test/ConfigurationApiTest.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System.Collections.Generic;
```
rynowak, line 1: needs copyright

test/Dapr.Client.Test/ConfigurationApiTest.cs
```diff
@@ -0,0 +1,38 @@
...
+            {
+                { "key1", "value1" },
+                { "key2", "value2" }
+            };
+            var request = await client.CaptureGrpcRequestAsync(async daprClient =>
+            {
+                return await daprClient.GetConfiguration("testStore", new List<string>() { "test_key" }, metadata);
+            });
+
+            request.Dismiss();
```
rynowak, line 25: Can you change this to return some data instead of doing nothing? Example: https://github.com/halspang/dotnet-sdk/blob/ae9f30e59cbd02f383f29c07af4ef41d96659941/test/Dapr.Client.Test/DaprClientTest.InvokeMethodGrpcAsync.cs#L123

Then we can validate the output as well.

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -1092,6 +1092,47 @@ private Autogenerated.StateItem ToAutogeneratedStateItem(StateTransactionRequest
...
+            string storeName,
+            IReadOnlyList<string> keys,
+            IReadOnlyDictionary<string, string> metadata = default,
+            CancellationToken token = default)
```
rynowak, line 1101: by convention this should be `cancellationToken`

src/Dapr.Client/ConfigurationItem.cs
halspang: I agree, done.

src/Dapr.Client/DaprClientGrpc.cs
halspang: Done.

src/Dapr.Client/DaprClientGrpc.cs
halspang: It appears to be. I have a test where we set it and it's OK with this. Other requests involving a map do the same.

src/Dapr.Client/DaprClient.cs
halspang: Using `Obsolete` as we discussed.

src/Dapr.Client/DaprClientGrpc.cs
halspang: Done.

src/Dapr.Client/ConfigurationItem.cs
```diff
@@ -0,0 +1,75 @@
...
+            this.key = key;
+            this.value = value;
+            this.version = version;
+            this.metadata = metadata;
+        }
+
+        /// <summary>
+        /// The name of the configuration item.
+        /// </summary>
+        public string Key
```
rynowak, line 33: By the way you want to simplify these can be autoprops. `public string Key { get; }`

rynowak (dismissed):

rynowak (approved):

Issue #814, opened 2022-01-17
ActorHost.JsonSerializerOptions is always null


mumumi:
## Expected Behavior

After configuring "JsonSerializerOptions" in "AddActors", im going to request the same options from "actorHost" in the constructor of my "TestActor".

## Actual Behavior

Got a null value instead.

## Steps to Reproduce the Problem

Simply check "actorHost.JsonSerializerOptions"

## Release Note

RELEASE NOTE:
**FIX** Missing JSON Serializer option in ActorHost

PR #815, opened 2022-01-17
Fix missing assignment

mumumi:
# Description

The property "JsonSerializerOptions" in the class "ActorHost" is always null due to missing assignment in the constructor.

## Issue reference

Please reference the issue this PR will close: #814

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


halspang (changes requested):
Change seems fine. I did look through the usages and it doesn't seem like we ever use this outside of testing.

Looks like you need to get your commits sorted out though. You've pulled in two other commits on master.

Issue #835, opened 2022-02-11
App API Token support


gdlcf88:
https://docs.dapr.io/operations/security/app-api-token/

Hi, can we support App API Token in .NET SDK? :)

https://github.com/dapr/go-sdk/blob/40d8a4eabe2bc6244a0de6789f74d2fe6e302db9/service/http/service.go#L53
https://github.com/dapr/go-sdk/blob/40d8a4eabe2bc6244a0de6789f74d2fe6e302db9/service/http/invoke.go#L40-L46

JadynWong:
Sdk contains authentication processing for requests from Dapr.
https://github.com/dapr/dotnet-sdk/blob/master/src/Dapr.AspNetCore/DaprAuthenticationHandler.cs

According to the documentation, `DaprAuthenticationOptions.Token` should read the value of the `APP_API_TOKEN` environment variable, but currently it reads the value of the `DAPR_API_TOKEN` environment variable.

Go sdk and php sdk both do this.
https://github.com/dapr/php-sdk/blob/f0d66c38fbe7ade9b7f51a432fa1f7ed1f015bf8/src/lib/Middleware/Defaults/TokenAuth.php#L20

https://github.com/dapr/go-sdk/blob/40d8a4eabe2bc6244a0de6789f74d2fe6e302db9/service/http/service.go#L53

PR #836, opened 2022-02-12
Fix App API token authentication

JadynWong:
# Description

Fix App API Token support

docs: https://docs.dapr.io/operations/security/app-api-token/

According to the documentation, DaprAuthenticationOptions.Token should read the value of the APP_API_TOKEN environment variable, but currently it reads the value of the DAPR_API_TOKEN environment variable.

## Issue reference

#835

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/src/Dapr.AspNetCore/DaprAuthenticationBuilderExtensions.cs
+++ b/src/Dapr.AspNetCore/DaprAuthenticationBuilderExtensions.cs
@@ -1,4 +1,4 @@
-// ------------------------------------------------------------------------
+﻿// ------------------------------------------------------------------------
 // Copyright 2021 The Dapr Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -23,22 +23,22 @@ namespace Microsoft.AspNetCore.Authentication
     public static class DaprAuthenticationBuilderExtensions
     {
         /// <summary>
-        /// Adds Dapr API token authentication.
-        /// See https://docs.dapr.io/operations/security/api-token/ for more information about API token authentication in Dapr.
-        /// By default, the token will be read from the DAPR_API_TOKEN environment variable.
+        /// Adds Dapr App API token authentication.
+        /// See https://docs.dapr.io/operations/security/app-api-token/ for more information about App API token authentication in Dapr.
+        /// By default, the token will be read from the APP_API_TOKEN environment variable.
         /// </summary>
         /// <param name="builder">The <see cref="AuthenticationBuilder"/>.</param>
         /// <returns>A reference to <paramref name="builder"/> after the operation has completed.</returns>
         public static AuthenticationBuilder AddDapr(this AuthenticationBuilder builder) => builder.AddDapr(configureOptions: null);
 
         /// <summary>
-        /// Adds Dapr API token authentication.
-        /// See https://docs.dapr.io/operations/security/api-token/ for more information about API token authentication in Dapr.
+        /// Adds Dapr App API token authentication.
+        /// See https://docs.dapr.io/operations/security/app-api-token/ for more information about App API token authentication in Dapr.
         /// </summary>
         /// <param name="builder">The <see cref="AuthenticationBuilder"/>.</param>
         /// <param name="configureOptions">
         /// A delegate that allows configuring <see cref="DaprAuthenticationOptions"/>.
-        /// By default, the token will be read from the DAPR_API_TOKEN environment variable.
+        /// By default, the token will be read from the APP_API_TOKEN environment variable.
         /// </param>
         /// <returns>A reference to <paramref name="builder"/> after the operation has completed.</returns>
         public static AuthenticationBuilder AddDapr(this AuthenticationBuilder builder, Action<DaprAuthenticationOptions> configureOptions)
--- a/src/Dapr.AspNetCore/DaprAuthenticationBuilderExtensions.cs
+++ b/src/Dapr.AspNetCore/DaprAuthenticationBuilderExtensions.cs
@@ -23,7 +23,7 @@ namespace Microsoft.AspNetCore.Authentication
     public static class DaprAuthenticationBuilderExtensions
     {
         /// <summary>
-        /// Adds Dapr App API token authentication.
+        /// Adds App API token authentication.
         /// See https://docs.dapr.io/operations/security/app-api-token/ for more information about App API token authentication in Dapr.
         /// By default, the token will be read from the APP_API_TOKEN environment variable.
         /// </summary>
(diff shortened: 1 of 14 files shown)


halspang:
@JadynWong - Thanks for submitting this! I think it's important to be careful around this kind of code though. Both `APP_API_TOKEN` and `DAPR_API_TOKEN` are valid and signify _different_ things. Changing the default token could break existing applications.

https://docs.dapr.io/operations/security/app-api-token/#authenticating-requests-from-dapr
https://github.com/dapr/dapr/blob/master/pkg/runtime/security/token.go#L9

JadynWong:
Hi @halspang,

You are right, it is important. `APP_API_TOKEN` and `DAPR_API_TOKEN` indicate different things.

`DAPR_API_TOKEN`: The value of the header `dapr-api-token` sent by the application to dapr.
`APP_API_TOKEN`: The value of header `dapr-api-token` sent by dapr to the application.

Both [go sdk](https://github.com/dapr/go-sdk/blob/40d8a4eabe2bc6244a0de6789f74d2fe6e302db9/service/http/service.go#L53) and [php sdk](https://github.com/dapr/php-sdk/blob/f0d66c38fbe7ade9b7f51a432fa1f7ed1f015bf8/src/lib/Middleware/Defaults/TokenAuth.php#L20) use `APP_API_TOKEN` to validate requests from dapr, I don't quite understand why dotnet-sdk uses `DAPR_API_TOKEN` to validate requests from dapr? Is this a bug or by design?

halspang:
@rynowak - Do you have any context on this?

msfussell:
Request from Dapr should be authenticated with APP_API_TOKEN. If it is using DAPR_API_TOKEN this is a bug that need to be fixed. 

What about requests to Dapr from the App, do these not use DAPR_API_TOKEN?

halspang:
@msfussell - It uses `DAPR_API_TOKEN` for both outgoing and incoming requests. 

Looks like the incoming request was added somewhat recently: https://github.com/dapr/dotnet-sdk/pull/688

This seems like it'll be worth adding into 1.8. We do need to support both tokens though, so this PR still requires changes. It'll also need to change the handler/tests. Can you add that to this PR @JadynWong?

halspang:
@JadynWong - Yes, the incoming request is using the wrong environment variable. However, outgoing requests need to use the `DAPR_API_TOKEN` environment variable, which is what they are currently using. 

So, this PR needs to add the additional field for the `APP_API_TOKEN` and use that for incoming requests instead of just replacing the existing implementation with the new variable.

@msfussell - That's the correct interpretation, right?

Incoming from Dapr: `APP_API_TOKEN`
Outgoing to Dapr: `DAPR_API_TOKEN`

src/Dapr.AspNetCore/DaprAuthenticationBuilderExtensions.cs
```diff
@@ -23,22 +23,22 @@ namespace Microsoft.AspNetCore.Authentication
     public static class DaprAuthenticationBuilderExtensions
     {
         /// <summary>
-        /// Adds Dapr API token authentication.
-        /// See https://docs.dapr.io/operations/security/api-token/ for more information about API token authentication in Dapr.
-        /// By default, the token will be read from the DAPR_API_TOKEN environment variable.
+        /// Adds Dapr App API token authentication.
```
gdlcf88, line 26: I think this can be:

Adds <del>Dapr</del> App API token authentication.

src/Shared/DaprDefaults.cs
```diff
@@ -20,6 +20,7 @@ internal static class DaprDefaults
         private static string httpEndpoint;
         private static string grpcEndpoint;
         private static string apiToken;
```
gdlcf88, line 22: Better rename to `daprApiToken`.

gdlcf88:
@halspang - I think @JadynWong has finished the change.

1. Created a new property named `appApiToken` and renamed `apiToken` to `daprApiToken`:
https://github.com/dapr/dotnet-sdk/pull/836/files#diff-ad21b6521d67ae97593539892631e47f6bd72bdd89fad9ad379ac1608616d3c7R22-R23

2. Tested the new API token:
https://github.com/dapr/dotnet-sdk/pull/836/files#diff-288ea7196a5c8ff56e7b3042a732d225be4693c82baff78ea5e9e714c8ad3f76R44

Outgoing requests are still using the Dapr token:
https://github.com/dapr/dotnet-sdk/pull/836/files#diff-b1abda64bd1598bcb9d7f0dee97b3330a388352a5f018f1302fc7f948ee0a8f7R43

src/Shared/DaprDefaults.cs
```diff
@@ -19,23 +19,33 @@ internal static class DaprDefaults
     {
         private static string httpEndpoint;
         private static string grpcEndpoint;
-        private static string apiToken;
+        private static string daprApiToken;
+        private static string appApiToken;
 
-        public static string GetDefaultApiToken()
+        public static string GetDefaultDaprApiToken()
```
halspang, line 25: If you want to do this, I'd recommend keeping the old method and having it point to this. With this change, you'll break anyone who references this method at compile time, which isn't a good experience.

halspang (changes requested):
Thanks for working on this! Quick comment about changing the method name.

msfussell:
> @msfussell - That's the correct interpretation, right?
> 
> Incoming from Dapr: `APP_API_TOKEN` Outgoing to Dapr: `DAPR_API_TOKEN`
Yes, this is correct. To check the logic, if the token is not present in the header it should succeed, and only fail if the token is present but does not match. With the SDKs that support this (like .NET) it is good practice to set both of these environment variables to provide authenticated communication. 
If you are allowed to reference docs from the code, then these URLs would be good to add in the comments
- https://docs.dapr.io/operations/security/api-token/
- https://docs.dapr.io/operations/security/app-api-token/

src/Shared/DaprDefaults.cs
gdlcf88: Yep, shall we keep the old method with
```c#
[Obsolete("Use GetDefaultDaprApiToken method.")]
```
?

Since this attribute can also remind APP developers who use the Dapr API token incorrectly.

gdlcf88:
This PR has brought a breaking change, even if we don't change the name of the `GetDefaultApiToken` method since the token value for incoming requests authentication has already actually changed.

halspang:
> This PR has brought a breaking change, even if we don't change the name of the `GetDefaultApiToken` method since the token value for incoming requests authentication has already actually changed.

Yes, it'll be marked as a breaking change no matter what :) I'm just trying to limit the breaking change to one that fixes the bug instead of extra.

src/Shared/DaprDefaults.cs
halspang: @JadynWong - I just spoke with @rynowak who explained to me that since this is an internal class, removing the method won't actually break anyone. Sorry to make you redo work that I called out but you're free to delete this method and use the better named one.

halspang:
@JadynWong - One last comment and I think we're good here. It's having you delete that method again. I misunderstood how the breaking changes worked with an internal class but have since gotten clarity.

Thanks again for all your work here!

src/Shared/DaprDefaults.cs
JadynWong: I have reverted it and add comments.

halspang (approved):

Issue #779, opened 2021-11-11
AddDaprSecretStore Extension always throws unhandled exception at container start


bgelens:
## Expected Behavior

Using `AddDaprSecretStore` extension will not throw an exception when the side car is not ready yet

## Actual Behavior

It throws an unhandled exception crashing my pod. On restart, it goes through fine

```
Unhandled exception. Dapr.DaprException: Bulk secret operation failed: the Dapr endpoint indicated a failure. See InnerException for details.
 ---> Grpc.Core.RpcException: Status(StatusCode="Internal", Detail="Error starting gRPC call. HttpRequestException: Connection refused (127.0.0.1:50001) SocketException: Connection refused", DebugException="System.Net.Http.HttpRequestException: Connection refused (127.0.0.1:50001)
 ---> System.Net.Sockets.SocketException (111): Connection refused
   at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.ThrowException(SocketError error, CancellationToken cancellationToken)
   at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.System.Threading.Tasks.Sources.IValueTaskSource.GetResult(Int16 token)
   at System.Net.Sockets.Socket.<ConnectAsync>g__WaitForConnectWithCancellation|283_0(AwaitableSocketAsyncEventArgs saea, ValueTask connectTask, CancellationToken cancellationToken)
   at System.Net.Http.HttpConnectionPool.DefaultConnectAsync(SocketsHttpConnectionContext context, CancellationToken cancellationToken)
   at System.Net.Http.ConnectHelper.ConnectAsync(Func`3 callback, DnsEndPoint endPoint, HttpRequestMessage requestMessage, CancellationToken cancellationToken)
   --- End of inner exception stack trace ---
   at System.Net.Http.ConnectHelper.ConnectAsync(Func`3 callback, DnsEndPoint endPoint, HttpRequestMessage requestMessage, CancellationToken cancellationToken)
   at System.Net.Http.HttpConnectionPool.ConnectAsync(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken)
   at System.Net.Http.HttpConnectionPool.GetHttp2ConnectionAsync(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken)
   at System.Net.Http.HttpConnectionPool.SendWithRetryAsync(HttpRequestMessage request, Boolean async, Boolean doRequestAuth, CancellationToken cancellationToken)
[...]

PR #838, opened 2022-02-15
Wait for sidecar in DaprSecretStoreConfiguration

halspang:
# Description

The secret store configuration provider was trying to access Dapr
during the app startup. If the app started faster than Dapr, it would
get an error trying to access the secrets. This commit lets the
provider wait for the sidecar to come up before making any requests.
If the sidecar does not come up at all, we will still fail.

https://github.com/dapr/dotnet-sdk/issues/779

## Issue reference

Please reference the issue this PR will close: #779 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/src/Dapr.Client/DaprClient.cs
+++ b/src/Dapr.Client/DaprClient.cs
@@ -313,6 +313,23 @@ namespace Dapr.Client
         /// <param name="cancellationToken">A <see cref="CancellationToken" /> that can be used to cancel the operation.</param>
         /// <returns>A <see cref="Task{T}" /> that will return the value when the operation has completed.</returns>
         public abstract Task<bool> CheckHealthAsync(CancellationToken cancellationToken = default);
+
+        /// <summary>
+        /// Perform health-check of Dapr sidecar's outbound APIs. Return 'true' if the sidecar is healthy. Otherwise false. This method should
+        /// be used over <see cref="CheckHealthAsync(CancellationToken)"/> when the health of Dapr is being checked before it starts. This
+        /// health endpoint indicates that Dapr has stood up its APIs and is currently waiting on this application to report fully healthy.
+        /// </summary>
+        /// <param name="cancellationToken">A <see cref="CancellationToken" /> that can be used to cancel the operation.</param>
+        /// <returns>A <see cref="Task{T}" /> that will return the value when the operation has completed.</returns>
+        public abstract Task<bool> CheckOutboundHealthAsync(CancellationToken cancellationToken = default);
+
+        /// <summary>
+        /// Calls <see cref="CheckOutboundHealthAsync(CancellationToken)"/> until the sidecar is reporting as healthy. If the sidecar
+        /// does not become healthy, an exception will be thrown.
+        /// </summary>
+        /// <param name="cancellationToken">A <see cref="CancellationToken" /> that can be used to cancel the operation.</param>
+        /// <returns>A <see cref="Task" /> that will return when the operation has completed.</returns>
+        public abstract Task WaitForSidecarAsync(CancellationToken cancellationToken = default);
         
         /// <summary>
         /// Perform service invocation using the request provided by <paramref name="request" />. The response will
--- a/src/Dapr.Client/DaprClientGrpc.cs
+++ b/src/Dapr.Client/DaprClientGrpc.cs
@@ -305,6 +305,34 @@ namespace Dapr.Client
             }
         }
 
+        public override async Task<bool> CheckOutboundHealthAsync(CancellationToken cancellationToken = default)
+        {
+            var path = "/v1.0/healthz/outbound";
+            var request = new HttpRequestMessage(HttpMethod.Get, new Uri(this.httpEndpoint, path));
+            try
+            {
+                var response = await this.httpClient.SendAsync(request, cancellationToken);
+                return response.IsSuccessStatusCode;
+            }
+            catch (HttpRequestException)
+            {
+                return false;
+            }
+        }
+
+        public override async Task WaitForSidecarAsync(CancellationToken cancellationToken = default)
+        {
+            while (true)
+            { 
+                var response = await CheckOutboundHealthAsync(cancellationToken);
+                if (response)
+                {
+                    break;
+                }
+                await Task.Delay(TimeSpan.FromMilliseconds(500), cancellationToken);
+            }
+        }
+
         public override async Task<HttpResponseMessage> InvokeMethodWithResponseAsync(HttpRequestMessage request, CancellationToken cancellationToken = default)
         {
             ArgumentVerifier.ThrowIfNull(request, nameof(request));
--- a/src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationExtensions.cs
+++ b/src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationExtensions.cs
@@ -17,6 +17,7 @@ using Dapr.Client;
 using Microsoft.Extensions.Configuration;
 using Dapr.Extensions.Configuration.DaprSecretStore;
 using System.Linq;
+using System.Threading;
 
 namespace Dapr.Extensions.Configuration
 {
@@ -58,6 +59,37 @@ namespace Dapr.Extensions.Configuration
         /// </summary>
         /// <param name="configurationBuilder">The <see cref="IConfigurationBuilder"/> to add to.</param>
         /// <param name="store">Dapr secret store name.</param>
+        /// <param name="secretDescriptors">The secrets to retrieve.</param>
+        /// <param name="client">The Dapr client.</param>
+        /// <param name="sidecarWaitTimeout">The <see cref="TimeSpan"/> used to configure the timeout waiting for Dapr.</param>
+        /// <returns>The <see cref="IConfigurationBuilder"/>.</returns>
+        public static IConfigurationBuilder AddDaprSecretStore(
+            this IConfigurationBuilder configurationBuilder,
+            string store,
+            IEnumerable<DaprSecretDescriptor> secretDescriptors,
+            DaprClient client,
+            TimeSpan sidecarWaitTimeout)
+        {
+            ArgumentVerifier.ThrowIfNullOrEmpty(store, nameof(store));
+            ArgumentVerifier.ThrowIfNull(secretDescriptors, nameof(secretDescriptors));
+            ArgumentVerifier.ThrowIfNull(client, nameof(client));
+
+            configurationBuilder.Add(new DaprSecretStoreConfigurationSource()
+            {
+                Store = store,
+                SecretDescriptors = secretDescriptors,
+                Client = client,
+                SidecarWaitTimeout = sidecarWaitTimeout
+            });
+
+            return configurationBuilder;
+        }
+
+        /// <summary>
+        /// Adds an <see cref="IConfigurationProvider"/> that reads configuration values from the Dapr Secret Store.
+        /// </summary>
+        /// <param name="configurationBuilder">The <see cref="IConfigurationBuilder"/> to add to.</param>
+        /// <param name="store">Dapr secret store name.</param>
         /// <param name="metadata">A collection of metadata key-value pairs that will be provided to the secret store. The valid metadata keys and values are determined by the type of secret store used.</param>
         /// <param name="client">The Dapr client</param>
         /// <returns>The <see cref="IConfigurationBuilder"/>.</returns>
@@ -85,6 +117,36 @@ namespace Dapr.Extensions.Configuration
         /// </summary>
         /// <param name="configurationBuilder">The <see cref="IConfigurationBuilder"/> to add to.</param>
         /// <param name="store">Dapr secret store name.</param>
+        /// <param name="metadata">A collection of metadata key-value pairs that will be provided to the secret store. The valid metadata keys and values are determined by the type of secret store used.</param>
+        /// <param name="client">The Dapr client</param>
+        /// <param name="sidecarWaitTimeout">The <see cref="TimeSpan"/> used to configure the timeout waiting for Dapr.</param>
+        /// <returns>The <see cref="IConfigurationBuilder"/>.</returns>
+        public static IConfigurationBuilder AddDaprSecretStore(
+            this IConfigurationBuilder configurationBuilder,
+            string store,
+            DaprClient client,
+            TimeSpan sidecarWaitTimeout,
+            IReadOnlyDictionary<string, string>? metadata = null)
+        {
+            ArgumentVerifier.ThrowIfNullOrEmpty(store, nameof(store));
+            ArgumentVerifier.ThrowIfNull(client, nameof(client));
+
+            configurationBuilder.Add(new DaprSecretStoreConfigurationSource()
+            {
+                Store = store,
+                Metadata = metadata,
+                Client = client,
+                SidecarWaitTimeout = sidecarWaitTimeout
+            });
+
+            return configurationBuilder;
+        }
+
+        /// <summary>
+        /// Adds an <see cref="IConfigurationProvider"/> that reads configuration values from the Dapr Secret Store.
+        /// </summary>
+        /// <param name="configurationBuilder">The <see cref="IConfigurationBuilder"/> to add to.</param>
+        /// <param name="store">Dapr secret store name.</param>
         /// <param name="keyDelimiters">A collection of delimiters that will be replaced by ':' in the key of every secret.</param>
         /// <param name="client">The Dapr client</param>
         /// <returns>The <see cref="IConfigurationBuilder"/>.</returns>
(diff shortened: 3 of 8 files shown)


src/Dapr.Client/DaprClient.cs
```diff
@@ -313,6 +313,21 @@ public HttpRequestMessage CreateInvokeMethodRequest<TRequest>(string appId, stri
         /// <param name="cancellationToken">A <see cref="CancellationToken" /> that can be used to cancel the operation.</param>
         /// <returns>A <see cref="Task{T}" /> that will return the value when the operation has completed.</returns>
         public abstract Task<bool> CheckHealthAsync(CancellationToken cancellationToken = default);
+
+        /// <summary>
+        /// Perform health-check of Dapr sidecar's outbound APIs. Return 'true' if the sidecar is health. Otherwise false.
```
rynowak, line 318:
```suggestion
        /// Perform health-check of Dapr sidecar's outbound APIs. Return 'true' if the sidecar is healthy. Otherwise false.
```

src/Dapr.Client/DaprClient.cs
```diff
@@ -313,6 +313,21 @@ public HttpRequestMessage CreateInvokeMethodRequest<TRequest>(string appId, stri
...
+        /// </summary>
+        /// <param name="cancellationToken">A <see cref="CancellationToken" /> that can be used to cancel the operation.</param>
+        /// <returns>A <see cref="Task{T}" /> that will return the value when the operation has completed.</returns>
+        public abstract Task<bool> CheckOutboundHealthAsync(CancellationToken cancellationToken = default);
+
+        /// <summary>
+        /// Calls <see cref="CheckHealthAsync(CancellationToken)"/> until the sidecar is reporting as healthy. This method checks once a second
+        /// and by default will check 30 times, totaling a 30 second wait.
```
rynowak, line 326: The default timeout here should just be infinite and the cancellation token required. The problem is that any default timeout will be **wrong for someone** and thus needs to be customizable. Cancellation is already the right way to signal a timeout 😆

src/Dapr.Client/DaprClient.cs
```diff
@@ -313,6 +313,21 @@ public HttpRequestMessage CreateInvokeMethodRequest<TRequest>(string appId, stri
         /// <param name="cancellationToken">A <see cref="CancellationToken" /> that can be used to cancel the operation.</param>
         /// <returns>A <see cref="Task{T}" /> that will return the value when the operation has completed.</returns>
         public abstract Task<bool> CheckHealthAsync(CancellationToken cancellationToken = default);
+
+        /// <summary>
+        /// Perform health-check of Dapr sidecar's outbound APIs. Return 'true' if the sidecar is health. Otherwise false.
+        /// </summary>
+        /// <param name="cancellationToken">A <see cref="CancellationToken" /> that can be used to cancel the operation.</param>
+        /// <returns>A <see cref="Task{T}" /> that will return the value when the operation has completed.</returns>
+        public abstract Task<bool> CheckOutboundHealthAsync(CancellationToken cancellationToken = default);
```
rynowak, line 322: What's the difference between this and `CheckHealthAsync` - how will users know which one is appropriate?

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -305,6 +305,35 @@ public override async Task<bool> CheckHealthAsync(CancellationToken cancellation
...
+            {
+                return false;
+            }
+        }
+
+        public override async Task<bool> WaitForSidecarAsync(CancellationToken cancellationToken = default)
+        {
+            while (!cancellationToken.IsCancellationRequested)
+            { 
+                var response = await CheckOutboundHealthAsync(cancellationToken);
```
rynowak, line 327: This code has an issue because **cancellation** can be signaled in two different ways:

1. We detect cancellation in line 325 and return false
2. Cancellation occurs while the request is trying to connect and results in an exception

I think the best fix is to remove the boolean value on wait and just always throw an exception. There's a method on cancellation token that will throw if it's canceled, so we can change line 325 to do that

src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationProvider.cs
```diff
@@ -37,14 +38,37 @@ internal class DaprSecretStoreConfigurationProvider : ConfigurationProvider
...
         /// <param name="secretDescriptors">The secrets to retrieve.</param>
         /// <param name="client">Dapr client used to retrieve Secrets</param>
-        public DaprSecretStoreConfigurationProvider(string store, bool normalizeKey, IEnumerable<DaprSecretDescriptor> secretDescriptors, DaprClient client)
+        /// <param name="cancellationTokenSource">Optional <see cref="CancellationTokenSource"/> used to configure the timeout waiting for Dapr.</param>
+        public DaprSecretStoreConfigurationProvider(
+            string store,
+            bool normalizeKey,
+            IEnumerable<DaprSecretDescriptor> secretDescriptors,
+            DaprClient client,
+            CancellationTokenSource? cancellationTokenSource)
```
rynowak, line 71: Can you explain why this is a source and not a token?

src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationSource.cs
```diff
@@ -54,6 +55,11 @@ public class DaprSecretStoreConfigurationSource : IConfigurationSource
         /// </summary>
         public DaprClient Client { get; set; } = default!;
 
+        /// <summary>
+        /// Gets or sets the cancellation token which controls the timeout for waiting on the Dapr Sidecar's health check.
+        /// </summary>
+        public CancellationTokenSource? CancellationTokenSource { get; set; } = default;
```
rynowak, line 61: It seems like this should be `CancellationToken` not `CancellationTokenSource`

niclasrothman:
Would it also be possible to update the asp.net core examples so they reflect latest changes;
https://github.com/halspang/dotnet-sdk/tree/secret_store_wait/examples/AspNetCore/SecretStoreConfigurationProviderSample
?

Apart from that, good work Halspang!!

src/Dapr.Client/DaprClient.cs
halspang: Done.

src/Dapr.Client/DaprClient.cs
halspang: Yeah, this comment was misleading as we did switch to the `CancellationToken` but forgot to update the comment :)

src/Dapr.Client/DaprClient.cs
halspang: Added to the comment explaining when to use one over the other. Essentially this is the one you want to hit when Dapr/your app are starting. It signifies that the endpoints that Dapr provides are up and ready. The other health check signifies when everything is up, including your application.

I'm also going to submit a docs issue in `dapr/docs` cause it's not referenced in the API Reference.

src/Dapr.Client/DaprClientGrpc.cs
halspang: Done, I was trying to be polite and eat the exception but after chatting it's clear that the exception is actually the expected case.

src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationProvider.cs
halspang: It just felt like an easier way to specify when things were null and use a default. We've switched it around now to use a `TimeSpan` so that we can still control the timeout and also properly dispose of the source.

src/Dapr.Client/DaprClient.cs
rynowak: Sounds good, thanks for adding the description.

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -305,6 +305,35 @@ public override async Task<bool> CheckHealthAsync(CancellationToken cancellation
...
+            }
+            catch (HttpRequestException)
+            {
+                return false;
+            }
+        }
+
+        public override async Task WaitForSidecarAsync(CancellationToken cancellationToken = default)
+        {
+            cancellationToken.ThrowIfCancellationRequested();
```
rynowak, line 325: I think you can remove this, see my other comment on Task.Delay

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -305,6 +305,35 @@ public override async Task<bool> CheckHealthAsync(CancellationToken cancellation
...
+            while (true)
+            { 
+                var response = await CheckOutboundHealthAsync(cancellationToken);
+                if (response)
+                {
+                    break;
+                }
+                await Task.Delay(TimeSpan.FromMilliseconds(500));
```
rynowak, line 333: Change this to https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.delay?view=net-6.0#system-threading-tasks-task-delay(system-int32-system-threading-cancellationtoken) and then this will allow you to cancel immediately out of the `Task.Delay`. 

This way you don't have to check the cancellation token manually in this method because all of *the work* also supports cancellation.

src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationExtensions.cs
```diff
@@ -53,6 +54,37 @@ public static class DaprSecretStoreConfigurationExtensions
...
+        /// <param name="secretDescriptors">The secrets to retrieve.</param>
+        /// <param name="client">The Dapr client.</param>
+        /// <param name="sidecarWaitDuration">The <see cref="TimeSpan"/> used to configure the timeout waiting for Dapr.</param>
+        /// <returns>The <see cref="IConfigurationBuilder"/>.</returns>
+        public static IConfigurationBuilder AddDaprSecretStore(
+            this IConfigurationBuilder configurationBuilder,
+            string store,
+            IEnumerable<DaprSecretDescriptor> secretDescriptors,
+            DaprClient client,
+            TimeSpan sidecarWaitDuration)
```
rynowak, line 71: Minor nitpick: you called this `...duration` but the docs say `... timeout`. Would be simpler to be consistent about the terms used.

src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationProvider.cs
```diff
@@ -37,14 +38,37 @@ internal class DaprSecretStoreConfigurationProvider : ConfigurationProvider
 
         private readonly DaprClient client;
 
+        private readonly TimeSpan sidecarWaitDuration;
+
+        /// <summary>
+        /// Creates a new instance of <see cref="DaprSecretStoreConfigurationProvider"/>.
+        /// </summary>
+        /// <param name="store">Dapr Secre Store name.</param>
```
rynowak, line 46:
```suggestion
        /// <param name="store">Dapr Secret Store name.</param>
```

src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationProvider.cs
```diff
@@ -69,7 +94,31 @@ public DaprSecretStoreConfigurationProvider(string store, bool normalizeKey, IEn
...
+            IList<string>? keyDelimiters,
+            IEnumerable<DaprSecretDescriptor> secretDescriptors,
+            DaprClient client) : this(store, normalizeKey, keyDelimiters, secretDescriptors, client, TimeSpan.FromSeconds(5))
+        {
+        }
+
+        /// <summary>
+        /// Creates a new instance of <see cref="DaprSecretStoreConfigurationProvider"/>.
+        /// </summary>
+        /// <param name="store">Dapr Secre Store name.</param>
```
rynowak, line 109:
```suggestion
        /// <param name="store">Dapr Secret Store name.</param>
```

src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationProvider.cs
```diff
@@ -94,7 +144,28 @@ public DaprSecretStoreConfigurationProvider(string store, bool normalizeKey, ILi
...
+            bool normalizeKey,
+            IReadOnlyDictionary<string, string>? metadata,
+            DaprClient client) : this(store, normalizeKey, metadata, client, TimeSpan.FromSeconds(5))
+        {
+        }
+
+        /// <summary>
+        /// Creates a new instance of <see cref="DaprSecretStoreConfigurationProvider"/>.
+        /// </summary>
+        /// <param name="store">Dapr Secre Store name.</param>
```
rynowak, line 158:
```suggestion
        /// <param name="store">Dapr Secret Store name.</param>
```

src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationProvider.cs
```diff
@@ -103,6 +174,24 @@ public DaprSecretStoreConfigurationProvider(string store, bool normalizeKey, IRe
...
+        /// <param name="normalizeKey">Indicates whether any key delimiters should be replaced with the delimiter ":".</param>
+        /// <param name="keyDelimiters">A collection of delimiters that will be replaced by ':' in the key of every secret.</param>
+        /// <param name="metadata">A collection of metadata key-value pairs that will be provided to the secret store. The valid metadata keys and values are determined by the type of secret store used.</param>
+        /// <param name="client">Dapr client used to retrieve Secrets</param>
+        public DaprSecretStoreConfigurationProvider(
```
rynowak, line 188: How many constructors do we absolutely **need** here? Users are going to interact with this type via `AddSecretStore`, really the constructors are just for proper layering. I think it would be fine to just add a single new constructor.

src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationSource.cs
```diff
@@ -64,11 +70,11 @@ public IConfigurationProvider Build(IConfigurationBuilder builder)
                     throw new ArgumentException($"{nameof(Metadata)} must be null when {nameof(SecretDescriptors)} is set", nameof(Metadata));
                 }
 
-                return new DaprSecretStoreConfigurationProvider(Store, NormalizeKey, KeyDelimiters, SecretDescriptors, Client);
+                return new DaprSecretStoreConfigurationProvider(Store, NormalizeKey, KeyDelimiters, SecretDescriptors, Client, SidecarWaitDuration ?? TimeSpan.FromSeconds(5));
```
rynowak, line 73: minor nitpick: It would be great to have ... 1 place in code where the default timeout lives. It feels wierd when multiple layers of the code all set it.

rynowak: Given that it's spread out over multiple classes, this definitely should be moved to a constant.

test/Dapr.Client.Test/DaprClientTest.InvokeMethodAsync.cs
```diff
@@ -308,6 +309,78 @@ public async Task CheckHealthAsync_WrapsHttpRequestException()
...
+        {
+            await using var client = TestClient.CreateForDaprClient();
+            var cts = new CancellationTokenSource();
+            var waitRequest = await client.CaptureHttpRequestAsync(async daprClient => await daprClient.WaitForSidecarAsync(cts.Token));
+            var healthRequest = await client.CaptureHttpRequestAsync<bool>(async daprClient => await daprClient.CheckOutboundHealthAsync());
+
+            cts.Cancel();
+
+            await healthRequest.CompleteAsync(new HttpResponseMessage(HttpStatusCode.InternalServerError));
+            await Assert.ThrowsAsync<TaskCanceledException>(async () => await waitRequest.CompleteWithExceptionAsync(new TaskCanceledException()));
```
rynowak, line 381: This is a pretty *cool* test. Glad to see this covered.

test/Dapr.Client.Test/DaprClientTest.InvokeMethodAsync.cs
```diff
@@ -308,6 +309,78 @@ public async Task CheckHealthAsync_WrapsHttpRequestException()
...
+
+            // If we don't throw, we're good.
+            await request.CompleteAsync(new HttpResponseMessage());
+        }
+
+        [Fact]
+        public async Task WaitForSidecarAsync_NotSuccessWhenSidecarNotHealthy()
+        {
+            await using var client = TestClient.CreateForDaprClient();
+            var cts = new CancellationTokenSource();
```
rynowak, line 374: nit: `using. .... `

src/Dapr.Client/DaprClientGrpc.cs
halspang: Nifty.

src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationProvider.cs
halspang: Since we're calling it from `DaprSecretStoreConfigurationSource`, which has two distinct paths to creating the Provider, I kept 2 of the constructors. Removed the others which would never be called though.

src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationSource.cs
halspang: Made it a constant.

rynowak (approved):

PR #839, opened 2022-02-17
Added the ability to supply a state manager for testing purposes.

TheRubble:
While trying to test a custom actor I ran into a situation where it wasn't possible to supply a custom state via the actor test options
```
ActorHost.CreateForTest<SubmissionChecksActor>(new ActorTestOptions
{
    TimerManager = timerManager.Object,
});
```

This PR introduces a new interface and allows it to be supplied as per the rest of the items.
I've added a test which shows the usage.

```
public async Task CanSupplyACustomStateStore()
    {
        var stateManager = new Mock<IDaprStateProvider>();
        var host = ActorHost.CreateForTest<CoolTestActor>(new ActorTestOptions(){ DaprStateProvider = stateManager.Object });
        var actor = new CoolTestActor(host);

        await actor.CanTalkToStateManager();

        stateManager.Verify(x => x.ContainsStateAsync(
            "CoolTestActor", 
            actor.Id.GetId(),
            "cool-test-state", CancellationToken.None), Times.Once);
        
    }
```
I've copied the code style where possible.

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Tests run


halspang:
@TheRubble - Thanks for submitting this! I believe you can already achieve this though. Take a look at the `TestActor` implementation and how it's used.

https://github.com/dapr/dotnet-sdk/blob/master/test/Dapr.Actors.Test/ITestActor.cs#L47
https://github.com/dapr/dotnet-sdk/blob/master/test/Dapr.Actors.Test/Runtime/ActorTests.cs#L30

TheRubble:
Hi,

Yes you're correct I can do it that way, I'd totally missed it as I'd focussed on using CreateForTest. 

Couple of comments, it feels a little odd to have a constructor overload purely for testing reasons, perhaps just to me it feels more obvious to supply the dependency through the existing test mechanism.

Looking at the original code ([src/Dapr.Actors/Runtime/ActorManager.cs]):

```
var host = new ActorHost(this.ActorTypeInfo, actorId, this.jsonSerializerOptions, this.loggerFactory, this.proxyFactory, this.timerManager)
{
    StateProvider = new DaprStateProvider(this.daprInteractor, this.jsonSerializerOptions),
};
```

The StateProvider is always set, so perhaps cleaner to supply through the constructor and make the property have a getter only (I did this in the PR)?

Anyways let me know your thoughts as I burnt far too much time looking in the wrong place :-) Once decided I'll knock some docs together to help others.

Thanks

src/Dapr.Actors/Runtime/ActorHost.cs
```diff
@@ -160,7 +164,10 @@ public static ActorHost CreateForTest(Type actorType, string actorTypeName, Acto
         /// Gets the <see cref="ActorTimerManager" />.
         /// </summary>
         public ActorTimerManager TimerManager { get; }
-
-        internal DaprStateProvider StateProvider { get; set; }
+        
+        /// <summary>
+        /// Gets the <see cref="DaprStateProvider" />.
+        /// </summary>
+        internal IDaprStateProvider StateProvider { get; }
```
halspang, line 171: Technically, changing this to an interface is actually a breaking change, so we can't accept it. I think you can accomplish this without it though.

src/Dapr.Actors/Runtime/ActorHost.cs
```diff
@@ -115,20 +116,23 @@ public static ActorHost CreateForTest(Type actorType, string actorTypeName, Acto
         /// <param name="loggerFactory">The logger factory.</param>
         /// <param name="proxyFactory">The <see cref="ActorProxyFactory" />.</param>
         /// <param name="timerManager">The <see cref="ActorTimerManager" />.</param>
+        /// <param name="daprStateProvider">The <see cref="DaprStateProvider" />.</param>
         internal ActorHost(
```
halspang, line 120: You should add a new method here with these params and keep the old one for backwards compatibility.

halspang (changes requested):
I'm onboard with you adding the param to ActorHost for testing, but can we do it without changing to an interface? That creates a breaking change. Instead, add the param and use a mock or an extended version of the `DaprStateProvider` class?

halspang:
@TheRubble - Are you still working on this?

Issue #846, opened 2022-03-07
Event payload doesn't get serialized when using a non-concrete type


wmeints:
## Expected Behavior

Event payload is serialized and delivered on the pub/sub broker.

## Actual Behavior

When using a non-concrete type like `IDomainEvent` in the call the `PublishEventAsync` it will not get serialized correctly.
The result is a cloud event without any payload. 

## Steps to Reproduce the Problem

Have the following event types:

```csharp
public interface IDomainEvent { }
public record MyDomainEvent(string FirstName, string LastName) : IDomainEvent;
```

Then make a call to the DaprClient instance like this:

```csharp
DaprClient daprClient = new DaprClientBuilder().Build();
IDomainEvent domainEvent = new MyDomainEvent("Test", "Test");
    
await daprClient.PublishEventAsync("pubsub", "mytopic", domainEvent);
```

This results in no payload being sent to the pub/sub broker.
However, when you call the same method like this:

```csharp
DaprClient daprClient = new DaprClientBuilder().Build();
MyDomainEvent domainEvent = new MyDomainEvent("Test", "Test");

await daprClient.PublishEventAsync("pubsub", "mytopic", domainEvent);
```

Then it works as expected!

## Release Note

RELEASE NOTE:

halspang:
@wmeints - Thanks for opening this issue! What do you mean by:

> This results in no payload being sent to the pub/sub broker.

Are you saying this path, as it exists today, is sending no data?

wmeints:
I'm getting a cloud event envelope at the receiving service, but the `data` property is missing. This is caused by the `System.Text.Json.JsonSerializer`. It doesn't like to serialize derived classes automatically. 

My PR #847 changes that behavior. However, thinking about it, I'm not sure how secure that is. There must be a reason why it doesn't like to serialize derived classes automatically. I leave it up to you to decide if my PR is a good idea or not.

halspang:
Yeah, this kind of behavior is actually intended. Since the underlying interface has no data, it doesn't serialize with any data. The PR would end up being a breaking change as it is a behavioral change and others may be depending on the existing behavior. So, unfortunately, this wouldn't be a change we could accept. I'll close this issue and the related PR.

wmeints:
Thanks for taking a look anyway! In the meantime, I've found a workaround that fixes my issue.

wmeints:
For people wondering: I've got a similar thing, how am I going to serialize my event? 
Here's the trick:

```csharp
[...]

PR #847, opened 2022-03-07
Fix issue with event serialization

wmeints:
# Description

This PR changes the serialization behavior so that when you're using a non-concrete type variable as input for the `PublishEventAsync` method, it will still be serialized as the concrete type it represents.

## Issue reference

#846

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


halspang:
As per the issue, we won't be able to accept this as this is the intended behavior.

Issue #822, opened 2022-01-25
Add Support for Configuration Subscribe API


halspang:
## Describe the proposal
In 1.6, we added support for the GetConfiguration API. We should also add support for the subscribe portion of the API.

PR #848, opened 2022-03-15
Add Subscribe/Unsubscribe API support

halspang:
# Description

This commit adds support for both the subscribe and unsubscribe
APIs for Darp's Configuration building block. The configuration can
now be exposed through the AspNet builder pattern or by making the
request directly with the DaprClient.

https://github.com/dapr/dotnet-sdk/issues/822

## Issue reference

Please reference the issue this PR will close: #822 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


src/Dapr.Client/ConfigurationSource.cs
```diff
@@ -0,0 +1,21 @@
...
+    public abstract class ConfigurationSource
+    {
+        /// <summary>
+        /// Read any updates that may have occurred in the source configuration. The source may be reading from an ongoing request
+        /// so it utilizes a callback to propagate information back to any concerned parties.
+        /// </summary>
+        /// <param name="callback"><see cref="Action"/> that allows for any changes to propagate back to whatever is watching.</param>
+        /// <param name="cancellationToken">Token used to cancel the operation.</param>
+        /// <returns></returns>
+        public abstract Task ReadSource(Action<ConfigurationItem> callback, CancellationToken cancellationToken = default);
```
rynowak, line 19: This type should probably be replaced with an `IAsyncEnumerable<ConfigurationItem>` - that would be "native"

halspang: Done.

examples/Client/ConfigurationApi/Program.cs
```diff
@@ -1,30 +1,51 @@
...
         {
-            using var client = new DaprClientBuilder().Build();
-
-            Console.WriteLine($"Querying Configuration with key: {QueryKey}");
-            var configItems = await client.GetConfiguration(ConfigStore, new List<string>() { QueryKey });
-
-            if (configItems.Items.Count == 0)
+            Console.WriteLine("Starting application.");
+            // This cancellation token is used to stop the Streaming configuration.
+            using (cts = new CancellationTokenSource())
```
rynowak, line 20: This appears to be unused now

examples/Client/ConfigurationApi/Program.cs
```diff
@@ -1,30 +1,51 @@
...
+        public static IHostBuilder CreateHostBuilder(string[] args)
+        {
+            var client = new DaprClientBuilder().Build();
+            return Host.CreateDefaultBuilder(args)
+                .ConfigureAppConfiguration(config =>
+                {
+                    // Get the initial value and continue to watch it for changes.
+                    config.AddDaprConfigurationStore("redisconfig", new List<string>() { "greeting", "response" }, client, TimeSpan.FromSeconds(20));
+                    config.AddStreamingDaprConfigurationStore("redisconfig", new List<string>() { "greeting", "response" }, client,
+                        TimeSpan.FromSeconds(20), cancellationToken: cts.Token);
```
rynowak, line 42: I'm a little confused about this cancellation token - configuration sources already have a lifecycle, they can implement `Dispose`. So I'm not sure what this is for.

Lifetimes in startup code are tricky, thats why we went with a timeout as a parameter instead of a cancellation token to start with.

examples/Client/ConfigurationApi/README.md
```diff
@@ -13,50 +14,201 @@ It demonstrates the following APIs:
...
-docker exec dapr_redis redis-cli SET greeting "hello world||1"
+### Dapr Client Call
+Both the Get and Subscribe APIs can be called directly through the DaprClient. The Get response
+can be read directly, however, the Subscribe response is recommended to be passed through to an alternate
+thread. This is due to the fact that it is a streaming call and as such, it is not known when the
+stream will stop. If it was read on the main thread, you could be waiting for it forever.
+
+The SDK provides a utility object to help with this, the `ConfigurationWatcher`. This will handle
+creating/monitoring an additional thread for you. Note that both the Get and Subscribe API results can
+be passed to the `ConfigurationWatcher` to simplify the number of sources you have to watch if desired.
```
rynowak, line 29: This sounds out of date

src/Dapr.Client/ConfigurationSource.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+    /// <summary>
+    /// Abstraction around a configuration source.
+    /// </summary>
+    public abstract class ConfigurationSource : IAsyncEnumerable<IEnumerable<ConfigurationItem>>
+    {
+        /// <summary>
+        /// The Id associated with this configuration source.
+        /// </summary>
+        public virtual string Id => string.Empty;
```
rynowak, line 16: What's this for?

src/Dapr.Client/ConfigurationSource.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
```
rynowak, line 1: need copyright

src/Dapr.Client/ConfigurationSource.cs
rynowak: This should be abstract not virtual. I assume `string.Empty` is not a valid value for `Id`?

src/Dapr.Client/DaprSubscribeConfigurationSource.cs
```diff
@@ -0,0 +1,91 @@
+﻿using System;
```
rynowak, line 1: Needs copyright

src/Dapr.Client/DaprSubscribeConfigurationSource.cs
```diff
@@ -0,0 +1,91 @@
...
+        /// </summary>
+        /// <param name="sourceStreamRequest">The request that is used to start the configuration streaming with Dapr.</param>
+        /// <param name="internalClient">The internal client used to call Dapr and begin a gRPC stream.</param>
+        /// <param name="callOptions">The options for the call to Dapr.</param>
+        internal DaprSubscribeConfigurationSource(
+            Autogenerated.Dapr.DaprClient internalClient,
+            Autogenerated.SubscribeConfigurationRequest sourceStreamRequest,
+            CallOptions callOptions) : base()
+        {
+            this.call = internalClient.SubscribeConfigurationAlpha1(sourceStreamRequest, callOptions);
```
rynowak, line 27: I think it makes more sense to put this code outside of this class. Eg: the source should be constructed with the `AsyncServerStreamingCall<>` rather than the client.

src/Dapr.Client/DaprSubscribeConfigurationSource.cs
```diff
@@ -0,0 +1,91 @@
...
+            return new DaprSubscribeConfigurationEnumerator(call, streamId => setIdIfNullOrEmpty(streamId), cancellationToken);
+        }
+
+        private void setIdIfNullOrEmpty(string id)
+        {
+            if (string.IsNullOrEmpty(this.id))
+            {
+                this.id = id;
+            }
+        }
```
rynowak, line 44: So is there no way to get the ID without enumerating the items? This design seems broken.

- What about cases where my subscribe doesn't match any existing keys? 
- What about cases where I subscribe, but then have to shut down before reading any keys?

src/Dapr.Client/DaprSubscribeConfigurationSource.cs
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    internal class DaprSubscribeConfigurationEnumerator : IAsyncEnumerator<IEnumerable<ConfigurationItem>>
```
rynowak, line 47: This seems like a really complex type to handle - `IAsyncEnumerator<IEnumerable<ConfigurationItem>>` means I need a nested-for-loop to read all of the keys.

Is this complexity based on the wire-protocol?

rynowak: BTW since you're implementing your own IEnumerator it would be possible for us to "unwrap" this into a `IAsyncEnumerator<ConfigurationItem>` - would that be better?

src/Dapr.Client/GetConfigurationResponse.cs
```diff
@@ -1,5 +1,4 @@
-﻿using System;
-using System.Collections.Generic;
+﻿using System.Collections.Generic;
```
rynowak, line 1: Needs copyright :)

src/Dapr.Client/SubscribeConfigurationResponse.cs
```diff
@@ -0,0 +1,40 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace Dapr.Client
+{
+    /// <summary>
+    /// Response for a Subscribe Configuration request.
+    /// </summary>
+    [Obsolete("This response utilizes an alpha API which is subject to change. This attribute will be removed when the API is no longer Alpha.")]
+    public class SubscribeConfigurationResponse
```
rynowak, line 10: Why do we need this class **and** `ConfigurationSource`. It looks like this could just be replaced with `ConfigurationSource` without losing anything.

src/Dapr.Client/UnsubscribeConfigurationResponse.cs
```diff
@@ -0,0 +1,30 @@
...
+    {
+        /// <summary>
+        /// Boolean representing if the request was successful or not.
+        /// </summary>
+        public bool Ok { get; }
+
+        /// <summary>
+        /// The message from the Configuration API.
+        /// </summary>
+        public string Message { get; }
```
rynowak, line 17: This seems confusing with the `Ok` and `Message` fields. What would happen if we threw an exception on failure? That might be more idiomatic.

src/Dapr.Extensions.Configuration/DaprConfigurationStoreExtension.cs
```diff
@@ -0,0 +1,94 @@
+﻿using System;
```
rynowak, line 1: Needs copyright

src/Dapr.Extensions.Configuration/DaprConfigurationStoreExtension.cs
```diff
@@ -0,0 +1,94 @@
...
+using Dapr.Client;
+using Microsoft.Extensions.Configuration;
+
+namespace Dapr.Extensions.Configuration
+{
+    /// <summary>
+    /// Extension used to call the Dapr Configuration API and store the values in a <see cref="IConfiguration"/>.
+    /// </summary>
+    [Obsolete]
+    public static class DaprConfigurationStoreExtension
```
rynowak, line 13: Do we have an existing class with extension methods for `IConfigurationBuilder`? I expected to see these methods added to that class.

src/Dapr.Extensions.Configuration/DaprConfigurationStoreExtension.cs
```diff
@@ -0,0 +1,94 @@
...
+        /// </summary>
+        /// <param name="configurationBuilder">The <see cref="IConfigurationBuilder"/> to add to.</param>
+        /// <param name="store">The configuration store to query.</param>
+        /// <param name="keys">The keys, if any, to request. If empty, returns all configuration items.</param>
+        /// <param name="client">The <see cref="DaprClient"/> used for the request.</param>
+        /// <param name="sidecarWaitTimeout">The <see cref="TimeSpan"/> used to configure the timeout waiting for Dapr.</param>
+        /// <param name="metadata">Optional metadata sent to the configuration store.</param>
+        /// <param name="cancellationToken"><see cref="CancellationToken"/> used to cancel the operation.</param>
+        /// <returns>The <see cref="IConfigurationBuilder"/>.</returns>
+        public static IConfigurationBuilder AddStreamingDaprConfigurationStore(
```
rynowak, line 67: BTW I like the approach of giving this a different method name. It's very obvious what the difference is.

src/Dapr.Extensions.Configuration/DaprConfigurationStoreProvider.cs
```diff
@@ -0,0 +1,99 @@
+﻿using System;
```
rynowak, line 1: Needs copyright.

src/Dapr.Extensions.Configuration/DaprConfigurationStoreProvider.cs
```diff
@@ -0,0 +1,99 @@
...
+    /// call or a streaming call.
+    /// </summary>
+    [Obsolete]
+    public class DaprConfigurationStoreProvider : ConfigurationProvider
+    {
+        private string store { get; }
+        private IReadOnlyList<string> keys { get; }
+        private DaprClient daprClient { get; }
+        private TimeSpan sidecarWaitTimeout { get; }
+        private bool isStreaming { get; } = false;
```
rynowak, line 22: By convention properties are `PascalCase`.

src/Dapr.Extensions.Configuration/DaprConfigurationStoreProvider.cs
```diff
@@ -0,0 +1,99 @@
...
+using Microsoft.Extensions.Primitives;
+
+namespace Dapr.Extensions.Configuration
+{
+    /// <summary>
+    /// A configuration provider that utilizes the Dapr Configuration API. It can either be a single, constant
+    /// call or a streaming call.
+    /// </summary>
+    [Obsolete]
+    public class DaprConfigurationStoreProvider : ConfigurationProvider
```
rynowak, line 16: Does this need to be public?

src/Dapr.Extensions.Configuration/DaprConfigurationStoreProvider.cs
```diff
@@ -0,0 +1,99 @@
...
+            this.isStreaming = isStreaming;
+            this.metadata = metadata;
+            this.cancellationToken = cancellationToken;
+        }
+
+        /// <inheritdoc/>
+        public override bool TryGet(string key, out string? value)
+        {
+            return Data.TryGetValue(key, out value);
+        }
```
rynowak, line 59: Can you explain why this override is needed? I would expect the default impl to do this.

src/Dapr.Extensions.Configuration/DaprConfigurationStoreProvider.cs
```diff
@@ -0,0 +1,99 @@
...
+                    await foreach (var items in subscribeConfigurationResponse.Source)
+                    {
+                        // Whenever we get an update, make sure to update the reloadToken.
+                        OnReload();
+                        foreach (var item in items)
+                        {
+                            Set(item.Key, item.Value);
+                        }
+                    }
+                }, cancellationToken);
```
rynowak, line 86: Is this cancellation token being passed into `Task.Run`? I don't think we need it if that's the case. 

https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.run?view=net-6.0#system-threading-tasks-task-run(system-func((system-threading-tasks-task))-system-threading-cancellationtoken)

The cancellation token is only useful to prevent running the function when it **has not yet been scheduled**. This will have no effect on a running function.

src/Dapr.Extensions.Configuration/DaprConfigurationStoreProvider.cs
```diff
@@ -0,0 +1,99 @@
...
+            {
+                await daprClient.WaitForSidecarAsync(tokenSource.Token);
+            }
+
+            if (isStreaming)
+            {
+                var subscribeConfigurationResponse = await daprClient.SubscribeConfiguration(store, keys, metadata, cancellationToken);
+                subscribeTask = Task.Run(async () =>
+                {
+                    await foreach (var items in subscribeConfigurationResponse.Source)
```
rynowak, line 77: What we **do** need is to pass a cancellation token here. This token should be triggered by disposal of the config provider.

src/Dapr.Extensions.Configuration/DaprConfigurationStoreProvider.cs
```diff
@@ -0,0 +1,99 @@
...
+        /// <param name="metadata">Optional metadata sent to the configuration store.</param>
+        /// <param name="cancellationToken"><see cref="CancellationToken"/> used to cancel the operation.</param>
+        public DaprConfigurationStoreProvider(
+            string store,
+            IReadOnlyList<string> keys,
+            DaprClient daprClient,
+            TimeSpan sidecarWaitTimeout,
+            bool isStreaming = false,
+            IReadOnlyDictionary<string, string>? metadata = default,
+            CancellationToken cancellationToken = default)
```
rynowak, line 44: I think we should remove this `cancellationToken` parameter. This is a very tricky design and it will be hard for users to get the right cancellation behavior.

Instead we should implement `Dispose` and use that to cancel.

src/Dapr.Extensions.Configuration/DaprConfigurationStoreProvider.cs
```diff
@@ -0,0 +1,99 @@
...
+
+            if (isStreaming)
+            {
+                var subscribeConfigurationResponse = await daprClient.SubscribeConfiguration(store, keys, metadata, cancellationToken);
+                subscribeTask = Task.Run(async () =>
+                {
+                    await foreach (var items in subscribeConfigurationResponse.Source)
+                    {
+                        // Whenever we get an update, make sure to update the reloadToken.
+                        OnReload();
```
rynowak, line 80: I think `OnReload` should be called after making updates. `OnReload` triggers all of the consumers to refresh their snapshot, so if you call it first then you won't see any of the new values yet :)

src/Dapr.Extensions.Configuration/DaprConfigurationStoreProvider.cs
```diff
@@ -0,0 +1,99 @@
...
+                        foreach (var item in items)
+                        {
+                            Set(item.Key, item.Value);
```
rynowak, line 83: Is this thread-safe? We might need to *copy-and-swap*.  Can you leave a comment if this **is** thread-safe?

src/Dapr.Extensions.Configuration/DaprConfigurationStoreProvider.cs
```diff
@@ -0,0 +1,99 @@
...
+                        }
+                    }
```
rynowak, line 85: We should probably do some error handling here, and that might include opening a new streaming connection. As this is written you'll stop seeing updates if the connection **ever** drops.

src/Dapr.Extensions.Configuration/DaprConfigurationStoreSource.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        /// <summary>
+        /// Gets or sets the <see cref="TimeSpan"/> that is used to control the timeout waiting for the Dapr sidecar to become healthly.
+        /// </summary>
+        public TimeSpan SidecarWaitTimeout { get; set; }
+
+        /// <summary>
+        /// The optional metadata to be sent to the configuration store.
+        /// </summary>
+        public IReadOnlyDictionary<string, string>? Metadata { get; set; } = default;
```
rynowak, line 43: Does the null-ness of this field affect the semantics/behavior? If not, then it would be better to initialize this to a non-null value.

src/Dapr.Extensions.Configuration/DaprConfigurationStoreSource.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        /// <summary>
+        /// The <see cref="CancellationToken"/> that is used to cancel the request.
+        /// </summary>
+        public CancellationToken CancellationToken { get; set; } = default;
```
rynowak, line 48: See comments elsewhere, we should try to get rid of this.

src/Dapr.Client/DaprSubscribeConfigurationSource.cs
halspang: Not currently, no. You've already commented on the issue where I ask for it in the dapr API but for anyone else who's interested: https://github.com/dapr/dapr/issues/4546

src/Dapr.Client/DaprSubscribeConfigurationSource.cs
halspang: It's based off the wire values right now. Each response can contain a list of items. I'll see about unwrapping them as I agree it is a bit unwieldy.

src/Dapr.Client/UnsubscribeConfigurationResponse.cs
halspang: I can add an exception for it, but this is also what the wire returns. 

Thoughts on adding an exception if we get Ok == false and instead throwing a `DaprException(Message)` @rynowak?

src/Dapr.Extensions.Configuration/DaprConfigurationStoreExtension.cs
halspang: We don't have any classes that extend `IConfigurationBuilder` as far as I can tell. I followed what the `DaprSecretStoreConfigurationExtensions.cs` class did.

src/Dapr.Extensions.Configuration/DaprConfigurationStoreProvider.cs
halspang: Sure doesn't.

src/Dapr.Extensions.Configuration/DaprConfigurationStoreProvider.cs
halspang: That's leftover from when I was trying to understand how the reload token works. I thought I had to implement the actual interface instead of the base class that I'm doing now. Removed!

src/Dapr.Extensions.Configuration/DaprConfigurationStoreProvider.cs
halspang: Nothing needs to access these so I just removed the `{ get; }` and left them camel case.

src/Dapr.Extensions.Configuration/DaprConfigurationStoreProvider.cs
halspang: I didn't that distinction, removed from the Task.

src/Dapr.Extensions.Configuration/DaprConfigurationStoreProvider.cs
halspang: Makes sense to me, added `IDisposable` to this as it looks like the base level on of `IConfigurationProvider` doesn't. Let me know if that's not the right interface for this :)

src/Dapr.Extensions.Configuration/DaprConfigurationStoreSource.cs
halspang: I don't believe it does. I'll try it with an empty dictionary and see if things still work as expected.

src/Dapr.Extensions.Configuration/DaprConfigurationStoreSource.cs
halspang: Done.

src/Dapr.Client/ConfigurationSource.cs
halspang: Yeah I didn't realize that you could make a property abstract but not a member variable, made it abstract.

src/Dapr.Client/DaprSubscribeConfigurationSource.cs
halspang: Playing around with this more, I don't think we can fully unwrap it without some complex logic on our side. Each call of `MoveNext` on the underlying stream gives us a new list of items. 

So we could track which item we're on in the list and then only call `MoveNext` when we're out of items but that feels like we could miss updates and like an easy place to introduce an error.

src/Dapr.Client/SubscribeConfigurationResponse.cs
halspang: At this point it was just to keep the similarity with the `GetConfigurationResponse`. I didn't want to give this the class the implementation of the `DaprSubscribeConfigurationSource` because it exposed some weird internals. Now that it's just taking the server stream it's a bit better. 

The other reasoning being I thought it'd be easier for things like testing but I guess you can just have the client give back a difference source anyway.

rynowak: Ok, I think it's fine I was just curious. If you want to keep them separated then lets keep them separated.

src/Dapr.Extensions.Configuration/DaprConfigurationStoreProvider.cs
halspang: I'm going to change it to a copy and swap as I think the interaction between the various readers and writers could be weird.

rynowak (approved):

PR #867, opened 2022-04-20
Fix Adds DaprClient in Dapr.AspNetCore from AddSingleton to TryAddSingleton

zhenlei520:
…gleton

Signed-off-by: zhenlei520 <wangzhenlei520@gmail.com>

# Description

I hope that AddDaprClient can be changed to TryAddSingleton. Although the AddDaprClient method only adds DaprClient to the current class library once, if other class libraries also add DaprClient and then use Dapr.AspNetCore, DaprClient will be added again.

### DIFF
--- a/src/Dapr.AspNetCore/DaprServiceCollectionExtensions.cs
+++ b/src/Dapr.AspNetCore/DaprServiceCollectionExtensions.cs
@@ -16,6 +16,7 @@ namespace Microsoft.Extensions.DependencyInjection
     using System;
     using System.Linq;
     using Dapr.Client;
+    using Extensions;
 
     /// <summary>
     /// Provides extension methods for <see cref="IServiceCollection" />.
@@ -44,7 +45,7 @@ namespace Microsoft.Extensions.DependencyInjection
 
             services.AddSingleton<DaprClientMarkerService>();
 
-            services.AddSingleton(_ =>
+            services.TryAddSingleton(_ =>
             {
                 var builder = new DaprClientBuilder();
                 if (configure != null)


halspang:
@zhenlei520 - Thanks for opening this! Though I think this is already solved in the code: https://github.com/dapr/dotnet-sdk/pull/867/files#diff-6b7b1ed59d457676840fddee293b56b4f8354d2b5bbad4df9c8d8fb11103ff15R41

zhenlei520:
> @zhenlei520 - Thanks for opening this! Though I think this is already solved in the code: https://github.com/dapr/dotnet-sdk/pull/867/files#diff-6b7b1ed59d457676840fddee293b56b4f8354d2b5bbad4df9c8d8fb11103ff15R41

I am very happy to receive your reply, but I think there is a problem here. If you directly use the methods in the `Dapr.AspNetCore` package, both AddDaprClient and AddDapr support multiple additions, and DaprClient will not be added multiple times.  but DaprClient is the core capability of Dapr.Client. If I only need to use service calls, then I can do Add DaprClient myself, but if I do this, once I use AddDapr again later, it will cause DaprClient to be called add multiple times

zhenlei520:
https://github.com/dapr/dotnet-sdk/blob/452ccad335498361b229795229398e26a42d9802/src/Dapr.AspNetCore/DaprServiceCollectionExtensions.cs#L47

I think this should be replaced by TryAddSingleton, or the AddDaprClient method should be provided by `Dapr.Client` instead of `Dapr.AspNetCore`

halspang:
@zhenlei520 - Is the use case you're talking about adding `DaprClient` to the services manually, without calling one of the helper methods?

Also, I don't think `Dapr.Client` should own this code path. The services model that we're going off of here is an AspNet behavior and thus the code belongs in the `Dapr.AspNetCore` package.

rynowak:
> Also, I don't think `Dapr.Client` should own this code path. The services model that we're going off of here is an AspNet behavior and thus the code belongs in the `Dapr.AspNetCore` package.

We left this out of `Dapr.Client` to avoid extra dependencies. It doesn't feel worthwhile to add a dependency so that we can add a method that includes one line of code.

zhenlei520:
> > Also, I don't think `Dapr.Client` should own this code path. The services model that we're going off of here is an AspNet behavior and thus the code belongs in the `Dapr.AspNetCore` package.
> 
> We left this out of `Dapr.Client` to avoid extra dependencies. It doesn't feel worthwhile to add a dependency so that we can add a method that includes one line of code.

I now have a package only for sending messages. In fact, I only need to handle the injection of `DaprClient` by myself, and then pass it through DaprClient's Pub. I don't need to rely on `Dapr.AspNetCore`, but if I do , if the user calls the `AddDapr` method after using my package, it will cause the `DaprClient` to register multiple times

zhenlei520:
Any news on this?

halspang (approved):

Issue #857, opened 2022-04-07
Add Support for Per Actor Type Configuration


halspang:
## Describe the proposal
The Dapr Runtime added support for per actor type configurations in 1.7. This support needs to be added to the SDKs.

The new configuration looks like:

```json
{
  "entities":["actorType1", "actorType2"],
  "actorIdleTimeout": "1h",
  "actorScanInterval": "30s",
  "drainOngoingCallTimeout": "30s",
  "drainRebalancedActors": true,
  "reentrancy": {
    "enabled": true,
    "maxStackDepth": 32
  },
  "entitiesConfig": [
      {
          "entities": ["actorType1"],
          "actorIdleTimeout": "1m",
          "drainOngoingCallTimeout": "10s",
          "reentrancy": {
              "enabled": false
          }
      }
  ]
}
```

Implementation can be varied per SDK but should output to the above JSON.

Docs: https://github.com/dapr/docs/pull/2334
Runtime: https://github.com/dapr/dapr/pull/4201

PR #870, opened 2022-04-29
Add support for per type actor configuration

halspang:
# Description

This commit allows different actor types to provide their own
configurations instead of relying on the top-level values. Any
value defined in them will be used instead of the top-level.
Anything that is left out will use the top-level or default if
it is undefined.

https://github.com/dapr/dotnet-sdk/issues/857

## Issue reference

Please reference the issue this PR will close: #857

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


src/Dapr.Actors/Runtime/ActorRuntime.cs
```diff
@@ -80,44 +81,70 @@ internal Task SerializeSettingsAndRegisteredTypes(IBufferWriter<byte> output)
 
             writer.WriteEndArray();
 
-            if (this.options.ActorIdleTimeout != null)
+            writeActorOptions(writer, this.options);
+
+            var actorsWithConfigs = this.options.Actors.Where(actor => actor.TypeOptions != null).ToList();
```
rynowak, line 86: Is this right? what about actors without configs? where are they handled

rynowak: I get it, actors without config don't need this data to be specified.

Issue #885, opened 2022-06-01
Add distributed lock API support


amulyavarote:
## Describe the proposal

Add methods for distributed lock API.

Main Issue: https://github.com/dapr/dapr/issues/3549

PR #886, opened 2022-06-03
Added dotnet methods for Distributed Lock API

amulyavarote:
Signed-off-by: Amulya Varote <amulyavarote@microsoft.com>

# Description

Added dotnet methods for Distributed Lock API

_Please explain the changes you've made_

## Issue reference

Please reference the issue this PR will close: #885 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


saber-wang:
Can I add an automatic unlocking method？
for example
```
using( var tryLockResponse = await client.TryLock(StoreName, ResourceId, LockOwner, ExpiryInSeconds))
{
//code
}
```
If a lock is held, it is automatically unlocked at the end of the scope.

halspang (changes requested):
Implementation itself seems mostly there, though I think I want an enum for the Unlock return.

My main concern is around the example. First, should we publish the example if people can't use it with an out of the box dapr? Second, I think it should be a little more real-world. Can we demonstrate it locking around a resource that two applications are both trying to use? I think that's better than just an example of the API being called.

examples/Client/DistributedLockApi/README.md
```diff
@@ -0,0 +1,76 @@
...
+## Prerequisites
+
+- [.NET Core 3.1 or .NET 5+](https://dotnet.microsoft.com/download) installed
+- [Dapr CLI](https://docs.dapr.io/getting-started/install-dapr-cli/)
+- [Initialized Dapr environment](https://docs.dapr.io/getting-started/install-dapr-selfhost/)
+- [Dapr .NET SDK](https://docs.dapr.io/developing-applications/sdks/dotnet/)
+
+## Overview
+This example shows the usage of two different Distributed Lock APIs. The TryLock call and
+the Unlock call. Both of these calls can be handled in two different ways.
```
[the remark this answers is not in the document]
amulyavarote: Yes!

examples/Client/DistributedLockApi/README.md
```diff
@@ -0,0 +1,76 @@
...
+- **Distributed Lock**: Try Lock
+- **Distributed Lock**: Unlock
+
+> **Note:** Make sure to use the latest proto bindings
+
+## Prerequisites
+
+- [.NET Core 3.1 or .NET 5+](https://dotnet.microsoft.com/download) installed
+- [Dapr CLI](https://docs.dapr.io/getting-started/install-dapr-cli/)
+- [Initialized Dapr environment](https://docs.dapr.io/getting-started/install-dapr-selfhost/)
```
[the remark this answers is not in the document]
amulyavarote: Yea! I added this for now. We can edit it after the dapr with distributed api lock is released.
`Initialize Dapr environment by pulling the code from master branch of [Dapr](https://github.com/dapr/dapr)`

src/Dapr.Client/DaprClient.cs
```diff
@@ -859,6 +859,39 @@ public async Task<StateEntry<TValue>> GetStateEntryAsync<TValue>(string storeNam
             string id,
             CancellationToken cancellationToken = default);
 
+        /// <summary>
+        /// Get a response if the lock is acquired from the given lockstore.
```
halspang, line 863: I'd change this to:

`Attempt to lock the given resourceId with response indicating success.`

Provided that that is accurate. Does this wait for the lock to become available or does it fail immediately if the lock is already there?

src/Dapr.Client/DaprClient.cs
```diff
@@ -859,6 +859,39 @@ public async Task<StateEntry<TValue>> GetStateEntryAsync<TValue>(string storeNam
...
+        public abstract Task<TryLockResponse> TryLock(
+            string storeName,
+            string resourceId,
+            string lockOwner,
+            Int32 expiryInSeconds,
+            CancellationToken cancellationToken = default);
+
+
+        /// <summary>
+        /// Get a response if the lock is unlocked.
```
halspang, line 881: Similar to above, this implies that you'll only get a response if the lock is unlocked. I'd change it to:

`Attempt to unlock the given resourceId with response indicating success.`

test/Dapr.Client.Test/DistributedLockApiTest.cs
```diff
@@ -0,0 +1,84 @@
...
+using System.Threading.Tasks;
+using Autogenerated = Dapr.Client.Autogen.Grpc.v1;
+using Xunit;
+using FluentAssertions;
+using System;
+
+namespace Dapr.Client.Test
+{
+    [System.Obsolete]
+    public class DistributedLockApiTest
```
halspang, line 23: I think it's worth it to make sure the call fails if it's missing the correct parameters. Especially the `lockDuration` since we define that logic ourselves instead of using the `ArgumentVerifier`.

examples/Client/DistributedLockApi/UnLockApplication/Program.cs
```diff
@@ -0,0 +1,44 @@
...
+            
+            //Trying to acquire the lock which is being used by the other process(TryLockApplication)
+            var tryLockResponse = await client.TryLock(StoreName, ResourceId, LockOwner, ExpiryInSeconds);
+            Console.WriteLine("Acquired Lock? " + tryLockResponse.Success);
+            Console.WriteLine("Lock cannot be acquired as it belongs to the other process");
+
+            var lockResponse = await client.Unlock(StoreName, ResourceId, LockOwner);
+            Console.WriteLine("Unlock API response when lock is acquired by a different process: " + lockResponse.status);
+
+            System.Threading.Thread.Sleep(25000);
```
halspang, line 28: Instead of a just a `Thread.Sleep` here, can we do a while loop with a sleep where we try and acquire the lock?

I think it's important to show how we think people should be handling the situation of waiting for another app to unlock. 

Not for this PR, but maybe that's something we add as a convenience feature to the client eventually? A blocking lock call?

examples/Client/DistributedLockApi/UnLockApplication/Program.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.Threading.Tasks;
+using System.Threading;
+using Dapr.Client;
+
+namespace DistributedLockApi
+{
+    public class Program2
+    {
+        [Obsolete]
+        public static async Task Main(string[] args)
```
halspang, line 12: Nit: your file path no longer matches the name you give this in the run command.

halspang (changes requested):
Instead of duplicating the components, can you have a shared folder for them?

The example is better now, but I think could still be a little more real world. I added a comment with a small change for that.

src/Dapr.Client/DaprClient.cs
```diff
@@ -859,6 +859,39 @@ public async Task<StateEntry<TValue>> GetStateEntryAsync<TValue>(string storeNam
...
+        /// Get a response if the lock is acquired from the given lockstore. 
+        /// </summary>
+        /// <param name="storeName">The name of the lock store to be queried.</param>
+        /// <param name="resourceId">Lock key that stands for which resource to protect.</param>
+        /// <param name="lockOwner">Indicates the identifier of lock owner.</param>
+        /// <param name="expiryInSeconds">The time after which the lock gets expired.</param>
+        /// <param name="cancellationToken">A <see cref="CancellationToken" /> that can be used to cancel the operation.</param>
+        /// <returns>A <see cref="Task"/> containing a <see cref="TryLockResponse"/></returns>
+        [Obsolete("This API is currently not stable as it is in the Alpha stage. This attribute will be removed once it is stable.")]
+        public abstract Task<TryLockResponse> TryLock(
```
rynowak, line 872: Some care is needed here to make this API idiomatic to use in .NET. For example, this is not the right API signature to use with a `TryXyz` API and would be confusing to use for a .NET developer.

I'd expect `TryLock` to return a boolean value, as it's very idiomatic to write:

```C#
if (await client.TryLock(...))
{
  try
  {
     // do work here
  }
  finally
  {
    await client.Unlock(...);
  }
}
```

Alternatively, rename the method. 

Examples: 

https://docs.microsoft.com/en-us/dotnet/api/system.threading.spinlock.tryenter?view=net-6.0
https://docs.microsoft.com/en-us/dotnet/api/system.threading.mutex.tryopenexisting?view=net-6.0

src/Dapr.Client/DaprClient.cs
```diff
@@ -859,6 +859,39 @@ public async Task<StateEntry<TValue>> GetStateEntryAsync<TValue>(string storeNam
...
+        /// <summary>
+        /// Get a response if the lock is unlocked. 
+        /// </summary>
+        /// <param name="storeName">The name of the lock store to be queried.</param>
+        /// <param name="resourceId">Lock key that stands for which resource to protect.</param>
+        /// <param name="lockOwner">Indicates the identifier of lock owner.</param>
+        /// <param name="cancellationToken">A <see cref="CancellationToken" /> that can be used to cancel the operation.</param>
+        /// <returns>A <see cref="Task"/> containing a <see cref="UnlockResponse"/></returns>
+        [Obsolete("This API is currently not stable as it is in the Alpha stage. This attribute will be removed once it is stable.")]
+        public abstract Task<UnlockResponse> Unlock(
```
rynowak, line 889: Something we need to think about here: what error cases exist and what categories do they fall into?

- Programming mistake
- Expected failure cases the user would want to handle
- Unlikely, unrecoverable failures

The API as it is doesn't throw exceptions for any of these cases, so it doesn't communicate well with the user.

examples/Client/DistributedLockApi/TryLockApplication/Program.cs
```diff
@@ -0,0 +1,42 @@
...
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Dapr.Client;
+
+namespace DistributedLockApi
+{
+    public class Program
+    {
+        [Obsolete]
+        public static async Task Main(string[] args)
```
rynowak, line 11: This feels strange to me as a sample because it intentionally exercises error cases, instead of demonstrating the correct way to do error handling. 

We should write a sample that shows how to use the API properly.

amulyavarote: I will be reworking on the examples to make them real world.

src/Dapr.Client/DaprClient.cs
amulyavarote: As a developer of the application, I would like to decide what my application does based on the status messages from the lock APIs. Therefore, the sdk returns the messages to the users and the users can decide what to do based on the message.

saber-wang:
Maybe you can return an 'IDisposable' or 'IAsycnDisposable', so that we can use the following example
```
using(var lock=await client.TryLock(...))
{
if(lock != null)
  //todo 
}
```

examples/Client/DistributedLockApi/UnLockApplication/Program.cs
```diff
@@ -0,0 +1,49 @@
...
+                            await client.Unlock(StoreName, ResourceId, LockOwner);
+                        }
+                        break;
+                    }    
+                }
+                catch(Exception ex)
+                {
+                    Console.WriteLine($"ERROR: Got exception while acquiring the lock. Exception: {ex}");   
+                }
+                Console.WriteLine("Lock is acquired by a different process");
```
halspang, line 44: Both of these will output when you get an exception. It should only be one or the other.

halspang (changes requested):
Nit: Can you make your `.csproj` files match the application name? `TryLockApplication.csproj` and `UnLockApplication.csproj`?

This and one other small change in the examples. @rynowak - Do you think this example is fitting for showing how to handle the API?

halspang:
@saber-wang - I think that's a good suggestion. 

@amulyavarote, can we look into adding it?

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -1278,6 +1278,81 @@ private Autogenerated.StateItem ToAutogeneratedStateItem(StateTransactionRequest
         }
         #endregion
 
+        #region Distributed Lock API
+        /// <inheritdoc/>
+        [Obsolete]
+        public async override Task<TryLockResponse> TryLock(
```
halspang, line 1284: Since this is no longer returning a boolean, it should be `Lock` instead of `TryLock`.

src/Dapr.Client/TryLockResponse.cs
```diff
@@ -0,0 +1,64 @@
...
+    /// <summary>
+    /// Class representing the response from a TryLock API call.
+    /// </summary>
+    public class TryLockResponse :  IDisposable
+    {
+        /// <summary>
+        /// The success value of the tryLock API call
+        /// </summary>
+        public bool Success { get; }
+        /// <summary>
```
halspang, line 27: Nit: new line between one member and the next set of comments.

src/Dapr.Client/TryLockResponse.cs
```diff
@@ -0,0 +1,64 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Disposes the resources associated with the object.
+        /// </summary>
+        /// <param name="disposing"><c>true</c> if called by a call to the <c>Dispose</c> method; otherwise false.</param>
+        protected virtual void Dispose(bool disposing)
+        {
+        }
```
halspang, line 62: You don't actually need both methods, you can just use `Dispose()`.

src/Dapr.Client/TryLockResponse.cs
```diff
@@ -0,0 +1,64 @@
...
+        public TryLockResponse(bool success)
+        {
+            this.Success = success;
+        }
+
+        /// <inheritdoc />
+        public void Dispose() {
+            using (var client = new DaprClientBuilder().Build()) {
+                client.Unlock(ResourceId, LockOwner, StoreName);
+            }
```
halspang, line 53: This should check the `Success` value and only unlock if it actually locked in the first place.

src/Dapr.Client/TryLockResponse.cs
```diff
@@ -0,0 +1,64 @@
...
+        /// Constructor for a TryLockResponse.
+        /// </summary>
+        /// <param name="success">The success value that is returned in the TryLock call.</param>
+        public TryLockResponse(bool success)
+        {
+            this.Success = success;
+        }
+
+        /// <inheritdoc />
+        public void Dispose() {
```
halspang, line 50: We should try and validate this behavior in a test.

halspang (changes requested):

examples/Client/DistributedLockApi/UnLockApplication/Program.cs
```diff
@@ -0,0 +1,51 @@
...
+                            {
+                                await client.SaveStateAsync(DAPR_STORE_NAME, "deposit", "300");
+                                var result = await client.GetStateAsync<string>(DAPR_STORE_NAME, "deposit");
+                                Console.WriteLine("Getting deposited value: " + result);
+                            }
+                            finally
+                            {
+                                await client.Unlock(StoreName, ResourceId, LockOwner);
+                            }
+                            break;
```
halspang, line 39: `using` does this for us.

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -1278,6 +1278,81 @@ private Autogenerated.StateItem ToAutogeneratedStateItem(StateTransactionRequest
...
+            catch (RpcException ex)
+            {
+                throw new DaprException("Lock operation failed: the Dapr endpoint indicated a failure. See InnerException for details.", ex);
+            }
+
+            return new TryLockResponse(response.Success);
+        }
+
+        /// <inheritdoc/>
+        public async override Task<UnlockResponse> Unlock(
```
halspang, line 1323: Missed this earlier, but this should also have an `[Obsolete]`.

examples/Client/DistributedLockApi/README.md
```diff
@@ -0,0 +1,78 @@
...
+> **Note:** Make sure to use the latest proto bindings
+
+## Prerequisites
+
+- [.NET Core 3.1 or .NET 5+](https://dotnet.microsoft.com/download) installed
+- [Dapr CLI](https://docs.dapr.io/getting-started/install-dapr-cli/)
+- [Dapr .NET SDK](https://docs.dapr.io/developing-applications/sdks/dotnet/)
+- Initialize Dapr environment by pulling the code from master branch of [Dapr](https://github.com/dapr/dapr)
+
+## Overview
```
greenie-msft, line 17: Can we provide a little more context in this overview? When would a developer need to ouse this API? At the very least, can we link back to the building block overview page that describes this API in more detail?

amulyavarote: Hi Nick! I am waiting for the distributed lock api overview page. Once it is merged on the docs side, I will go ahead and add that link here.

examples/AspNetCore/SecretStoreConfigurationProviderSample/SecretStoreConfigurationProviderSample.csproj
```diff
@@ -5,6 +5,7 @@
     <TargetFramework>netcoreapp3.1</TargetFramework>
   </PropertyGroup>
 
+  <PropertyGroup Condition=" '$(RunConfiguration)' == 'blq' " />
```
rynowak, line 8: What is this?

examples/Client/DistributedLock/1fa9b73d-36f1-4be1-80ff-034f28b65352
```diff
@@ -0,0 +1 @@
+{"number":87,"analysis":null}
```
rynowak, line 1: What is this? it looks like something needs to be added to `.gitignore`

examples/Client/DistributedLock/Controllers/BindingController.cs
```diff
@@ -0,0 +1,109 @@
...
+            
+            return Ok();
+        }
+
+
+        [Obsolete]
+        private async Task AttemptToProcessFile(string fileName)
+        {
+            // We don't need to lock for reading, so try that first to see if the object still exists before locking.
+            if (await GetFile(fileName) == null)
```
rynowak, line 56: This pattern is problematic, it's a TOCTOU https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use

You have to check null twice, or just do the check inside the lock.

rynowak: This is a better sample overall though.

examples/AspNetCore/SecretStoreConfigurationProviderSample/SecretStoreConfigurationProviderSample.csproj
halspang: I'm not sure, it looks like visual studio added something to that project file without me noticing. I'll remove it.

examples/Client/DistributedLock/1fa9b73d-36f1-4be1-80ff-034f28b65352
halspang: Yup, that's a test file that got missed cause I moved it out of the appropriate place :) It will be removed.

examples/Client/DistributedLock/Controllers/BindingController.cs
halspang: Yeah, I did that on purpose to show that once you're inside the lock you're safe.

I do agree though that it's not a good practice, it was forced to demo the lock. If you think it just promotes bad practices I'll go ahead and remove it.

halspang: Chatted offline, removed it to avoid showing people poor practices :)

rynowak (approved):

Issue #788, opened 2021-11-23
add support for TTL and time/interval formats in actor timers/reminders


dmitsh:
This feature has been implemented in Dapr runtime: dapr/dapr#2906

It needed to be available via the SDK.

PR #912, opened 2022-07-22
Add Actor Reminder/Timer TTL support

halspang:
# Description

This commit adds the TTL field to Actor reminders/timers. This allows
reminders and timers to expire after a given TimeSpan instead of
having to be manually deleted.

https://github.com/dapr/dotnet-sdk/issues/788

## Issue reference

Please reference the issue this PR will close: #788 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/examples/Actor/ActorClient/Program.cs
+++ b/examples/Actor/ActorClient/Program.cs
@@ -14,6 +14,7 @@
 namespace ActorClient
 {
     using System;
+    using System.Threading;
     using System.Threading.Tasks;
     using Dapr.Actors;
     using Dapr.Actors.Client;
@@ -86,8 +87,8 @@ namespace ActorClient
             var res = await nonRemotingProxy.InvokeMethodAsync<MyData>("GetData");
 
             Console.WriteLine("Registering the timer and reminder");
-            await proxy.RegisterTimer();
-            await proxy.RegisterReminder();
+            await proxy.RegisterTimer(null);
+            await proxy.RegisterReminder(null);
             Console.WriteLine("Waiting so the timer and reminder can be triggered");
             await Task.Delay(6000);
 
@@ -100,6 +101,14 @@ namespace ActorClient
             Console.WriteLine("Deregistering reminder. Reminders are durable and would not stop until an explicit deregistration or the actor is deleted.");
             await proxy.UnregisterReminder();
 
+            Console.WriteLine("Registering reminder and Timer with TTL - The reminder will self delete after 10 seconds.");
+            await proxy.RegisterReminder(TimeSpan.FromSeconds(10));
+            await proxy.RegisterTimer(TimeSpan.FromSeconds(10));
+
+            // Track the reminder.
+            var timer = new Timer(async state => Console.WriteLine($"Received data: {await proxy.GetData()}"), null, TimeSpan.FromSeconds(5), TimeSpan.FromSeconds(5));
+            await Task.Delay(TimeSpan.FromSeconds(21));
+            await timer.DisposeAsync();
 
             Console.WriteLine("Creating a Bank Actor");
             var bank = ActorProxy.Create<IBankActor>(ActorId.CreateRandom(), "DemoActor");
--- a/examples/Actor/ActorClient/Program.cs
+++ b/examples/Actor/ActorClient/Program.cs
@@ -87,8 +87,8 @@ namespace ActorClient
             var res = await nonRemotingProxy.InvokeMethodAsync<MyData>("GetData");
 
             Console.WriteLine("Registering the timer and reminder");
-            await proxy.RegisterTimer(null);
-            await proxy.RegisterReminder(null);
+            await proxy.RegisterTimer();
+            await proxy.RegisterReminder();
             Console.WriteLine("Waiting so the timer and reminder can be triggered");
             await Task.Delay(6000);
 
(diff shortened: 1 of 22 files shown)


examples/Actor/ActorClient/Program.cs
```diff
@@ -86,8 +87,8 @@ public static async Task Main(string[] args)
             var res = await nonRemotingProxy.InvokeMethodAsync<MyData>("GetData");
 
             Console.WriteLine("Registering the timer and reminder");
-            await proxy.RegisterTimer();
-            await proxy.RegisterReminder();
+            await proxy.RegisterTimer(null);
+            await proxy.RegisterReminder(null);
```
rynowak, line 91: nit: if this is a default-valued parameter you don't have to pass null here.

src/Dapr.Actors/Runtime/Actor.cs
```diff
@@ -213,9 +214,10 @@ protected virtual Task OnPostActorMethodAsync(ActorMethodContext actorMethodCont
             string reminderName,
             byte[] state,
             TimeSpan dueTime,
-            TimeSpan period)
+            TimeSpan period,
+            TimeSpan? ttl = null)
```
rynowak, line 218: This is a breaking change. See: https://github.com/ltrzesniewski/dotnet-runtime/blob/master/docs/coding-guidelines/breaking-change-rules.md#signatures

This needs to be a new overload.

src/Dapr.Actors/Runtime/Actor.cs
```diff
@@ -263,13 +265,15 @@ protected Task UnregisterReminderAsync(string reminderName)
...
+        /// <param name="ttl">The <see cref="TimeSpan"/> that dictates when this timer will expire.</param>
         /// <returns>Returns IActorTimer object.</returns>
         public async Task<ActorTimer> RegisterTimerAsync(
             string timerName,
             string callback,
             byte[] callbackParams,
             TimeSpan dueTime,
-            TimeSpan period)
+            TimeSpan period,
+            TimeSpan? ttl = null)
```
rynowak, line 276: Same deal, this is a breaking change.

src/Dapr.Actors/Runtime/Actor.cs
rynowak: When you add the new overload, we probably can't make `ttl` optional, because that will cause an ambiguity.

src/Dapr.Actors/Runtime/ActorReminder.cs
```diff
@@ -34,13 +34,15 @@ public class ActorReminder : ActorReminderToken, IActorReminder
...
+        /// <param name="ttl">The time at which the reminder will expire.</param>
         public ActorReminder(
             string actorType,
             ActorId actorId,
             string name,
             byte[] state,
             TimeSpan dueTime,
-            TimeSpan period)
+            TimeSpan period,
+            TimeSpan? ttl = null)
```
rynowak, line 45: This is also a breaking change.

rynowak: These are the best references I know of if you want to learn more:

- https://github.com/ltrzesniewski/dotnet-runtime/blob/master/docs/coding-guidelines/breaking-changes.md
- https://github.com/ltrzesniewski/dotnet-runtime/blob/master/docs/coding-guidelines/breaking-change-rules.md
- https://github.com/ltrzesniewski/dotnet-runtime/blob/master/docs/coding-guidelines/breaking-change-definitions.md

rynowak: I accidentally linked to someone's fork, but those should all be valid.

src/Dapr.Actors/Runtime/ActorTimer.cs
```diff
@@ -35,14 +35,16 @@ public class ActorTimer : ActorTimerToken
...
         public ActorTimer(
             string actorType,
             ActorId actorId,
             string name,
             string timerCallback,
             byte[] data,
             TimeSpan dueTime,
-            TimeSpan period)
+            TimeSpan period,
+            TimeSpan? ttl = null)
```
rynowak, line 47: Also a breaking change :)

src/Dapr.Actors/Runtime/ReminderInfo.cs
```diff
@@ -25,11 +25,13 @@ internal struct ReminderInfo
...
             byte[] data,
             TimeSpan dueTime,
-            TimeSpan period)
+            TimeSpan period,
+            TimeSpan? ttl = null)
         {
             this.Data = data;
             this.DueTime = dueTime;
             this.Period = period;
+            this.Ttl = ttl;
```
rynowak, line 34: This is not a breaking change 😁 - internal type

src/Dapr.Actors/Runtime/TimerInfo.cs
```diff
@@ -34,14 +34,16 @@ public class TimerInfo
             string callback,
             byte[] state,
             TimeSpan dueTime,
-            TimeSpan period)
+            TimeSpan period,
+            TimeSpan? ttl = null)
```
rynowak, line 38: This is not a breaking change 👍 - internal type

examples/Actor/ActorClient/Program.cs
halspang: The ActorProxy complained about the interface having a default value and without that, the top level doesn't appear to accept the default parameter. I'll look again as maybe I was just being wrong, but that's the behavior I observed.

src/Dapr.Actors/Runtime/Actor.cs
halspang: Dang, I thought the default saved me from the breaking change. Thanks for letting me know! I'll fix this and apply throughout.

halspang: > When you add the new overload, we probably can't make ttl optional, because that will cause an ambiguity in overload resolution.

So this was a problem because null is semantically important here unfortunately. I didn't want to put in something like looking for a specific TimeSpan value made us remove the TTL because who knows what people want to put in (and if it's wrong I'd rather error). So I introduced a class that allows us to pass in a null while keeping a consistent API surface. Let me know what you think, it's generally internal because it felt weird to expose it.

rynowak (approved):

Issue #925, opened 2022-08-11
Support for .NET standard 2.0 and/or 2.1


rangers-globecar:
Is there any reason this project is not compiled against `netstandard2.0` ?   I checked all the dependencies, and they would all be compatible with `netstandard2.0`

this would enable us to use it with legacy (for example 4.8) application that are migrating....  to be able to breakout components-by-components....

halspang:
@robert-stratecglobal - Thanks for opening this issue! However, we have no plans to support `netstandard` and have turned it down in the past. We have limited resources here and want to limit our scope to match the supported versions of .NET. `netstandard` is past LTS (so is 5.0). Though we'll retain 5.0 since we already support it and can't remove it without a breaking change, we don't want to bring on `netstandard`.

WhitWaldo:
Closing this issue as it's been answered and there aren't any follow up questions.

PR #927, opened 2022-08-12
support netstandard2.0 and netstandard2.1

bobbyangers:
This PR is a suggestion.    Please review and comment;  I am glad to update or fix anything you see fit.

This suggestion would make `Dapr.Client` usable by a `net48` project;  making it available for migration project that need to update component-by-component

# Description

updated xunit and other references
Updated:
- `Dapr.Client`  -> `netstandard2.0;netstandard2.1`
- `Dapr.Actor`  -> `netstandard2.1`
- `Dapr.AspNetCore` -> `netcoreapp3.1;net5;`**net6**
- `Dapr.Actor.AspNetCore` -> `netcoreapp3.1;net5;`**net6**
- `Dapr.Client.Test` => **net48** `;netcoreapp3.1;net5;net6

Updated outdated/obsolete references:
- `<PackageReference Include="Google.Protobuf" Version="3.19.4" />`
- `<PackageReference Include="System.Net.Http.Json" Version="6.0.0" />`

Update nuget in test projects
- `<PackageReference Include="FluentAssertions" Version="6.7.0" />`
- `<PackageReference Include="Microsoft.AspNetCore.Mvc.Testing" Version="6.0.8" />`
- `<PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.3.0" />`
- `<PackageReference Include="Moq" Version="4.14.7" />`
- `<PackageReference Include="xunit" Version="2.4.2" />`
- `<PackageReference Include="xunit.runner.visualstudio" Version="2.4.5">`

Minor adjustments to the code to compile with lower level requirements.

Also some tests surrounding `Gprc` where decorated with `#precompile` directives to exclude them from `net48` scope.

## Issue reference

Please reference the issue this PR will close: #925 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


halspang (changes requested):
@bobbyangers - Thanks for opening this PR! Some of the dependency updates, we could take. However, we have no plans on support `netstandard2.0/2.1`.

We also tend to only build to the lowest common framework, in this case `netcoreapp3.1`. The clients have not used any non-backwards compatible language features so adding the extra binaries doesn't add anything for us here.

bobbyangers:
> @bobbyangers - Thanks for opening this PR! Some of the dependency updates, we could take. However, we have no plans on support `netstandard2.0/2.1`.
> 
> We also tend to only build to the lowest common framework, in this case `netcoreapp3.1`. The clients have not used any non-backwards compatible language features so adding the extra binaries doesn't add anything for us here.

@halspang as you can analyze from the chart provided by Microsoft [.NET Standard](https://docs.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-1); all the of the .NET core framework are based on `netstandard2.1`
image

After reviewing all the dependencies in this projects;  all of them are actually `netstandard2.1` (ie not .net core);  therefore the lowest common framework is in fact `netstandard2.1` and by the same token will be compatible with .net core 3, 3.1, 5.0, 6.0.

Please reconsider.

terrajobst:
@bobbyangers 

.NET Standard is a good choice for anyone that wants (or needs) to support multiple flavors of .NET, such as .NET Core and .NET Framework. From talking to @halspang it sounds like their plan is to only support .NET Core. In that case, targeting .NET Standard doesn't make much sense. Considering that ASP.NET Core went .NET Core only starting with ASP.NET Core 3, having Dapr on .NET Core only seems sensible to me.

What is your scenario that made your asking for a .NET Standard implementation here? Would you like to use Dapr on .NET Framework?

bobbyangers:
> @bobbyangers
> 
> .NET Standard is a good choice for anyone that wants (or needs) to support multiple flavors of .NET, such as .NET Core and .NET Framework. From talking to @halspang it sounds like their plan is to only support .NET Core. In that case, targeting .NET Standard doesn't make much sense. Considering that ASP.NET Core went .NET Core only starting with ASP.NET Core 3, having Dapr on .NET Core only seems sensible to me.
> 
> What is your scenario that made your asking for a .NET Standard implementation here? Would you like to use Dapr on .NET Framework?

first off, I'd like to use the `DaprClient` on .NET 4.8,  the project I am working is a monolith ASPNET MVC app, and we are migrating towards a microservice architecture, but it all cannot be done in "one-pass";   I'd like to choose some component extract it in a microservice, and leverage the `DaprClient` to bind the new component.    Since technically the `DaprClient` is a specialised `HttpClient`... could we consider separating that functionality in it's own NuGet package?  which would target the lowest common framework ?

As stated earlier, since all the dependencies are themselves `netstandard2.0` or `netstandard2.1`;  and there is no technical necessity to set the minimum .netcore 3.x;   .netcore is compiler compatible with .net5, .net6.. it's not native.

Of course the test projects would target `netcore3.1`, `net5`*, `net6`, and soon `net7`;  but that is pretty common scenario.

What is the motivation to be `netcore3.1` which is on it's way out, and rather support a simpler plateform (ie netstandard) that is more future proof, and able to reach a wider audience ?

Event the "inbound/outbound" bindings can be implemented in the `netstandard2.1`.

My suggestion is that the `DaprClient` itself be compiled separately targeting both `netstandard2.0` and `netstandard2.1`, to be available to a wider audience (ie starting `net46`) and the more advance features for the 'inbound' bindings in it's own package targeting `netstandard2.1` which is the base for `netcore` and up.

n.b.  for the moment, I have forked the project and using my in-house version which does just what I described above.

terrajobst:
> What is the motivation to be `netcore3.1` which is on it's way out, and rather support a simpler plateform (ie netstandard) that is more future proof, and able to reach a wider audience ?

I don't work on Dapr, so @halspang would have to give the authoritative answer with respect to their motivation.

I do, however, work on .NET Standard so I can talk to that in great detail. The way to think about .NET Standard is that it is just an API set. The standard itself isn't an implementation of the APIs which is why we don't subject it to the same lifetime policies as we do for .NET Core or .NET Framework. But .NET Standard is similar to .NET Core in that it is a versioned API set. Since .NET Standard is implemented by .NET Framework 4.6.1 and .NET Core 2.0 you can think of either has a strict superset of .NET Standard 2.0, which in practical terms is akin to "a later version of .NET Standard 2.0". Said differently, .NET Core 3.1 has more APIs than .NET Standard 2.0. We [already announced there won't be a new version of .NET Standard](https://devblogs.microsoft.com/dotnet/the-future-of-net-standard/) and we consider .NET 5 and later versions as the spiritual successor of .NET Standard. In that vein I wouldn't say that .NET Standard is more future proof than targeting .NET Core 3.1; either one can easily be upgraded to .NET 5, 6, 7 and so on.

.NET Standard does, however, give a library developer more reach. But if the Dapr doesn't want to support it on anything but .NET Core then reach doesn't matter. In contrast, .NET Core 3.1 or later has more APIs than .NET Standard which allows Dapr to more easily take advantage of the latest features.

So in the end choosing between .NET Standard and a specific .NET implementation is the trade-off between reach and feature set. And it seems the Dapr folks are prioritizing feature set over reach, which is also what ASP.NET Core have done.

Does this make sense?

bobbyangers:
@terrajobst 

I understand your explanation.   I respect your choices.

As for `DaprClient`, which doesn't necessarily need the possible advanced features, it's a specialized `HttpClient`;  `HttpClient` is an api/executable code; it make sense to me to package it separately in a more accessible to a wider audience (as `Micrsoft.Extensions.Http` which is `netstandard`). 

For the moment, I am going to go back to the basics and implement and use my own client for my migration project.

Thanks for your explanation and your time.

terrajobst:
You're welcome. And thank you for all the work you have put into this PR!

Issue #904, opened 2022-07-08
Exception from Actor's method is not surfaced to the caller


artursouza:
## Expected Behavior

Exception's error message (not stacktrace) from Actor's method is surfaced to the caller.

## Actual Behavior

![image](https://user-images.githubusercontent.com/130954/177892380-af0c80ab-f5b7-41e3-b82e-2bf195fbf9c0.png)

## Steps to Reproduce the Problem

See example from [actor exception](https://github.com/fabistb/dotnet-dapr-actor-exception).

## Release Note

RELEASE NOTE: **FIXED** issue that was not surfacing actor error to the caller's app

artursouza:
@amulyavarote @halspang Please, consider ramifications of this change to the runtime as well. We might need to agree on a metadata/header to propagate the error message. In addition, consider truncating error message from the Actor SDK to avoid header size limit issues, as pointed out by @berndverst.

halspang:
As an alternative approach, would something like additional logging at the time of error be sufficient here? It seems like the issue is we raise the exception and it's tossed across enough areas that we lose the stack. However, we could just dump the stack trace when we catch the first exception and then it requires no change to the runtime. Thoughts @artursouza?

rynowak:
@artursouza can you provide a more detailed explanation? It's not super clear to me from the screenshot what is supposed to be different. 

If this is about serializing .NET exception types over the wire, we deliberately avoid that for security reasons. It's very unsafe. 

If this is about serializing .NET call stacks over the wire, we deliberately also avoid that for security reasons.

We should make sure the call stacks and exception info are being logged visibly inside the process that throws the exception.

berndverst:
@rynowak I think the request is essentially that an ExceptionName and Message are returned to Dapr Actor Runtime so that the callee can see what error occurred in the Actor Method. Feels a bit weird, but that is my understanding.

In JSON I sometimes see things like:

```
{
 "status": "200",
 "errormsg": "Error text here or empty"
 "response": "Response text on success only"
}
```

Of course changing what the .NET SDK returns upon encountering an exception in the Actor case would break some people.. so this doesn't seem like a good idea.

As long as we enforce a reasonably small message size limit (e.g. 256 characters) it might be ok to simply return the error message via gRPC metadata / http header.

At the end of the day, this all still feels wrong to me. Proper error logging on the actor service implementations should be added, and the error logs should be examined that way.

halspang:
[...]

PR #931, opened 2022-08-18
Added exception details to the actor invocation

amulyavarote:
Signed-off-by: Amulya Varote <amulyavarote@microsoft.com>

# Description

Added a method that gives line number and the method name for the actor invocation.

## Issue reference

Please reference the issue this PR will close: #904 
## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/src/Dapr.Actors.AspNetCore/ActorsEndpointRouteBuilderExtensions.cs
+++ b/src/Dapr.Actors.AspNetCore/ActorsEndpointRouteBuilderExtensions.cs
@@ -19,6 +19,7 @@ using Dapr.Actors.Runtime;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Routing;
 using Microsoft.Extensions.DependencyInjection;
+using System.Diagnostics;
 
 namespace Microsoft.AspNetCore.Builder
 {
@@ -189,6 +190,7 @@ namespace Microsoft.AspNetCore.Builder
     {
         var responseHeader = new ActorResponseMessageHeader();
         responseHeader.AddHeader("HasRemoteException", Array.Empty<byte>());
+        responseHeader.AddHeader(GetExceptionInfo(ex), Array.Empty<byte>());
         var headerSerializer = new ActorMessageHeaderSerializer();
         var responseHeaderBytes = headerSerializer.SerializeResponseHeader(responseHeader);
         var serializedHeader = Encoding.UTF8.GetString(responseHeaderBytes, 0, responseHeaderBytes.Length);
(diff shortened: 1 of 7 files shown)


src/Dapr.Actors/DaprHttpInteractor.cs
```diff
@@ -188,6 +191,18 @@ HttpRequestMessage RequestFunc()
...
+            XmlNodeList exceptionValueXML = xmlHeader.GetElementsByTagName(EXCEPTION_HEADER_TAG);
+            string exceptionDetails = "";
+            if (exceptionValueXML != null && exceptionValueXML[1] != null)
+            {
+                exceptionDetails = exceptionValueXML[1].InnerText;
+            }
+            return exceptionDetails;
+        }
+
         public async Task<Stream> InvokeActorMethodWithoutRemotingAsync(string actorType, string actorId, string methodName, string jsonPayload, CancellationToken cancellationToken = default)
```
halspang, line 206: Have you validated that this works for non-remoting actors as well?

src/Dapr.Actors.AspNetCore/ActorsEndpointRouteBuilderExtensions.cs
```diff
@@ -189,6 +190,7 @@ private static IEndpointConventionBuilder MapActorHealthChecks(this IEndpointRou
     {
         var responseHeader = new ActorResponseMessageHeader();
         responseHeader.AddHeader("HasRemoteException", Array.Empty<byte>());
+        responseHeader.AddHeader(GetExceptionInfo(ex), Array.Empty<byte>());
```
halspang, line 193: Is this sending the exception info as the header name, instead of the value?

src/Dapr.Actors/DaprHttpInteractor.cs
```diff
@@ -133,14 +136,13 @@ HttpRequestMessage RequestFunc()
...
-
+            var header = "";
             IActorResponseMessageHeader actorResponseMessageHeader = null;
             if (retval != null && retval.Headers != null)
             {
                 if (retval.Headers.TryGetValues(Constants.ErrorResponseHeaderName, out IEnumerable<string> headerValues))
                 {
-                    var header = headerValues.First();
-
+                    header = headerValues.First();
```
halspang, line 145: We should be getting the header info by it's name since we know there's more than one now.

halspang (changes requested):
I did check, and changing the wording of an exception is _not_ considered a breaking change :)

src/Dapr.Actors.AspNetCore/ActorsEndpointRouteBuilderExtensions.cs
```diff
@@ -198,6 +200,20 @@ private static IEndpointConventionBuilder MapActorHealthChecks(this IEndpointRou
         return new Tuple<string, byte[]>(serializedHeader, responseMsgBody);
     }
 
+    /// <summary>
+    /// Generate exception info
+    /// </summary>
+    /// <param name="ex">Exception of the method.</param>
+    /// <returns>Exception info string</returns>
+    private static string GetExceptionInfo(Exception ex) {
```
halspang, line 208: You should encode this as a Base64 string and then return the `Array<byte>` value for the header.

src/Dapr.Actors.AspNetCore/ActorsEndpointRouteBuilderExtensions.cs
```diff
@@ -189,6 +190,7 @@ private static IEndpointConventionBuilder MapActorHealthChecks(this IEndpointRou
     {
         var responseHeader = new ActorResponseMessageHeader();
         responseHeader.AddHeader("HasRemoteException", Array.Empty<byte>());
+        responseHeader.AddHeader("RemoteMethodException", Encoding.ASCII.GetBytes(GetExceptionInfo(ex)));
```
halspang, line 193: This is just a byte conversion, you didn't appear to do a base64 encoding first.

src/Dapr.Actors.AspNetCore/ActorsEndpointRouteBuilderExtensions.cs
```diff
@@ -198,6 +200,20 @@ private static IEndpointConventionBuilder MapActorHealthChecks(this IEndpointRou
...
+    /// Generate exception info
+    /// </summary>
+    /// <param name="ex">Exception of the method.</param>
+    /// <returns>Exception info string</returns>
+    private static string GetExceptionInfo(Exception ex) {
+        var frame = new StackTrace(ex, true).GetFrame(0);
+        string methodName = frame.GetMethod().Name;
+        int lineNumber = frame.GetFileLineNumber();
+        string exceptionName = ex.GetType().Name;
+        string uuid = Guid.NewGuid().ToString();
```
halspang, line 213: nit: These could be vars, or even just put directly into the string below.

halspang (changes requested):
Thinking on it, this is probably worth an e2e test to validate the exception behavior. Can we create one?

Issue #934, opened 2022-08-22
Streaming configuration - updates are not propagated


tomhreb:
## Expected Behavior

Using IOptionsMonitor<T> I can receive updates from dapr configuration store.

## Actual Behavior

I am not receiving updates from the configuration store.

## Steps to Reproduce the Problem

This can be reproduced using minimal api app

Project file
```xml
<Project Sdk="Microsoft.NET.Sdk.Web">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net6.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.Extensions.Hosting" Version="3.1.8" />
  </ItemGroup>

  <ItemGroup>
    <ProjectReference Include="..\src\Dapr.Extensions.Configuration\Dapr.Extensions.Configuration.csproj" />
  </ItemGroup>

</Project>
```

Program.cs
```csharp
using Dapr.Client;
using Dapr.Extensions.Configuration;
using Microsoft.Extensions.Options;

var client = new DaprClientBuilder().Build();
var builder = WebApplication.CreateBuilder(args);
var services = builder.Services;
var configuration = builder.Configuration;

configuration.AddDaprConfigurationStore("redisConfig", new List<string> { "version" }, client, TimeSpan.FromSeconds(20));
configuration.AddStreamingDaprConfigurationStore("redisConfig", new List<string> { "version" }, client, TimeSpan.FromSeconds(20));

services.Configure<MyOptions>(builder.Configuration);
var app = builder.Build();

app.MapGet("/", context =>
{
    var config = context.RequestServices.GetRequiredService<IOptionsMonitor<MyOptions>>();

    return context.Response.WriteAsJsonAsync(config.CurrentValue.Version);
});

app.Run();

class MyOptions
{
    public string Version { get; set; }
}
```

By periodically quering this endpoint while updating the key in redis, the response shculd report updated value, which is not happening and the first value is reported.

EDIT: I've found the root cause and created  #935 
```

RELEASE NOTE: Fixed `DaprConfigurationStoreProvider` not applying case-insensitive configuration updates correctly.

WhitWaldo:
Closing this issue as the bugfix has been merged

Issue #887, opened 2022-06-06
Error Suppression and Vault Name Mapping 


BCoskun:
## Describe the proposal
We are working on Azure Key Vault in our project and notice that there is no error suppression functionality (or not granular enough for certain use case) when there the Key Vault Secret name doesn't have match with application configuration name. 

For example; 
We couldn't define .NET configuration hierarchy  in Azure KeyVault due to naming convention limitation for secret names. Key Vault only accepts dash and Alphanumeric values for the names so we can't use "ConnectionStrings:DatabaseConnStr" or "ConnectionStrings__DatabaseConnStr" as name. 

**Purposed solution**
Allocate special metadata prefix like "Dapr." in metadata keys and use them to control library behavior.

**Example Solution in Library**

``` C#

LoadAsync()  method In  DaprSecretStoreConfigurationProvider.cs file
...

    if (secretDescriptors != null)
            {
                foreach (var secretDescriptor in secretDescriptors)
                {
                    var vaultSecretName = String.Empty; 
                        secretDescriptor.Metadata.TryGetValue("Dapr.VaultMapping.Name", out vaultSecretName);

                        if (!String.IsNullOrWhiteSpace(vaultSecretName))
                            secretName = vaultSecretName;

                        Dictionary<string, string> result =  new Dictionary<string, string>();
                        
                        try
                        {
                            result = await client
                                .GetSecretAsync(store, secretName, secretDescriptor.Metadata)
                                .ConfigureAwait(false);
                        }
                        catch (DaprException e)
                        {
                            string? ensure_exist;
                            secretDescriptor.Metadata.TryGetValue("Dapr.Ensure", out ensure_exist);
                            if (String.IsNullOrWhiteSpace(ensure_exist))
                                ensure_exist = "false";
                            
                            if (Convert.ToBoolean(ensure_exist))
                            {
                                throw e;
                            }
                            
                            Console.WriteLine($"'{secretName}' doesn't exists in Key Vault but because it doesn't require existence so all errors suppressed!");
                           result = new Dictionary<string, string>();

                        }

                        foreach (var key in result.Keys)
                        {
                            if (data.ContainsKey(key))
                            {
                                throw new InvalidOperationException(
                                    $"A duplicate key '{key}' was found in the secret store '{store}'. Please remove any duplicates from your secret store.");
                            }
                            
                            data.Add(normalizeKey ? NormalizeKey(secretDescriptor.SecretName) : secretDescriptor.SecretName, result[key]);
[...]

PR #935, opened 2022-08-23
preserve comparer of the original dictionary

tomhreb:
Signed-off-by: Tomas Hrebicek <tomhreb@users.noreply.github.com>

# Description

Fixes issue #934 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


halspang:
@tomhreb - Thanks for opening this! I'm a little bit worried about removing case sensitivity though. What if people are depending on that?

I think this may be more what you're looking for #887?

tomhreb:
@halspang No, I don't think #887 is related.

Without the case sensitivity, this functionality does not work. The original dictionary in the .net core library has this case sensitivity. I still regard this as a bugfix.

tomhreb:
Hi @halspang, any idea if this will be approved?

Check that the base class in .net SDK really has ```StringComparer.OrdinalIgnoreCase```.

https://github.com/dotnet/dotnet/blob/b8bc661a7429baa89070f3bee636b7fbc1309489/src/runtime/src/libraries/Microsoft.Extensions.Configuration/src/ConfigurationProvider.cs

WhitWaldo:
I agree that this is a bugfix. I've flagged it as a breaking change as it might indeed break some people that are using it, but I agree that the code in `DaprConfigurationStoreProvider` is flawed here in that it's taking a `Dictionary<string,string>(StringComparer.OrdinalIgnoreCase)` and then doing update operations on a `Dictionary<string,string>()` that it's then setting the original implementation to which is not right.

@tomhreb Once all the tests finish running against it, I'll merge this. Thank you for your contribution!

WhitWaldo (approved):
Solid bugfix. Looks good to me!

Issue #486, opened 2020-11-22
Migrate calls to Dapr's sidecar Actor APIS to GRPC


artursouza:
Applications can use GRPC (in addition to HTTP) to make API calls for Actor.
This impacts Actor client and Actor runtime.
A property or environment variable can be used to switch between HTTP or GRPC communication.
GRPC communication is the default.

RELEASE NOTE: Support GRPC communication for Actor APIs into Dapr.

artursouza:
Actor Client making HTTP calls: https://github.com/dapr/dotnet-sdk/blob/master/src/Dapr.Actors/Communication/Client/ActorNonRemotingClient.cs

Actor Proxy, using the HTTP client: https://github.com/dapr/dotnet-sdk/blob/fe1e697af3f8c0983bb587b177581ee65c7ecfe5/src/Dapr.Actors/Client/ActorProxy.cs#L110

berndverst:
@halspang please assign to @addjuarez. Priority P1, for upcoming v1.9 release. FYI @artursouza 

This is related to https://github.com/dapr/dapr/issues/5118

adamreed90:
Curious of the status of this? :D

PR #954, opened 2022-09-21
Migrate actors to use grpc

addjuarez:
Signed-off-by: addjuarez <addiajuarez@gmail.com>

# Description

Migrate .net actor implementation to use grpc by default.

## Issue reference

Please reference the issue this PR will close: #486 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


src/Dapr.Actors/DaprGrpcInteractor.cs
```diff
@@ -0,0 +1,364 @@
...
+            {
+                throw new DaprApiException("GetActorState operation failed: the Dapr endpoint indicated a failure. See InnerException for details.", ex);
+            }
+            return response.Data.ToStringUtf8();
+        }
+
+        public Task SaveStateTransactionallyAsync(string actorType, string actorId, string data, CancellationToken cancellationToken = default)
+        {
+            return null;
+        }
```
halspang, line 84: Why did we decide to do this and create a new method instead of just using the gRPC client here?

src/Dapr.Actors/DaprHttpInteractor.cs
```diff
@@ -75,6 +76,11 @@ HttpRequestMessage RequestFunc()
             return stringResponse;
         }
 
+        public Task SaveStateTransactionallyAsyncGrpc(string actorType, string actorId, List<Autogenerated.TransactionalActorStateOperation> data, CancellationToken cancellationToken = default)
```
halspang, line 79: As commented above, I don't think we need to distinguish this.

src/Dapr.Actors/DaprInteractorBuilder.cs
```diff
@@ -0,0 +1,109 @@
...
+{
+    using System;
+    using System.Net.Http;
+    using Grpc.Net.Client;
+    using Autogenerated = Dapr.Client.Autogen.Grpc.v1;
+
+    /// <summary>
+    /// Builder for building <see cref="DaprGrpcInteractor"/>
+    /// </summary>
+    public sealed class DaprInteractorBuilder
```
halspang, line 24: Any reason we made this a builder? Also, since it seems generic, if we want to keep it, it should be able to provide an http or grpc one.

src/Dapr.Actors/IDaprInteractor.cs
```diff
@@ -44,6 +46,16 @@ internal interface IDaprInteractor
...
 
+        /// <summary>
+        /// Saves state batch to Dapr.
+        /// </summary>
+        /// <param name="actorType">Type of actor.</param>
+        /// <param name="actorId">ActorId.</param>
+        /// <param name="data">JSON data with state changes as per the Dapr spec for transaction state update.</param>
+        /// <param name="cancellationToken">Cancels the operation.</param>
+        /// <returns>A task that represents the asynchronous operation.</returns>
+        Task SaveStateTransactionallyAsyncGrpc(string actorType, string actorId, List<Autogenerated.TransactionalActorStateOperation> data, CancellationToken cancellationToken = default);
```
halspang, line 57: As stated above, I don't think we need this.

src/Dapr.Actors/Runtime/ActorRuntime.cs
```diff
@@ -50,7 +50,16 @@ internal ActorRuntime(ActorRuntimeOptions options, ILoggerFactory loggerFactory,
...
             foreach (var actor in options.Actors)
             {
-                var daprInteractor = new DaprHttpInteractor(clientHandler: null, httpEndpoint: options.HttpEndpoint, apiToken: options.DaprApiToken, requestTimeout: null);
+                IDaprInteractor daprInteractor;
+                if (options.useGrpc) {
+                daprInteractor = new DaprInteractorBuilder()
+                    .UseGrpcEndpoint(options.GrpcEndpoint)
+                    .UseDaprApiToken(options.DaprApiToken)
+                    .UseGrpcChannelOptions(options.GrpcChannelOptions)
+                    .Build();
```
halspang, line 59: Needs an indent but I also feel it shouldn't use a builder.

If you do want to use a builder, I think it should take in the `options.UseGrpc` field and return the appropriate interactor. The benefit of that being an interface is that we don't need to know the actual implementation.

src/Dapr.Actors/Runtime/ActorRuntimeOptions.cs
```diff
@@ -225,5 +226,26 @@ public string DaprApiToken
...
+        /// <c>http://127.0.0.1:DAPR_GRPC_PORT</c> where <c>DAPR_GRPC_PORT</c> represents the value of the 
+        /// <c>DAPR_GRPC_PORT</c> environment variable.
+        /// </remarks>
+        /// <value></value>
+        public string GrpcEndpoint { get; set; } = DaprDefaults.GetDefaultGrpcEndpoint();
+
+        /// <summary>
+        /// Option to use GRPC or HTTP
+        /// </summary>
+        public bool useGrpc { get; set; } = true;
```
halspang, line 244: `useGrpc` -> `UseGrpc`

src/Dapr.Actors/Runtime/DaprStateProvider.cs
```diff
@@ -77,10 +80,60 @@ public async Task<bool> ContainsStateAsync(string actorType, string actorId, str
 
         public async Task SaveStateAsync(string actorType, string actorId, IReadOnlyCollection<ActorStateChange> stateChanges, CancellationToken cancellationToken = default)
         {
-            await this.DoStateChangesTransactionallyAsync(actorType, actorId, stateChanges, cancellationToken);
+            if (this.daprInteractor.GetType().Name.Equals("DaprGrpcInteractor")){
+                await this.DoStateChangesTransactionallyAsyncGrpc(actorType, actorId, stateChanges, cancellationToken);
+            } else {
+                await this.DoStateChangesTransactionallyAsyncHttp(actorType, actorId, stateChanges, cancellationToken);
+            }
```
halspang, line 87: The interface of the interactor makes this unnecessary. Whatever the underlying object is, we'll use its method. So no need to separate them like this.

src/Dapr.Actors/Runtime/TimerInfo.cs
```diff
@@ -24,7 +24,6 @@ namespace Dapr.Actors.Runtime
     /// <summary>
     /// Represents the details of the timer set on an Actor.
     /// </summary>
-    [Obsolete("This class is an implementation detail of the framework and will be made internal in a future release.")]
```
halspang, line 27: Did we make this internal? I don't think we did. And if we did, that's a breaking change so we shouldn't just do it. Any reason why you removed this?

src/Dapr.Actors/Runtime/DaprStateProvider.cs
```diff
@@ -77,10 +80,60 @@ public async Task<bool> ContainsStateAsync(string actorType, string actorId, str
...
+        }
+
+        private async Task DoStateChangesTransactionallyAsyncGrpc(string actorType, string actorId, IReadOnlyCollection<ActorStateChange> stateChanges, CancellationToken cancellationToken = default)
```
halspang, line 90: Same comments, we can remove this.

halspang (changes requested):

src/Dapr.Actors/Runtime/DaprStateProvider.cs
```diff
@@ -77,10 +80,56 @@ public async Task<bool> ContainsStateAsync(string actorType, string actorId, str
 
         public async Task SaveStateAsync(string actorType, string actorId, IReadOnlyCollection<ActorStateChange> stateChanges, CancellationToken cancellationToken = default)
         {
-            await this.DoStateChangesTransactionallyAsync(actorType, actorId, stateChanges, cancellationToken);
+            await this.daprInteractor.DoStateChangesTransactionallyAsync(this, actorType, actorId, stateChanges, cancellationToken);
         }
 
-        private async Task DoStateChangesTransactionallyAsync(string actorType, string actorId, IReadOnlyCollection<ActorStateChange> stateChanges, CancellationToken cancellationToken = default)
+        public async Task DoStateChangesTransactionallyAsyncGrpc(string actorType, string actorId, IReadOnlyCollection<ActorStateChange> stateChanges, CancellationToken cancellationToken = default)
```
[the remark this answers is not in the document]
addjuarez: For this method we do. The interface wants a ActorStateChange objects but grpc requires a proto object here.

addjuarez:
> How hard would it be to replicate any tests that the `DaprHttpInteractor` has for the `DaprGrpcInteractor`? It would also be good to make a base e2e test for the actors to ensure this works correctly.

I can make e2e tests for the grpcInteractor but the unit tests for grpc are a bit more complex.

src/Dapr.Actors/DaprHttpInteractor.cs
```diff
@@ -75,7 +77,7 @@ HttpRequestMessage RequestFunc()
             return stringResponse;
         }
 
-        public Task SaveStateTransactionallyAsync(string actorType, string actorId, string data, CancellationToken cancellationToken = default)
+        public Task SaveStateTransactionallyAsync(string actorType, string actorId, string data, CancellationToken cancellationToken = default, List<Autogenerated.TransactionalActorStateOperation> grpcData = default)
```
halspang, line 80: Is `grpcData` used?

src/Dapr.Actors/DaprHttpInteractor.cs
```diff
@@ -93,6 +95,11 @@ HttpRequestMessage RequestFunc()
             return this.SendAsync(RequestFunc, relativeUrl, cancellationToken);
         }
 
+        public Task DoStateChangesTransactionallyAsync(DaprStateProvider provider, string actorType, string actorId, IReadOnlyCollection<ActorStateChange> stateChanges, CancellationToken cancellationToken = default)
+        {
+            return provider.DoStateChangesTransactionallyAsyncHttp(actorType, actorId, stateChanges, cancellationToken);
```
halspang, line 100: I thought we wouldn't need to split them up between http and gRPC anymore.

src/Dapr.Actors/Runtime/DaprStateProvider.cs
halspang, line 86: Moving this logic into the `daprGrpcInteractor` would remove the need to distinguish between http/grpc right?

halspang (changes requested):

src/Dapr.Actors/DaprHttpInteractor.cs
addjuarez: Yes the grpcInteractor uses this for its data. It doesnt use the string argument

WhitWaldo:
I'm going to go ahead and close this issue for now in light of the new proposal at the runtime for formally adding (bidirectional streaming)[https://github.com/dapr/proposals/pull/72] and [PubSub support](https://github.com/dapr/proposals/pull/74) on a new gRPC actor implementation.

PR #955, opened 2022-09-22
remove .NET5 (out of support runtime) from test projects

heavenwing:
# Description

remove .NET5 TargetFrameworks from test projects, because .NET5 is out of support runtime.

## Issue reference

N/A

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


halspang:
@heavenwing - Thanks for this PR! However, we'd want to give more warning about removing a build target. These are also only test targets, so we're making sure we still support that version.

Instead of just the PR, can you create an issue and we can prioritize it through the standard milestone process?

halspang:
I actually ended up doing this in https://github.com/dapr/dotnet-sdk/pull/1005.

PR #961, opened 2022-09-29
Support custom subscriptions response.

maliming:
# Description

This PR supports custom subscriptions response. Mainly to integrate [Event-Bus](https://docs.abp.io/en/abp/latest/Event-Bus) in the [ABP framework](https://github.com/abpframework/abp).

## Issue reference

No

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


maliming:
hi @halspang

Do you have time to review this PR?

halspang:
Thanks for opening this PR, but I'm not really sure what the point of this change is. Can you provide some more background on it?

maliming:
Thanks @halspang 

The `dapr/subscribe` endpoint gets all endpoints that use `TopicAttribute` and returns JSON data.

This PR allows developers to change the subscription response, Such as hiding or dynamically adding some subscriptions.

This may not be commonly used, but it is convenient for some frameworks to integrate Dapr.

halspang:
@maliming - Thanks. However, it seems like `WithTopic` would do this, right? And any additional integrations would need to come from the runtime since this is a json that we return to the runtime, we need to adhere to its contract. Or am I missing something here?

maliming:
Yes, You're right.

We also have some concepts similar to publish/subscribe in our class library, and with the current design, we need to create a lot of controllers and methods for them and use `TopicAttribute`.

We want to create a controller and dynamically specify multiple topic/route.

gdlcf88:
Hey guys, any update? 🌹

WhitWaldo:
Hello @maliming and thank you for your contribution!

I believe the new dynamic PubSub subscriptions capability introduced in 1.14 covers the idea behind your contribution - this will receive .NET SDK support with 1.15 via the PR at #1346.

If you disagree, please feel free to re-open and we can continue to discuss it.

maliming:
Thanks @WhitWaldo 👍  , I will check the #1346

Issue #979, opened 2022-11-03
Add initial Dapr workflow authoring APIs


amulyavarote:
## Describe the proposal

Add workflow authoring APIs to the dotnet sdk based on the initial proposal for dapr workflows for dotnet-sdk. 
Linking the issue here.

PR #973, opened 2022-10-21
[WIP] Dapr Workflow initial methods for dotnet-sdk

amulyavarote:
# Description

This is a work in progress PR for dapr workflow initial methods for dotnet-sdk.
This needs e2e tests still.

## Issue reference

We strive to have all PR being opened based on an issue, where the problem or feature have been discussed prior to implementation.

Please reference the issue this PR will close: #979 

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [x] Code compiles correctly
* [ ] Created/updated tests
* [ ] Extended the documentation

PR #981, opened 2022-11-18
Workflow Authoring: initial methods

tmacam:
# Description

(This PR is a re-submission of PR dapr/dotnet-sdk#973 with extra fixes when appropriate)

This is a work in progress PR for dapr workflow initial methods for dotnet-sdk.
This needs e2e tests still.

## Issue reference

Please reference the issue this PR will close: #979

## Checklist

* [X] Code compiles correctly
* [ ] Created/updated tests


cgillum:
Hi @tmacam, can you please accept my PR [here](https://github.com/tmacam/dapr-dotnet-sdk/pull/1), which will update the Microsoft.DurableTask.* dependencies in this PR to v1.0.0-rc.1? There were several breaking changes and important bug fixes in that release, and we should make sure that those changes are reflected in this new Dapr Workflow SDK.

src/Dapr.Workflow/WorkflowServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,63 @@
...
+    /// Contains extension methods for using Dapr Workflow with dependency injection.
+    /// </summary>
+    public static class WorkflowServiceCollectionExtensions
+    {
+        /// <summary>
+        /// Adds Dapr Workflow support to the service collection.
+        /// </summary>
+        /// <param name="serviceCollection">The <see cref="IServiceCollection"/>.</param>
+        /// <param name="configure">A delegate used to configure actor options and register workflow functions.</param>
+        public static IServiceCollection AddWorkflow(
```
philliphoff, line 31: `AddDaprWorkflow()`?

src/Dapr.Workflow/WorkflowMetadata.cs
```diff
@@ -0,0 +1,44 @@
...
+using Microsoft.DurableTask.Client;
+
+namespace Dapr.Workflow
+{
+    /// <summary>
+    /// Represents a snapshot of a workflow instance's current state, including metadata.
+    /// </summary>
+    public class WorkflowMetadata
+    {
+        internal WorkflowMetadata(OrchestrationMetadata? metadata)
```
philliphoff, line 23: Nit: `details` instead of `metadata`?

examples/Workflow/WorkflowWebApp/Properties/launchSettings.json
```diff
@@ -0,0 +1,12 @@
+{
```
philliphoff, line 1: Should this be in the repo? (I was asked to remove it from samples in another repo.)

examples/Workflow/WorkflowWebApp/WorkflowWebApp.csproj
```diff
@@ -0,0 +1,15 @@
...
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\..\src\Dapr.Workflow\Dapr.Workflow.csproj" />
+  </ItemGroup>
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net6.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <LangVersion>latest</LangVersion>
```
philliphoff, line 11: `10.0`?  (As .NET 6 is compatible only with C# <=10.)

examples/Workflow/WorkflowWebApp/Program.cs
```diff
@@ -0,0 +1,51 @@
...
+    {
+        return Results.Ok(metadata);
+    }
+    else
+    {
+        return Results.NotFound($"No workflow with ID = '{id}' was found.");
+    }
+}).WithName("GetWorkflowEndpoint");
+
+app.Run("http://0.0.0.0:8080");
```
philliphoff, line 50: Nit: perhaps choose a default other than `8080`, as that could conflict with Dapr's own dashboard?

src/Dapr.Workflow/Dapr.Workflow.csproj
```diff
@@ -0,0 +1,20 @@
...
+    <Title>Dapr Workflow Authoring SDK</Title>
+    <Description>Dapr Workflow SDK for building workflows as code with Dapr</Description>
+    <VersionPrefix>0.1.0</VersionPrefix>
+    <VersionSuffix>alpha</VersionSuffix>
+  </PropertyGroup>
+  
+  <ItemGroup>
+    <PackageReference Include="Microsoft.DurableTask.Client.Grpc" Version="1.0.0-rc.1" />
+    <PackageReference Include="Microsoft.DurableTask.Worker.Grpc" Version="1.0.0-rc.1" />
+    <PackageReference Include="Dapr.AspNetCore" Version="1.8.0" />
```
philliphoff, line 17: Nit: `1.9.0`?

tmacam: Yes, support for pluggable components only appears on 1.9+

examples/Workflow/WorkflowWebApp/Program.cs
tmacam: Not sure about dapr dashboard but, for instance, examples and [quickstarts](https://docs.dapr.io/getting-started/quickstarts/) tend to stay away from this port. Might be worth using a different one just to be on the safe side.

examples/Workflow/WorkflowWebApp/WorkflowWebApp.csproj
tmacam: Your call. As I understand we have a bit more freedom on choosing newer SDKs/language versions for pluggable components than on the regular SDKs

src/Dapr.Workflow/WorkflowServiceCollectionExtensions.cs
cgillum: I could go either way, but the Actors SDK names the equivalent extension method `AddActors()` (no "Dapr") so `AddDaprWorkflow` would be a bit inconsistent with the actor SDK.

src/Dapr.Workflow/WorkflowContext.cs
```diff
@@ -0,0 +1,78 @@
...
+        {
+            return this.innerContext.CreateTimer(delay, cancellationToken);
+        }
+
+        /// <summary>
+        /// Waits for an event to be raised with name <paramref name="eventName"/> and returns the event data.
+        /// </summary>
+        public Task<T> WaitForExternalEventAsync<T>(string eventName, TimeSpan timeout)
+        {
+            return this.innerContext.WaitForExternalEvent<T>(eventName, timeout);
```
philliphoff, line 67: Does this inner context not have an overload that accepts a cancellation token?  (I expect not, given the timeout, but just checking.)

examples/Workflow/WorkflowWebApp/Program.cs
```diff
@@ -0,0 +1,51 @@
...
+// The workflow host is a background service that connects to the sidecar over gRPC
+WebApplicationBuilder builder = WebApplication.CreateBuilder(args);
+
+// Dapr workflows are registered as part of the service configuration
+builder.Services.AddWorkflow(options =>
+{
+    // Example of registering a "Hello World" workflow function
+    options.RegisterWorkflow<string, string>("HelloWorld", implementation: async (context, input) => 
+    {
+        string result  = "";
```
philliphoff, line 12: Nit: remove extra space and use `String.Empty`?

src/Dapr.Workflow/WorkflowContext.cs
```diff
@@ -0,0 +1,78 @@
...
+        }
+
+        /// <summary>
+        /// Asynchronously invokes an activity by name and with the specified input value.
+        /// </summary>
+        public Task<T> CallActivityAsync<T>(TaskName name, object? input = null, TaskOptions? options = null)
+        {
+            return this.innerContext.CallActivityAsync<T>(name, input, options);
```
philliphoff, line 75: Similar question re: support for cancellation.

philliphoff (commented):
Just some nits and comments re: naming, overall looks good though.

src/Dapr.Workflow/WorkflowClient.cs
```diff
@@ -0,0 +1,71 @@
...
+        /// <exception cref="ArgumentNullException">Thrown if <paramref name="innerClient"/> is <c>null</c>.</exception>
+        public WorkflowClient(DurableTaskClient innerClient)
+        {
+            this.innerClient = innerClient ?? throw new ArgumentNullException(nameof(innerClient));
+        }
+
+        /// <summary>
+        /// Schedules a new workflow instance for execution.
+        /// </summary>
+        public Task<string> ScheduleNewWorkflowAsync(
```
RyanLettieri, line 42: Should we also have the param descriptions in this method header similar to how we have done above?

src/Dapr.Workflow/WorkflowRuntimeOptions.cs
```diff
@@ -0,0 +1,75 @@
...
+    {
+        /// <summary>
+        /// Dictionary to name and register a workflow.
+        /// </summary>
+        readonly Dictionary<string, Action<DurableTaskRegistry>> factories = new();
+
+        /// <summary>
+        /// Registers a workflow as a function that takes a specified input type and returns a specified output type.
+        /// </summary>
+        public void RegisterWorkflow<TInput, TOutput>(string name, Func<WorkflowContext, TInput, Task<TOutput>> implementation)
```
RyanLettieri, line 34: Param descriptions in function header (here and throughout)

examples/Workflow/WorkflowWebApp/Program.cs
tmacam: Sorry, misread your comment. Yes, changing from 8080.

examples/Workflow/WorkflowWebApp/WorkflowWebApp.csproj
tmacam: done

examples/Workflow/WorkflowWebApp/Program.cs
tmacam: Just removed this temporary variable that was being overwritten in the next line anyway.

src/Dapr.Workflow/WorkflowMetadata.cs
tmacam: Metadata is a common idiom on Dapr, even on the .Net SDK. While I don't have strong opinions about this, I just want to make sure this is to align with .Net SDKs expectations or more a matter of preference.

cgillum: I'm fine either way on this too. If "metadata" is too overloaded, then I think "details" is fine.

src/Dapr.Workflow/WorkflowContext.cs
cgillum: No, cancellation tokens would make it easy for developers to do some wrong things with workflows (for example, you wouldn't want to cancel a durable workflow operation simply because the app host is shutting down), so we opted for a design that instead incorporated timeouts via TimeSpan.

src/Dapr.Workflow/WorkflowContext.cs
cgillum: The base SDK doesn't support this, currently.

src/Dapr.Workflow/WorkflowMetadata.cs
halspang: Metadata and Options are both used in the SDK. I'm fine either way, but would at least prefer the member to match as it feels weird to me to have something like `workflowMetadata.Details` and would prefer to either have `workflowDetails` or `workflowMetadata.OrchestrationMetadata`.

src/Dapr.Workflow/WorkflowRuntimeOptions.cs
```diff
@@ -0,0 +1,75 @@
...
+                    WorkflowActivityContext activityContext = new(innerContext);
+                    return implementation(activityContext, input);
+                });
+            });
+        }
+
+        /// <summary>
+        /// Method to add workflow to the registry.
+        /// </summary>
+        internal void AddWorkflowsToRegistry(DurableTaskRegistry registry)
```
halspang, line 66: This adds Workflows and Activities, yes? I think I may be confused as to what the difference of those is or their relation to one another.

src/Dapr.Workflow/WorkflowServiceCollectionExtensions.cs
halspang: I'm fine either way. I'm not super familiar with the existing options though, so my concern would be if this would clobber anything. Do we expect people to have our libraries and the other workflow libraries together?

src/Dapr.Workflow/WorkflowServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,63 @@
...
+            this IServiceCollection serviceCollection,
+            Action<WorkflowRuntimeOptions> configure)
+        {
+            if (serviceCollection == null)
+            {
+                throw new ArgumentNullException(nameof(serviceCollection));
+            }
+
+            serviceCollection.AddSingleton<WorkflowRuntimeOptions>();
```
halspang, line 40: Do we think it's worth checking for this to have been registered already? Things like `DaprClient` do this:

```csharp
if (services.Any(s => s.ImplementationType == typeof(DaprClientMarkerService)))
{
    return;
}

services.AddSingleton<DaprClientMarkerService>();
```

halspang (commented):
Mostly minor comments, honestly. Most comments stem from wanting to understand the workflow experience a bit more.

examples/Workflow/WorkflowWebApp/Program.cs
```diff
@@ -0,0 +1,49 @@
...
+    options.RegisterWorkflow<string, string>("HelloWorld", implementation: async (context, input) => 
+    {
+        return await context.CallActivityAsync<string>("SayHello", "World");
+    });
+
+    // Example of registering a "Say Hello" workflow activity function
+    options.RegisterActivity<string, string>("SayHello", implementation: (context, input) =>
+    {
+        return Task.FromResult($"Hello, {input}!");
+    });
```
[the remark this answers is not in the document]
tmacam: I will let other correct me but if you think of a workflow as defining a graph, activities are the nodes. Both concepts are linked.  https://docs.temporal.io/activities

src/Dapr.Workflow/WorkflowContext.cs
cgillum: Yes, I'm primarily worried that users might use it incorrectly since this is not a normal in-memory operation. It's very common in .NET to flow cancellation tokens to handle app shutdown events and it could result in unexpected behavior if developers blindly try to flow cancellation tokens to workflow context APIs. But I see your point that users may want to cancel for reasons other than timeout. I suggest we defer on this for now since the underlying API already doesn't support cancellation tokens (but I'm pinging the primary maintainer of this SDK to see if we should consider adding support for it later).

examples/Workflow/WorkflowWebApp/Program.cs
```diff
@@ -0,0 +1,49 @@
...
+});
+
+WebApplication app = builder.Build();
+
+// POST starts new workflow instances
+app.MapPost("/workflow", async (HttpContext context, WorkflowClient client) =>
+{
+    string id = Guid.NewGuid().ToString()[..8];
```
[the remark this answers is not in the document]
tmacam: Just to shorten it. Not required, the whole GUID could be used instead.

examples/Workflow/WorkflowWebApp/Program.cs
```diff
@@ -0,0 +1,49 @@
...
+    await client.ScheduleNewWorkflowAsync("HelloWorld", id);
+
+    // return an HTTP 202 and a Location header to be used for status query
+    return Results.AcceptedAtRoute("GetWorkflowEndpoint", new { id });
+});
+
+// GET fetches metadata for specific workflow instances
+app.MapGet("/workflow/{id}", async (string id, WorkflowClient client) =>
```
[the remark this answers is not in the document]
tmacam: I think that this example, as a web application might be making things a bit more complicated than they need to be.

It does make it simpler to handle the async nature of workflow execution by delegating the "status checking" to the user running the example. I think this should rather be a CLI that keeps pooling for a status change (just to make it more... interactive 😅), prints things as it goes and is done with it.

Thoughts?

examples/Workflow/WorkflowWebApp/Properties/launchSettings.json
tmacam: Ok. So we are keeping this. I will resolve this conversation then.

src/Dapr.Workflow/WorkflowContext.cs
cgillum: Ugh, sorry, I was just corrected. We _do_ have an overload in the Durable Task SDK that accepts a cancellation token. The `TimeSpan` overload was mainly added for convenience. Since there's precedence for this, we can add a `WaitForExternalEventAsync` overload that takes a cancellation token if that would be preferred.

tmacam: Do you folks think it would be worth also opening an issue on this SDK to keep this registered and so we revisit this in the future? @cgillum, is there a GitHub issue to link here or in this possible issue?

examples/Workflow/WorkflowWebApp/Program.cs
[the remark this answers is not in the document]
cgillum: This example is confusing because it treats workflows as first-class things, whereas a real app doesn't think in terms of workflows, but rather in terms of "orders", "jobs", etc. Ideally, this route would be something like "/orders/{orderId}" where the implementation maps the orderId to a workflow ID.

examples/Workflow/WorkflowWebApp/WorkflowWebApp.csproj
```diff
@@ -0,0 +1,15 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\..\src\Dapr.Workflow\Dapr.Workflow.csproj" />
+  </ItemGroup>
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net6.0</TargetFramework>
```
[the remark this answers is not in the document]
cgillum: As you mentioned, .NET 3.1 is no longer supported, so I figured it didn't make sense to add new code that requires .NET 3.1. However, the dependencies are all .NET Standard 2.0, so it's fine if you'd rather make this `netcore3.1` and then migrate everything wholesale later.

src/Dapr.Workflow/WorkflowContext.cs
cgillum: Not sure if our replies raced with each other, but the Durable Task SDK already has support for cancellation tokens here. It's up to use if we want to expose it. We can either do that now, open an issue to add support later, or just wait for customer feedback to see if they actually need/want this.

examples/Workflow/WorkflowWebApp/Program.cs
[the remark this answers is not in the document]
tmacam: I did a small renaming inside this file which should be sent along the next commit.

examples/Workflow/WorkflowWebApp/Program.cs
```diff
@@ -0,0 +1,49 @@
...
+    {
+        return Results.Ok(metadata);
+    }
+    else
+    {
+        return Results.NotFound($"No workflow with ID = '{id}' was found.");
+    }
+}).WithName("GetWorkflowEndpoint");
+
+app.Run("http://0.0.0.0:10080");
```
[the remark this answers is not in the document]
tmacam: Updated the URL in the last commit.

src/Dapr.Workflow/WorkflowServiceCollectionExtensions.cs
tmacam: Curious: should this be done for each `AddSingleton<T>` ?

examples/Workflow/WorkflowWebApp/WorkflowWebApp.csproj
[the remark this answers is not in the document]
tmacam: The current example code as it is requires net6 :/

src/Dapr.Workflow/WorkflowServiceCollectionExtensions.cs
tmacam: Added an extension method to handle it

src/Dapr.Workflow/WorkflowRuntimeOptions.cs
tmacam: You have a good point. @cgillum, is this the appropriate name for this?

src/Dapr.Workflow/WorkflowServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,102 @@
...
+                throw new ArgumentNullException(nameof(services));
+            }
+
+            if (services.None(s => s.ImplementationType == typeof(TService)))
+            {
+                return services.AddSingleton(typeof(TService));
+            }
+
+            return services;
+        }
```
tmacam, line 63: I wonder if I should move this to `DaprServiceCollectionExtensions` and refactor `AddDaprClient` to use it...

src/Dapr.Workflow/WorkflowMetadata.cs
tmacam: Marking this as resolved

src/Dapr.Workflow/WorkflowContext.cs
tmacam: I am exposing it in ad7ade292b2a495500da2f151b0f448e3c31fa53. Resolving this for now. Reopen if required.

src/Dapr.Workflow/WorkflowClient.cs
tmacam: Done.

src/Dapr.Workflow/WorkflowRuntimeOptions.cs
tmacam: Done. I did a quick check but let me know if anything is missing.

src/Dapr.Workflow/WorkflowRuntimeOptions.cs
```diff
@@ -65,10 +65,10 @@ public sealed class WorkflowRuntimeOptions
         }
 
         /// <summary>
-        /// Method to add workflow to the registry.
+        /// Method to add activities to the registry.
         /// </summary>
-        /// <param name="registry">The registry we will add workfows to</param>
-        internal void AddWorkflowsToRegistry(DurableTaskRegistry registry)
+        /// <param name="registry">The registry we will add activities to</param>
+        internal void AddActivitiesToRegistry(DurableTaskRegistry registry)
```
cgillum, line 71: It seems to me that the code in this method works for both workflows and activities. If we want to be really specific, we'd probably want to name this `AddWorkflowsAndActivitiesToRegistry`.

cgillum (approved):
I'm good with this PR as it is. There are a several additions I'd like to make, but those can be done in subsequent PRs.

src/Dapr.Workflow/WorkflowRuntimeOptions.cs
tmacam: But it doesn't really add/register any workflow with the registry, does it?

src/Dapr.Workflow/WorkflowServiceCollectionExtensions.cs
halspang: I think we could refactor it in a future PR :)

examples/Workflow/WorkflowWebApp/Program.cs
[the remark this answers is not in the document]
halspang: I wouldn't say this is required, just musing since this is a preview/alpha feature:

That was my initial thought, but I guess my question was more do the activities need to be directly related to the workflow, or can they exist outside of one? Should we have some kind of validation that checks that "SayHello" is defiend? Or should we define the activity in the workflow registration/via actual object references?

src/Dapr.Workflow/WorkflowServiceCollectionExtensions.cs
tmacam: I talked to @philliphoff  and I am using TryAddSingleton which does exactly the same verifications and addition and is part of [`microsoft.extensions.dependencyinjection.extensions`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.extensions.servicecollectiondescriptorextensions.tryaddsingleton?view=dotnet-plat-ext-7.0#microsoft-extensions-dependencyinjection-extensions-servicecollectiondescriptorextensions-tryaddsingleton-1(microsoft-extensions-dependencyinjection-iservicecollection)).

examples/Workflow/WorkflowWebApp/Program.cs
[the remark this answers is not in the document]
halspang: You should be able to remove the URL from `app.Run()` then, yes?

src/Dapr.Workflow/WorkflowRuntimeOptions.cs
halspang: I'm in favor of being more specific here, though if anyone feels super strongly we can leave it as is.

halspang (commented):
Overall, the code seems fine. I think the big thing for me is that we should remove the example entirely to start off. Examples should show how we want users to invoke/use the code instead of just providing the syntax. 

What I'd prefer to do here is create an issue to create a workflow and once the code is complete, spend some time to make a more real-world example that shows workflows as part of a larger picture.

src/Dapr.Workflow/WorkflowRuntimeOptions.cs
tmacam: Done in 509656b35a9282d4740a781c1800005b96b36262

tmacam: Resolving.

src/Dapr.Workflow/WorkflowRuntimeOptions.cs
tmacam: Renamed. Resolving.

src/Dapr.Workflow/WorkflowServiceCollectionExtensions.cs
tmacam: And this was addressed using `TryAddSingleton`. Resolving.

tmacam:
> Overall, the code seems fine. I think the big thing for me is that we should remove the example entirely to start off. Examples should show how we want users to invoke/use the code instead of just providing the syntax.
> 
> What I'd prefer to do here is create an issue to create a workflow and once the code is complete, spend some time to make a more real-world example that shows workflows as part of a larger picture.

@cgillum, you will be updating the example in a follow up PR, yes? So I will defer comment regarding the removal of the example to you.

cgillum:
Yes, I can update the example to be closer to the recent demo I gave. I agree that would be much better. However, I don't necessarily agree that we should exclude the existing example from this app. That's just creating unnecessary work for us, and I think it's honestly better than having nothing at all.

examples/Workflow/WorkflowWebApp/Program.cs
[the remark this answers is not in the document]
tmacam: Fixed in fb8589cf6100f7481532685e77de911e3ca71954. Resolving.

halspang:
> Yes, I can update the example to be closer to the recent demo I gave. I agree that would be much better. However, I don't necessarily agree that we should exclude the existing example from this PR. That's just creating unnecessary work for us, and I think it's honestly better than having nothing at all.

If it we're planning on updating it shortly, that works for me. I just wanted to make sure we had a full example before we eventually hit a full release for workflows.

halspang (approved):

Issue #976, opened 2022-10-31
Should  CloudEvent.Data property  be lowercase?


saber-wang:
## Describe the proposal

https://github.com/dapr/dotnet-sdk/blob/56017dd94ad860911beba4b6d15081d9f4d6aae8/src/Dapr.Client/CloudEvent.cs#L60

halspang:
@saber-wang - Has this caused any problems for you? If it did, it feels like something that would've come up before. Making a change to the serialization/naming conventions though would probably be considered a breaking change, so we need to be careful with what we do here.

saber-wang:
@halspang Recently processed the serializer inside the project, and the case is not processed during serialization,so...

At present, other attributes are all lowercase, and the 'CloudEvent' protocol is also lowercase, so this problem exists.

marcelk-workstreampeople-com:
Had the same issue, found out that I incorrectly configured daprClient with default JsonSerializerOptions() while it should be JsonSerializerOptions(JsonSerializerDefaults.Web). 

JsonSerializerDefaults.Web sets naming policy to CamelCase and case insensitive naming policy.

Issue #944, opened 2022-09-05
Custom CloudEvent serialization problem - Data property


mzj:
## Intro
I've stumbled on a problem while publishing events with a custom CloudEvent wrapper.
I'm trying to use custom CloudEvent<TData> so that I can specify the type property and use it for routing functionality inside subscribers.

Related to **Allow publishing cloud events directly #868**
The project is using docker-compose to set up the whole infrastructure (pub/sub services, dapr sidecars, Kafka, etc) for testing.
Demo project based on Dapr and Dapr.AspNetCore ver. 1.8, and .Net 5 (Web Api)

## Expected Behavior
Published cloud event JSON schema should look like events published without custom cloud events.

## Actual Behavior

Dapr sidecar log shows the following warning/error:

`level=warning msg="retriable error returned from app while processing pub/sub event <nil>, topic: RoutingIntegrationEvent, body: {\"type\":\"https://tools.ietf.org/html/rfc7231#section-6.5.13\",\"title\":\"Unsupported Media Type\",\"status\":415,\"traceId\":\"00-9c97aa770091eb243c1f904fbd88c27f-cfe48e0d27c8fc42-01\"}. status code returned: 415" app_id=subscriber instance=b398cfe88dc9 scope=dapr.runtime type=log ver=1.8.0`

After comparing published events with and without a custom CE wrapper I've noticed that the "data" property is different between those two types of events. 
**Specifically, pure event serializes cloud event with "data" property, while custom CE serializes it into uppercase "Data" property.** 

## Steps to Reproduce the Problem

1. Clone the repository with the fixed and runnable project (under docker-compose). https://github.com/mzj/PubSubRouting
2. Remove the above-mentioned attribute to reproduce the problem.
3. Run the solution (tested in VS2022)

## Fix
Adding `[JsonPropertyName("data")]` attribute to CloudEvent<TData>.Data property, inside Dapr.Client project.

## Release Note
 **FIX** Bug in runtime.

LotteVoorhorst:
This is still a problem today. The event cannot be used because of the parsing mismatch. Any updates on this soon?

PR #1000, opened 2022-12-20
add missing JsonPropertyName to Data property

NenadSteric:
without it you need to force the JsonSerializer to use camelCase with JsonSerializerOptions

# Description

adds JsonPropertyName to make the casing consistent with the other properties.

This also makes it possible to use just Serialize without JsonSerializerOptions
as the pubsub json handling code in Dapr needs this property to be lowercase.

This problem had to be fixed during testing of the new Bulk-publish API
where I needed to serialize the CloudEvent classes manually without the DaprClient.

# Issue reference
as discussed in
#976 and #944

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests - no new tests necessary, existing ones are green
* [x] Extended the documentation - fixes an oversight - wouldnt know where to document this

Please comment if any new tests or documentation needs to be created.


NenadSteric:
@approvers-dotnet-sdk:  anything else needed ?

WhitWaldo:
@philliphoff Should we announce this as an upcoming breaking change in the 1.15 release notes and then release with 1.16? Anyone that has been overriding the JsonSerializationOptions with any rules for the names would have their own values here whereas this change would fix the value to "data" regardless of those rules.

WhitWaldo:
@NenadSteric Thank you for your contribution!

Upon digging into this, the various Dapr clients utilize a `JsonSerializerOptions` initialized with `new JsonSerializerOptions(JsonSerializerDefaults.Web)`. One of the characteristics of the `Web` defaults is the use of "camelCase" name formatting, meaning that even without the `JsonPropertyName` attribute, the default value used will be "dapr" even without specifying it as much.

Further, because a `JsonPropertyName` would override the `JsonSerializerOptions`, merging your PR would result in developers being unable to trivially change the name format even if they really wanted to for some reason.

As such, I'm going to go ahead and reject this PR because as the underlying intent is to use a camel-cased name and it's already doing that by default and in an easily overridable manner, we're already at the desired outcome without any changes at all.

That said, if you think I'm missing something, please feel free to re-open with any follow-ups and I'd be happy to re-engage!

PR #1003, opened 2023-01-03
Workflow Management - Initial Methods

RyanLettieri:
Signed-off-by: Ryan Lettieri <ryanLettieri@microsoft.com>

# Description

Initial draft for the workflows SDK in DotNET

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


src/Dapr.Client/Protos/dapr/proto/dapr/v1/dapr.proto
```diff
@@ -287,6 +300,53 @@ message PublishEventRequest {
   map<string, string> metadata = 5;
 }
 
+// BulkPublishRequest is the message to bulk publish events to pubsub topic
+message BulkPublishRequest {
```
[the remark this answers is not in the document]
tmacam: Is this just updating the protos to match what was done in Dapr to add management support for workflows?

src/Dapr.Client/GetWorkflowResponse.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        /// <summary>
+        /// Gets the instance ID assocated with this response.
+        /// </summary>
+        public string InstanceId { set; get; }
+
+        /// <summary>
+        /// Gets the time that the workflow started.
+        /// </summary>
+        public Int64 StartTime { set; get; }
```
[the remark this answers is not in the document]
RyanLettieri: The proto definition of start_time is an int64 so StartTime in this scenario was created as an Int64

src/Dapr.Client/Protos/dapr/proto/dapr/v1/dapr.proto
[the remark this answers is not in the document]
RyanLettieri: So my understanding is that there're proto changes inside of Dapr/dapr unrelated to the work I've done that still need to be pulled into the DotNet SDK. When I'm pulling in the protos for the workflows, they're also getting pulled in.

src/Dapr.Client/DaprClient.cs
```diff
@@ -941,6 +941,53 @@ public async Task<StateEntry<TValue>> GetStateEntryAsync<TValue>(string storeNam
...
+        /// </summary>
+        /// <param name="instanceID">Identifier of the specific run.</param>
+        /// <param name="workflowComponent">The component to interface with.</param>
+        /// <param name="workflowType">Name of the workflow to run (function name).</param>
+        /// <param name="workflowOptions">The list of options that are potentially needed to start a workflow.</param>
+        /// <param name="input">The input input for the given workflow.</param>
+        /// <param name="cancellationToken">A <see cref="CancellationToken" /> that can be used to cancel the operation.</param>
+        /// <returns>A <see cref="Task"/> containing a <see cref="WorkflowReference"/></returns>
+        [Obsolete("This API is currently not stable as it is in the Alpha stage. This attribute will be removed once it is stable.")]
+        public abstract Task<WorkflowReference> StartWorkflow(
```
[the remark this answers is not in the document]
tmacam: Should I we have a "Deprecation" warning to the Workflow Authoring SDK too?

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -1374,6 +1374,118 @@ private Autogenerated.StateItem ToAutogeneratedStateItem(StateTransactionRequest
...
+            {
+                InstanceId = instanceId,
+                WorkflowComponent = workflowComponent,
+                WorkflowName = workflowType,
+                Input = (ByteString)input
+            };
+
+            if (workflowOptions?.Count > 0)
+            {
+                foreach (var item in workflowOptions)
```
tmacam, line 1405: Do we really need this `if` here?

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -1374,6 +1374,118 @@ private Autogenerated.StateItem ToAutogeneratedStateItem(StateTransactionRequest
...
+            {
+                InstanceId = instanceID,
+                WorkflowComponent = workflowComponent
+            };
+
+            var options = CreateCallOptions(headers: null, cancellationToken);
+            Autogenerated.TerminateWorkflowResponse response = new Autogenerated.TerminateWorkflowResponse(); // RRL Do we need a response?
+            try
+            {
+                response = await client.TerminateWorkflowAlpha1Async(request, options);
```
tmacam, line 1477: We aren't returning or consuming `response`. Why is it needed?

src/Dapr.Client/GetWorkflowResponse.cs
```diff
@@ -0,0 +1,56 @@
...
+    /// </summary>
+    public sealed class GetWorkflowResponse
+    {
+        /// <summary>
+        /// Initializes a new <see cref="GetWorkflowResponse" />.`
+        /// </summary>
+        /// <param name="instanceId">The instance ID assocated with this response.</param>
+        /// <param name="startTime">The time at which the workflow started executing.</param>
+        /// <param name="metadata">The response metadata.</param>
+        public GetWorkflowResponse(string instanceId, Int64 startTime, IReadOnlyDictionary<string, string> metadata)
```
tmacam, line 30: Could the class be a record? Aside construction, we never really overwrite any of its fields, right?

src/Dapr.Client/GetWorkflowResponse.cs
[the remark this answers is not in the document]
cgillum: Yes, but it's not very friendly to present a .NET developer with an integer value when a much more friendly types are available. I think we really need to use `DateTime` or `DateTimeOffset` here, and do the conversion from int64 internally.

src/Dapr.Client/DaprClient.cs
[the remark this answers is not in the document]
cgillum: @tmacam my initial thinking is that this isn't necessary in the WF authoring SDK because it's a new SDK, whereas Ryan's changes are to an existing SDK. However, we may want @halspang to weigh in on this.

src/Dapr.Client/DaprClientGrpc.cs
cgillum: I think you need it for the null check.

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -1397,10 +1397,10 @@ private Autogenerated.StateItem ToAutogeneratedStateItem(StateTransactionRequest
                 InstanceId = instanceId,
                 WorkflowComponent = workflowComponent,
                 WorkflowName = workflowType,
-                Input = input
+                Input = (ByteString)input
```
cgillum, line 1400: Yikes, this is not what I had in mind when I suggested allowing `object` instead of `ByteString`. What you have here seems very problematic because it will result in `InvalidCastException` whenever a user specifies an input that can't be cast directly into `ByteString`. What I was actually proposing is that we accept any object, [JSON-serialize it](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/how-to?pivots=dotnet-core-3-1), and then convert the serialized input JSON into whatever the protobuf contract requires (`ByteString`, in this case).

src/Dapr.Client/DaprClientGrpc.cs
RyanLettieri: It is here to check against the possibility of workflowOptions being Null.

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -1446,7 +1446,8 @@ private Autogenerated.StateItem ToAutogeneratedStateItem(StateTransactionRequest
             {
                 var options = CreateCallOptions(headers: null, cancellationToken);
                 var response = await client.GetWorkflowAlpha1Async(request, options);
-                return new GetWorkflowResponse(response.InstanceId, response.StartTime, response.Metadata);
+                var dateTimeValue = new DateTime(response.StartTime);
```
cgillum, line 1449: Let's use the [this constructor overload](https://learn.microsoft.com/en-us/dotnet/api/system.datetime.-ctor?view=net-6.0#system-datetime-ctor(system-int64-system-datetimekind)) and specify `DateTimeKind.Utc`. Developers will appreciate this.

src/Dapr.Client/GetWorkflowResponse.cs
```diff
@@ -19,15 +19,15 @@ namespace Dapr.Client
     /// <summary>
     /// Represents the response from invoking a binding.
     /// </summary>
-    public sealed class GetWorkflowResponse
+    public sealed record GetWorkflowResponse
```
cgillum, line 22: This actually isn't the right way to use `record`, so I'm a little surprised that C# lets you do this. The right way would be to do something like this:

```csharp
public record GetWorkflowResponse(string InstanceId, DateTime StartTime, IReadOnlyDictionary<string, string> Metadata);
```

And that's it - you can remove your constructor and the properties below because the C# compiler will auto-generate them.

test/Dapr.E2E.Test/Workflows/WorkflowTest.cs
```diff
@@ -0,0 +1,64 @@
...
+    using Google.Protobuf;
+    using Dapr.Client;
+
+    [System.Obsolete]
+    public partial class E2ETests
+    {
+        [Fact]
+        public async Task TestWorkflows()
+        {
+            Thread.Sleep(10000); // Sleep for 10s to wait for engine to start
```
halspang, line 32: Do we still need this after we figured out that startup issue?

test/Dapr.E2E.Test/Workflows/WorkflowTest.cs
```diff
@@ -0,0 +1,64 @@
...
+            object input = ByteString.CopyFrom(0x01);
+            Dictionary<string, string> workflowOptions = new Dictionary<string, string>();
+            workflowOptions.Add("task_queue", "testQueue");
+            CancellationToken cts = new CancellationToken();
+
+            using var daprClient = new DaprClientBuilder().UseGrpcEndpoint(this.GrpcEndpoint).UseHttpEndpoint(this.HttpEndpoint).Build();
+            var health = await daprClient.CheckHealthAsync();
+            health.Should().Be(true, "DaprClient is not healthy");
+
+            Thread.Sleep(10000); // Sleep for 10s to wait for engine to start
```
halspang, line 45: Uh oh, we do it twice!

test/Dapr.E2E.Test/Workflows/WorkflowTest.cs
```diff
@@ -0,0 +1,64 @@
...
+
+            // START WORKFLOW TEST
+            var startResponse = await daprClient.StartWorkflow(instanceID, workflowComponent, workflowType, input, workflowOptions, cts);
+            startResponse.InstanceId.Should().Be("TestWorkflowInstanceID", "Instance ID was not correct");
```
halspang, line 49: Include the ID we got from the response in the error.

test/Dapr.E2E.Test/Workflows/WorkflowTest.cs
```diff
@@ -0,0 +1,64 @@
...
+
+            // GET INFO TEST
+            var getResponse = await daprClient.GetWorkflow(instanceID, workflowComponent, workflowType);
+            getResponse.instanceId.Should().Be("TestWorkflowInstanceID");
+            getResponse.metadata["dapr.workflow.runtime_status"].Should().Be("RUNNING", "The workflow is not running when it is expected to be running");
+
+            // TERMINATE TEST:
+            await daprClient.TerminateWorkflow(instanceID, workflowComponent);
+            getResponse = await daprClient.GetWorkflow(instanceID, workflowComponent, workflowType);
+            getResponse.metadata["dapr.workflow.runtime_status"].Should().Be("TERMINATED", "The workflow is still running when it is expected to be terminated");
```
halspang, line 59: Same with these two, include the response we got in the error.

halspang (commented):
Minor comments, thanks for the work on this!

test/Dapr.E2E.Test/Workflows/WorkflowTest.cs
```diff
@@ -0,0 +1,64 @@
...
+            startResponse.InstanceId.Should().Be("TestWorkflowInstanceID", "Instance ID was not correct");
+
+            // GET INFO TEST
+            var getResponse = await daprClient.GetWorkflow(instanceID, workflowComponent, workflowType);
+            getResponse.instanceId.Should().Be("TestWorkflowInstanceID");
+            getResponse.metadata["dapr.workflow.runtime_status"].Should().Be("RUNNING", "The workflow is not running when it is expected to be running");
+
+            // TERMINATE TEST:
+            await daprClient.TerminateWorkflow(instanceID, workflowComponent);
+            getResponse = await daprClient.GetWorkflow(instanceID, workflowComponent, workflowType);
```
cgillum, line 58: I suggest doing this in a loop for maybe 10 seconds max. Workflow termination is asynchronous, so it may take a second or two for `GetWorkflow` to actually see that the workflow has terminated.

cgillum (commented):
A few more things to consider. Not necessarily blocking, except maybe the last one about how we can make the test a bit more robust.

src/Dapr.Workflow/Dapr.Workflow.csproj
```diff
@@ -2,7 +2,7 @@
 
   <!-- NuGet configuration -->
   <PropertyGroup>
-    <TargetFramework>net6</TargetFramework>
+    <TargetFrameworks>netcoreapp3.1;net5;net6</TargetFrameworks>
```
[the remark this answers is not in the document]
cgillum: Yes, it targets `netstandard2.0`, so we should be good with any of these targets for the SDK.

I would recommend `net6` instead of `net7`, however, since some users are hesitant to move to a non-LTS version of .NET.

src/Dapr.Client/DaprClient.cs
```diff
@@ -975,6 +975,53 @@ public async Task<StateEntry<TValue>> GetStateEntryAsync<TValue>(string storeNam
...
             CancellationToken cancellationToken = default);
 
+        /// <summary>
+        /// Attempt to start the given workflow with response indicating success.
+        /// </summary>
+        /// <param name="instanceID">Identifier of the specific run.</param>
+        /// <param name="workflowComponent">The component to interface with.</param>
+        /// <param name="workflowType">Name of the workflow to run.</param>
+        /// <param name="workflowOptions">The list of options that are potentially needed to start a workflow.</param>
+        /// <param name="input">The input input for the given workflow.</param>
```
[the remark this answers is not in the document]
cgillum: Yeah, there's no benefit to using a generic here. We don't really care what the type of the input is. We just need to be able to serialize it to JSON.

test/Dapr.E2E.Test/Workflows/WorkflowTest.cs
```diff
@@ -0,0 +1,63 @@
...
+            workflowOptions.Add("task_queue", "testQueue");
+            CancellationToken cts = new CancellationToken();
+
+            using var daprClient = new DaprClientBuilder().UseGrpcEndpoint(this.GrpcEndpoint).UseHttpEndpoint(this.HttpEndpoint).Build();
+            var health = await daprClient.CheckHealthAsync();
+            health.Should().Be(true, "DaprClient is not healthy");
+
+            // START WORKFLOW TEST
+            var startResponse = await daprClient.StartWorkflowAsync(instanceId, workflowComponent, workflowName, input, workflowOptions, cts);
+            startResponse.InstanceId.Should().Be("TestWorkflowInstanceID", "Instance ID $(startResponse.InstanceId) was not correct");
```
halspang, line 46: This string interpolation isn't correct.

```csharp
startResponse.InstanceId.Should().Be("TestWorkflowInstanceID", $"Instance ID {startResponse.InstanceId} was not correct");
```

halspang (commented):
Issue with string interpolation in the tests, but otherwise seems good. If @cgillum also agrees I think we can get this merged once you fix the error message.

src/Dapr.Client/Protos/dapr/proto/dapr/v1/dapr.proto
```diff
@@ -531,6 +543,23 @@ message RegisteredComponents {
   repeated string capabilities = 4;
 }
 
+message PubsubSubscription {
```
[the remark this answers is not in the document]
yash-nisar: @halspang  Nope, I don't think we're calling any auto generated code for bulk subscribe but I'll update the proto anyway just to be consistent with what we have in dapr/dapr.

test/Dapr.E2E.Test/Workflows/WorkflowTest.cs
```diff
@@ -0,0 +1,61 @@
...
+            health.Should().Be(true, "DaprClient is not healthy");
+
+            // START WORKFLOW TEST
+            var startResponse = await daprClient.StartWorkflowAsync(instanceId, workflowComponent, workflowName, input, workflowOptions, cts);
+            startResponse.InstanceId.Should().Be("TestWorkflowInstanceID", $"Instance ID {startResponse.InstanceId} was not correct");
+
+            // GET INFO TEST
+            var getResponse = await daprClient.GetWorkflowAsync(instanceId, workflowComponent, workflowName);
+            getResponse.instanceId.Should().Be("TestWorkflowInstanceID");
+            getResponse.metadata["dapr.workflow.runtime_status"].Should().Be("RUNNING", $"Instance ID {getResponse.metadata["dapr.workflow.runtime_status"]} was not correct");
```
cgillum, line 51: Non-blocking, we can probably do this later: I'm not sure what happened to my other comment, but I'm still worried about a race condition here that could cause the test to be flakey. Starting a workflow is asynchronous and the call to `GetWorkflowAsync` might return a "PENDING" status before it returns "RUNNING". We should put the `GetWorkflowAsync` call in a loop so that we can check a few times.

test/Dapr.E2E.Test/Workflows/WorkflowTest.cs
```diff
@@ -0,0 +1,61 @@
...
+
+            // TERMINATE TEST:
+            await daprClient.TerminateWorkflowAsync(instanceId, workflowComponent);
+            getResponse = await daprClient.GetWorkflowAsync(instanceId, workflowComponent, workflowName);
+            getResponse.metadata["dapr.workflow.runtime_status"].Should().Be("TERMINATED", $"Instance ID {getResponse.metadata["dapr.workflow.runtime_status"]} was not correct");
```
cgillum, line 56: Non-blocking, we can probably do this later: I'm still worried about a race condition here that could cause the test to be flakey. Workflow termination is asynchronous, so we can't guarantee that it will be terminated by the time we call `GetWorkflowAsync`. We should put the `GetWorkflowAsync` call in a loop so that we can check a few times.

cgillum (approved):
Signing off, but added some non-blocking comments for how we can make this E2E test more robust.

halspang (approved):

PR #1005, opened 2023-01-06
Add net7 to testing matrix

halspang:
# Description

Add net7 into the testing framework to ensure compatibility.

Signed-off-by: halspang <halspang@microsoft.com>

## Issue reference

Please reference the issue this PR will close: NA

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


test/Dapr.Actors.AspNetCore.IntegrationTest/ActivationTests.cs
```diff
@@ -33,7 +33,7 @@ public class ActivationTests
         public async Task CanActivateActorWithDependencyInjection()
         {
             using var factory = new AppWebApplicationFactory();
-            var httpClient = factory.CreateClient();
+            var httpClient = factory.CreateClient(new Microsoft.AspNetCore.Mvc.Testing.WebApplicationFactoryClientOptions { HandleCookies = false });
```
rynowak, line 36: Does this work on downlevel .NET versions? eg: is this API present pre-7.0?

test/Dapr.Actors.AspNetCore.IntegrationTest/ActivationTests.cs
```diff
@@ -50,7 +50,7 @@ private async Task<string> IncrementCounterAsync(HttpClient httpClient, string a
             var actorTypeName = nameof(DependencyInjectionActor);
             var methodName = nameof(DependencyInjectionActor.IncrementAsync);
 
-            var request = new HttpRequestMessage(HttpMethod.Put, $"http://localhost/actors/{actorTypeName}/{actorId}/method/{methodName}");
+            var request = new HttpRequestMessage(HttpMethod.Put, $"http://localhost:65326/actors/{actorTypeName}/{actorId}/method/{methodName}");
```
rynowak, line 53: What's the reason for this change?

rynowak:
You should be able to also remove .NET 5 from your testing matrix as it was EOL in May 2022. This is not a breaking change since it doesn't break anything for a .NET 5 user. We'll just stop testing it.

test/Dapr.Actors.AspNetCore.IntegrationTest/ActivationTests.cs
halspang: Yes, it is present for 3.1-7, though only 7 appears to handle cookies incorrectly. Given this comes from the integration suite, I believe it doesn't represent an issue with our standard code and I did manually test the actor example with net7 to ensure this isn't an issue.

test/Dapr.Actors.AspNetCore.IntegrationTest/ActivationTests.cs
halspang: I added that when I was debugging the cookie issue. I'll remove it.

rynowak (approved):

PR #1012, opened 2023-01-18
Additional API surface area for Dapr Workflow authoring SDK

cgillum:
# Description

This PR adds a significant amount of API surface area to the workflow authoring SDK. It's a follow-up to https://github.com/dapr/dotnet-sdk/pull/981, which had some feedback comments still unresolved. The other motivation for this work was that building the demo highlighted some significant gaps in the development experience.

Summary of changes:

* Workflows and activities can now be defined as classes, deriving from `Workflow` or `WorkflowActivity` base classes respectively.
* Updated example app to look more like a real order processing web app rather than "hello, world"
* Added several more APIs to the `WorkflowContext` object, allowing workflows to be far more expressive (matching what the Durable Task SDK supports).
* Renamed `WorkflowClient` to `WorkflowEngineClient` to make it more distinct from the Dapr client and added several more APIs to hide the internal dependencies.
* Renamed `WorkflowMetadata` to `WorkflowState` (warning, Git made some incorrect assumptions about which files were renamed)
* More detailed documentation

NOTE: The careful reader will notice that a lot of the code in this SDK is a very thin wrapper around the Durable Task SDK. I do wonder if trying to hide the Durable Task SDK from the Dapr SDK will at some point become unsustainable and confusing for developers. I worry that we won't be able to fully hide the fact that the Dapr Workflow engine is implemented using the Durable Task Framework. An alternative approach to what I'm doing now would be to tell Dapr developers to simply use the Durable Task SDK directly, and we could ship some Dapr-specific extension methods that work with it. This would be far less confusing and less burdensome to Dapr maintainers since the amount of code they need to maintain is dramatically reduced. However, I'm not sure how the Dapr maintainers/TOC would feel about this from a branding/control perspective. I'm open to opinions, though we don't necessarily need to block this PR on that discussion.

## Issue reference

This PR is related to the following GitHub issues:

* https://github.com/dapr/dapr/issues/4576
* https://github.com/dapr/dapr/issues/5537

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


examples/Workflow/WorkflowWebApp/Program.cs
```diff
@@ -1,51 +1,78 @@
...
 builder.Services.AddDaprWorkflow(options =>
 {
-    // Example of registering a "PlaceOrder" workflow function
-    options.RegisterWorkflow<string, string>("PlaceOrder", implementation: async (context, input) =>
-    {
-        // In real life there are other steps related to placing an order, like reserving
-        // inventory and charging the customer credit card etc. But let's keep it simple ;)
-        return await context.CallActivityAsync<string>("ShipProduct", "Coffee Beans");
-    });
+    options.RegisterWorkflow<OrderProcessingWorkflow>();
```
tmacam, line 21: Nice! It might be worth commenting that one _can_ define workflows and activities inline _but_ that in this example we define them in their own classesmore organized but that doing so isn't required...

src/Dapr.Workflow/WorkflowState.cs
```diff
@@ -0,0 +1,132 @@
...
+                        return WorkflowRuntimeStatus.Completed;
+                    case OrchestrationRuntimeStatus.Failed:
+                        return WorkflowRuntimeStatus.Failed;
+                    case OrchestrationRuntimeStatus.Terminated:
+                        return WorkflowRuntimeStatus.Terminated;
+                    case OrchestrationRuntimeStatus.Pending:
+                        return WorkflowRuntimeStatus.Pending;
+                    default:
+                        return WorkflowRuntimeStatus.Unknown;
+                }
```
tmacam, line 83: So the only thing this is doing is:

* null => Unknown
* Suspended => Unknown
* everything else is returned as is. Is that it?

I am trying to understand if this couldn't be replaced with `return this.workflowState?.RuntimeStatus ?? WorkflowRuntimeStatus.Unknown` and the `Suspended` state is the only thing that would break this, so I am trying to understand `Suspended` was omitted from that `switch` on purpose or by accident.

src/Dapr.Workflow/WorkflowState.cs
```diff
@@ -0,0 +1,132 @@
...
+        /// Deserializes the workflow output into <typeparamref name="T"/>.
+        /// </summary>
+        /// <typeparam name="T">The type to deserialize the workflow output into.</typeparam>
+        /// <returns>Returns the output as <typeparamref name="T"/>, or returns a default value if the workflow doesn't exist.</returns>
+        public T? ReadOutputAs<T>()
+        {
+            if (this.Exists)
+            {
+                return default;
+            }
```
tmacam, line 112: Is this `if` right? It looks reversed to me

src/Dapr.Workflow/WorkflowState.cs
tmacam: Ok, just noticed that this is doing a type mapping from `OrchestrationRuntimeStatus` to `WorkflowRuntimeStatus`. Anyway, re-checking if `Suspended` => should be equivalent to `Unknown`.

examples/Workflow/WorkflowWebApp/Program.cs
cgillum: Agreed. I'll add a comment.

examples/Workflow/WorkflowWebApp/Program.cs
```diff
@@ -1,51 +1,78 @@
...
+        return Results.AcceptedAtRoute("GetOrderInfoEndpoint", new { orderId }, httpResponsePayload);
     }
     else
     {
-        return Results.NotFound($"No workflow created for order with ID = '{id}' was found.");
+        // HTTP 200 OK
+        return Results.Ok(httpResponsePayload);
     }
-}).WithName("GetOrderEndpoint");
+}).WithName("GetOrderInfoEndpoint");
```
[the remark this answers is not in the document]
cgillum: We could, but I don't think it will add much value since this is just ASP.NET Core web app routing stuff.

examples/Workflow/WorkflowWebApp/Workflows/OrderProcessingWorkflow.cs
```diff
@@ -0,0 +1,44 @@
...
+
+    class OrderProcessingWorkflow : Workflow<OrderPayload, OrderResult>
+    {
+        public override async Task<OrderResult> RunAsync(WorkflowContext context, OrderPayload order)
+        {
+            string orderId = context.InstanceId;
+
+            await context.CallActivityAsync(
+                nameof(NotifyActivity),
+                new Notification($"Received order {orderId} for {order.Name} at {order.TotalCost:c}"));
```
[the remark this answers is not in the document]
cgillum: We could potentially do something like a mediator pattern, but the direction I'd rather pursue is to use C# source generators to dynamically generate type-safe wrappers for making these calls. We actually already support it in the Durable Task .NET SDK. However, I want to come up with a plan for how we can support this in Dapr Workflow too (this is related to my comment about whether we really want this SDK to hide the implementation details, because it means there are certain tooling investments that can't be shared).  For this first release, I suggest we start with this and then revisit for the next big release.

src/Dapr.Workflow/Workflow.cs
```diff
@@ -0,0 +1,137 @@
...
+    ///  don't call into external services or do complex computation directly. Rather, they delegate these tasks to
+    ///  <em>activities</em>, which perform the actual work.
+    /// </para>
+    /// <para>
+    ///   Workflows can be scheduled using the Dapr client or by other workflows as child-workflows using the
+    ///   <see cref="WorkflowContext.CallChildWorkflowAsync"/> method.
+    /// </para>
+    /// <para>
+    ///   Workflows may be replayed multiple times to rebuild their local state after being reloaded into memory.
+    ///   workflow code must therefore be <em>deterministic</em> to ensure no unexpected side effects from execution
```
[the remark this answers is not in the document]
cgillum: Actually, workflows don't need to be idempotent because they don't have any external side effects, which essentially guarantees _effectively-once_ execution guarantees. The only restriction therefore is that workflows are deterministic.

Activities don't need to be deterministic, but it is a good practice to make them idempotent since they have at-least-once execution guarantees. I believe we discuss this in the comments for the `WorkflowActivity` class.

src/Dapr.Workflow/Workflow.cs
```diff
@@ -0,0 +1,137 @@
...
+    ///     <item>
+    ///       If possible, only make changes to workflow code that won't impact its history or execution path. For
+    ///       example, renaming variables or adding log statements have no impact on a workflow's execution path and
+    ///       are safe to apply to existing workflows.
+    ///     </item>
+    ///   </list>
+    /// </para>
+    /// </remarks>
+    /// <typeparam name="TInput">The type of the input parameter that this workflow accepts. This type must be JSON-serializable.</typeparam>
+    /// <typeparam name="TOutput">The type of the return value that this workflow produces. This type must be JSON-serializable.</typeparam>
```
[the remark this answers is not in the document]
cgillum: Unfortunately, no, modern .NET doesn't have anything like this.

src/Dapr.Workflow/WorkflowState.cs
cgillum: Yes, so I unfortunately the `1.0.0-rc.1` version of the Durable Task SDK doesn't have the `Suspended` enum yet, so I can't make that change until the `1.0.0` version gets published. So for now, `WorkflowRuntimeStatus.Suspended` is unused until probably the next release.

src/Dapr.Workflow/WorkflowState.cs
cgillum: Whoops - good catch. I'll add output reading to the example so that this API gets some coverage.

cgillum (commented):
Thanks for the review, @tmacam! See my responses. I'll also push an update based on your feedback.

src/Dapr.Workflow/Workflow.cs
```diff
@@ -0,0 +1,137 @@
...
+        /// </summary>
+        Type OutputType { get; }
+
+        /// <summary>
+        /// Invokes the workflow with the specified context and input.
+        /// </summary>
+        /// <param name="context">The workflow's context.</param>
+        /// <param name="input">The workflow's input.</param>
+        /// <returns>Returns the workflow output as the result of a <see cref="Task"/>.</returns>
+        Task<object?> RunAsync(WorkflowContext context, object? input);
```
halspang, line 44: For my own edification, the reason we're doing it this way is so that when the customer extends the abstract class, they can still specify the type variables, right?

src/Dapr.Workflow/Workflow.cs
```diff
@@ -0,0 +1,137 @@
...
+    using System.Threading.Tasks;
+
+    /// <summary>
+    /// Common interface for workflow implementations.
+    /// </summary>
+    /// <remarks>
+    /// Users should not implement workflows using this interface, directly.
+    /// Instead, <see cref="Workflow{TInput, TOutput}"/> should be used to implement workflows.
+    /// </remarks>
+    public interface IWorkflow
```
halspang, line 26: If we don't want users to use this, why not make it internal or just provide the abstract class?

src/Dapr.Workflow/WorkflowActivity.cs
```diff
@@ -0,0 +1,93 @@
...
+    using System.Threading.Tasks;
+
+    /// <summary>
+    /// Common interface for workflow activity implementations.
+    /// </summary>
+    /// <remarks>
+    /// Users should not implement workflow activities using this interface, directly.
+    /// Instead, <see cref="WorkflowActivity{TInput, TOutput}"/> should be used to implement workflow activities.
+    /// </remarks>
+    public interface IWorkflowActivity
```
halspang, line 26: Same questions from Workflow here.

src/Dapr.Workflow/WorkflowState.cs
```diff
@@ -0,0 +1,132 @@
...
+
+        /// <summary>
+        /// Gets a value indicating whether the requested workflow is in a running state.
+        /// </summary>
+        public bool IsWorkflowRunning => this.workflowState?.RuntimeStatus == OrchestrationRuntimeStatus.Running;
+
+        /// <summary>
+        /// Gets a value indicating whether the requested workflow is in a terminal state.
+        /// </summary>
+        public bool IsWorkflowCompleted => this.workflowState?.IsCompleted == true;
```
halspang, line 45: Nit: Redundant `== true`?

src/Dapr.Workflow/WorkflowContext.cs
```diff
@@ -44,47 +44,299 @@ internal WorkflowContext(TaskOrchestrationContext innerContext)
...
+        /// You can use this property if you have logic that needs to run only when <em>not</em> replaying. For example,
+        /// certain types of application logging may become too noisy when duplicated as part of replay. The
+        /// application code could check to see whether the function is being replayed and then issue the log statements
+        /// when this value is <c>false</c>.
+        /// </para>
+        /// </remarks>
+        /// <value>
+        /// <c>true</c> if the workflow is currently replaying a previous execution; otherwise <c>false</c>.
+        /// </value>
+        public bool IsReplaying => this.innerContext.IsReplaying;
```
halspang, line 75: I'm not necessarily opposed to this. However, doesn't giving this kind of violate the deterministic nature of the workflows? Or at least the one we're encouraging? Cause this is giving users the ability to have workflows diverge on if it's the first execution or not?

Or am I misunderstanding the purpose of this?

examples/Workflow/WorkflowWebApp/Workflows/OrderProcessingWorkflow.cs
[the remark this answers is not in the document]
tmacam: Cool. I was wondering if there was a way to do define an extension method that would allow for something like this:

````
 await context.CallActivityAsync<NotifyActivity>(
                new Notification($"Received order {orderId} for {order.Name} at {order.TotalCost:c}"));
```` 

and inside it we would only have the `nameof(T)` thing but I don't know if that is even possible in C#. From what I learning it _might_.

I think it would improve legibility but I guess that's debatable. And this extension method would probably need a better name than just `CallActivityAsync`

File: src/Dapr.Workflow/WorkflowContext.cs [hunk 13]
```diff
@@ -44,47 +44,299 @@ internal WorkflowContext(TaskOrchestrationContext innerContext)
...
+        {
+            return this.innerContext.CallSubOrchestratorAsync(workflowName, input, options);
+        }
+
+        /// <summary>
+        /// Restarts the workflow with a new input and clears its history.
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// This method is primarily designed for eternal workflows, which are workflows that
```
halspang, line 297: Are eternal workflows common? The idea of that scares me 😅 I'm fine keeping this, just curious.

halspang (commented):
Overall this looks good pretty good to me. I just have a few questions but I don't see anything glaring in here.

examples/Workflow/WorkflowWebApp/Workflows/OrderProcessingWorkflow.cs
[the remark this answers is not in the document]
cgillum: You can certainly replace `nameof(NotifyActivity)` with `<NotifyActivity>(...)` but I don't think it creates enough of a better experience to justify creating this new overload. It would eliminate the problem of specifying the name of something completely wrong at compile time, but it doesn't help with the input data type validation at compile time. We could write a bunch of reflection code to solve that at runtime, but I think a better solution would be to leverage source generation in a way that solves both the name and input type problem at compile time.

src/Dapr.Workflow/WorkflowContext.cs
cgillum: You're right that this could be used incorrectly, but I haven't seen that happen in practice with Durable Functions over the past several years. It's mostly used for things like logging, where you only want to write a log statement the first time a piece of code executes.

src/Dapr.Workflow/WorkflowContext.cs
cgillum: Yes, it's actually very common. 😅 It's a way for you to create some kind of long running agent that doesn't have an explicit lifetime, like a monitor that maybe runs in a loop and only takes action when some event occurs.

src/Dapr.Workflow/WorkflowState.cs
cgillum: It's actually not redundant because of the `this.workflowState?.IsCompleted` resolves to `Nullable<bool>` and not `bool`.

src/Dapr.Workflow/Workflow.cs
cgillum: I can't make it internal because there are public APIs that depend on it (like the `RegisterWorkflow<T>()` API). I believe there was some reason I needed this because of constraints related to how generics work in .NET, though I can't remember exactly what it was...

src/Dapr.Workflow/Workflow.cs
cgillum: The main reason we're doing this is so that we can have common code that can run the user's code without needing to worry about the specific type parameters. I don't recall exactly, but I think there wasn't a way for me to write this common code when generics were involved. It was a problem I originally encountered in the design of the Durable Task SDK that we depend on.

cgillum (commented):
Thanks @halspang for your comments! I've added my responses.

tmacam (dismissed):
LGTM

src/Dapr.Workflow/WorkflowState.cs
halspang: Sad, I thought .NET would figure that out for you.

src/Dapr.Workflow/WorkflowServiceCollectionExtensions.cs
```diff
@@ -41,12 +39,35 @@ public static class WorkflowServiceCollectionExtensions
             }
 
             serviceCollection.TryAddSingleton<WorkflowRuntimeOptions>();
-            serviceCollection.TryAddSingleton<WorkflowClient>();
+            serviceCollection.TryAddSingleton<WorkflowEngineClient>();
             serviceCollection.AddDaprClient();
 
+            serviceCollection.AddOptions<WorkflowRuntimeOptions>().Configure(configure);
+
+            static bool TryGetGrpcAddress(out string address)
```
halspang, line 47: We have a shared package that may already do what you need here.

https://github.com/dapr/dotnet-sdk/blob/master/src/Shared/DaprDefaults.cs

halspang (commented):
One small comment and can I also ask you to update the `.csproj` for workflows to no longer target `net5`? We're removing it in another PR so if you don't want to I can get it there as well.

src/Dapr.Workflow/WorkflowServiceCollectionExtensions.cs
cgillum: Thanks for pointing this out, I hadn't seen it. I ran into a couple problems trying to use it as-is though. For now, I'm going to put a TODO comment in the code suggesting that we adopt the common code if/when the problems can be safely addressed.

halspang (approved):

PR #1014, opened 2023-01-24
Inroduce OnActorMethodFailedAsync virtual method to simplify error logging

vlardn:
# Description

Currently, to catch exceptions thrown by any actor method (and log them somewhere) you need to add the `try/catch` block to ALL actor methods. For timer and reminder actor callbacks, this is almost mandatory, otherwise, exceptions will never appear anywhere. But all these `try/catch` blocks bloat the actors' code and make it less readable. 

Given change introduces the `OnActorMethodFailedAsync` virtual method which can be overridden by the user the same way as the `OnPreActorMethodAsync` or `OnPostActorMethodAsync` methods and can be used (in most cases) to log exceptions thrown by any actor method as well as by `OnPreActorMethodAsync` or `OnPostActorMethodAsync` overridden methods. Overriding of the `OnActorMethodFailedAsync` method significantly simplifies actor error logging code for the user.


vlardn:
Could someone either approve or at least comment on this?

halspang (approved):

halspang:
Thanks for the contribution, sorry for the long delay!

Issue #1034, opened 2023-02-14
Workflow testability 


olitomlinson:
WorkflowContext does not have an interface and the methods are not virtual, which makes unit testing the workflow, and mocking the context challenging.

Raising on behalf of @cgillum

PR #1035, opened 2023-02-14
Enable unit testing for Dapr Workflows

cgillum:
# Description

An early user of Dapr Workflow pointed out that they aren't able to unit test their workflows. This PR addresses that problem by:

1. Making `WorkflowContext` a mockable abstract base class
2. Providing an example app that shows how to unit test workflows

The new unit test project isn't meant as an actual test to be run by us, but as a reference for developers so that they can learn how to write their own unit tests for workflows.

Note that the unit test project references the workflow console app sample, so I made many of the types in that sample `public` so that they could be easily tested from another project.

I also found a bug while refactoring the `WorkflowContext` - calls to `WorkflowContext.SendEvent()` would have resulted in a `StackOverflowException`. That is fixed in this PR too.

## Issue reference

Resolves https://github.com/dapr/dotnet-sdk/issues/1034

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/examples/Workflow/WorkflowUnitTest/WorkflowUnitTest.csproj
+++ b/examples/Workflow/WorkflowUnitTest/WorkflowUnitTest.csproj
@@ -1,0 +1,28 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
+    <Nullable>enable</Nullable>
+    <LangVersion>10</LangVersion>
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.1.0" />
+    <PackageReference Include="Moq" Version="4.18.4" />
+    <PackageReference Include="xunit" Version="2.4.1" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.4.3">
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+      <PrivateAssets>all</PrivateAssets>
+    </PackageReference>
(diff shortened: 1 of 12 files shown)


examples/Workflow/WorkflowUnitTest/Usings.cs
```diff
@@ -0,0 +1 @@
+global using Xunit;
```
halspang, line 1: Personal preference, you don't need to change this, but I don't really like global usings just cause it makes it harder to see what individual file/classes are bringing in.

examples/Workflow/WorkflowUnitTest/WorkflowUnitTest.csproj
```diff
@@ -0,0 +1,29 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
```
halspang, line 5: Again, personal preference, so feel free to ignore. I just don't like these 😅

halspang (commented):
Code is fine overall, hooray testing! Just some nit-picks you can feel free to ignore. However, you do have a build issue that you'll need to look into.

examples/Workflow/WorkflowUnitTest/Usings.cs
cgillum: I'm fine removing it. I just left it there since VS added it, but I agree that it can make the code harder to reason about.

halspang (dismissed):

halspang (approved):

mylona79:
Hi Dapr team, @cgillum , 

In which release version of .Net SDK we have that available?

cgillum:
@halspang do you know the answer?

mylona79:
Hi @halspang ,

Kindly provide the Dapr sdk release where this feature is available.

Thanks a lot!
Maria

Issue #1032, opened 2023-02-11
Use "localhost" instead of "127.0.0.1" for the sidecar host address


cgillum:
## Describe the proposal

Currently we use `127.0.0.1` when accessing the Dapr sidecar from the .NET SDK. This is hardcoded [here](https://github.com/dapr/dotnet-sdk/blob/9dcae7b0e771d7328559bef1dd65df4c1a54b793/src/Shared/DaprDefaults.cs#L69). We should change this to instead be `localhost` because:

1. `127.0.0.1` only works with IPv4 whereas `localhost` works with both IPv4 and IPv6.
2. Unlike with `localhost`, using `127.0.0.1` can't resolve to local WSL2 endpoints from the Windows OS, which is a hindrance for local dev/test on Windows.

As an alternative, we could consider making the host address configurable via an environment variable.

ondmal:
Hi @cgillum.

>> As an alternative, we could consider making the host address configurable via an environment variable.

- Do you have any proposal for the environment variable name? 
- Do you still want to use localhost if the environment variable is not defined?

ondmal:
I do not know the system well, so I can´t say what is already reserved. But `DAPR_HOST` looks good to me.

Can we evaluate it once and cache it in static variable or it can vary during runtime?

itpropro:
Any updates on this? This is regularly coming up in local Windows dev environments.

WhitWaldo:
@itpropro This is accommodated by the new DI registration for each Dapr client that will release with 1.15.

I would encourage you to try using the latest Dapr .NET SDK RC and see if it solves your local issues.

itpropro:
> [@itpropro](https://github.com/itpropro) This is accommodated by the [new DI registration](https://github.com/dapr/dotnet-sdk/pull/1363) for each Dapr client that will release with 1.15.
> 
> The conversation specifically covered on this thread was superseded by a consistent envvar used by all the SDKs for `DAPR_HTTP_ENDPOINT` or `DAPR_GRPC_ENDPOINT`. The update I referenced will first look for the necessary values from the options provided during the DI registration, then attempt to retrieve from an `IConfiguration` and finally pull from the environment variables as a fallback.
> 
> If not set across the board, it defaults to using `localhost` instead of '127.0.0.1` with the default Dapr ports (3500 for HTTP and 50001 for gRPC).
> 
[...]

PR #1040, opened 2023-02-17
Using environment variable for dapr sidecar host and using localhost instead of 127.0.0.1 as DAPR default sidecar host

ondmal:
# Description

-  retrieve default value of DAPR sidecar host from DAPR_HOST environment variable
-  use localhost instead of 127.0.0.1 as DAPR sidecar host default value if the environment variable is not specified

## Issue reference

This PR will close: #1032


ondmal:
One thing to consider.

`WorkflowServiceCollectionExtensions` class uses method `TryGetGrpcAddress` to validate the default GRPC endpoint and uses empty string if it is not well formed. 

Maybe we should either validate the endpoint address when constructing the default (in `DaprDefaults.cs`) or do not validate it at all.

artursouza:
This change impacts this PR: https://github.com/dapr/dotnet-sdk/pull/1124

WhitWaldo:
@ondmal Thank you for your contribution!

I'm going to go ahead and close this as I believe it's been superseded by the work in #1124 and recently in #1363, but please don't hesitate to re-open if you think I've closed it inaccurately.

Issue #1053, opened 2023-03-08
[Workflow] Improvements to WorkflowEngineClient


cgillum:
#### Rename `WorkflowEngineClient` to `DaprWorkflowClient`
The latter name is much clearer. For backwards compatibility, we can continue to have a `WorkflowEngineClient` type, but it should be an empty child class of the new `DaprWorkflowClient` and have an `[Obsolete]` attribute that points to the new class.

```csharp
[Obsolete("Use DaprWorkflowClient instead")]
public class WorkflowEngineClient : DaprWorkflowClient
{
}
```

#### Get to parity with DaprClient

The primary reason to keep DaprWorkflowClient separate but equal to DaprClient APIs for workflows is that the latter is bound to the common workflow API contract whereas the former can be optimized for users of the built-in Dapr Workflow engine.

This includes the following APIs:
* WaitForWorkflowStartAsync
* WaitForInstanceCompletionAsync
* TerminateWorkflowAsync
* RaiseWorkflowEventAsync
* SuspendWorkflowAsync
* ResumeWorkflowAsync
* PurgeWorkflowDataAsync

Note that these are simple wrappers over similarly named APIs on the `DurableTaskClient` class, which is the internal implementation of all these APIs.

## Release Note

RELEASE NOTE: **UPDATE** Rename `WorkflowEngineClient` to `DaprWorkflowClient` and added additional APIs.

PR #1059, opened 2023-03-14
Workflow SDK changes for v1.11

cgillum:
# Description

This PR contains the changes outlined in https://github.com/dapr/dotnet-sdk/issues/1053. I'll summarize them here at a higher level:

* Rename `WorkflowEngineClient` to `DaprWorkflowClient`
  * I'm keeping `WorkflowEngineClient` around as a sub-class with an `[Obsolete]` attribute for backwards compatibility
* Added several new methods (terminate, wait for workflow start/complete, raise event, suspend, resume, purge)
* Added error information to the `WorkflowState` class. Previously we needed to use `DaprClient` for this, which wasn't ideal.
* Added the ability to get a `DaprWorkflowClient` object without needing to also define workflows.

I've also updated the console app sample to use some of the new APIs. In the future, we'll want to enhance the sample scenario further to incorporate additional functionality, such as raise event for human approvals of purchase orders.

## Issue reference

Resolves https://github.com/dapr/dotnet-sdk/issues/1053

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


src/Dapr.Workflow/DaprWorkflowClient.cs
```diff
@@ -0,0 +1,310 @@
...
+        /// <param name="cancellation">
+        /// A <see cref="CancellationToken"/> that can be used to cancel the purge operation.
+        /// </param>
+        /// <returns>
+        /// Returns a task that completes when the purge operation has completed. The value of this task will be
+        /// <c>true</c> if the workflow state was found and purged successfully; <c>false</c> otherwise.
+        /// </returns>
+        public async Task<bool> PurgeInstanceAsync(string instanceId, CancellationToken cancellation = default)
+        {
+            PurgeResult result = await this.innerClient.PurgeInstanceAsync(instanceId, cancellation);
```
cgillum, line 283: Note that purge isn't yet implemented by the engine and won't work until that implementation is complete.

src/Dapr.Workflow/DaprWorkflowClient.cs
```diff
@@ -0,0 +1,310 @@
...
+        {
+            return ((IAsyncDisposable)this.innerClient).DisposeAsync();
+        }
+    }
+
+    /// <summary>
+    /// Deprecated. Use <see cref="DaprWorkflowClient"/> instead.
+    /// </summary>
+    [Obsolete($"Deprecated. Use {nameof(DaprWorkflowClient)} instead.")]
+    public sealed class WorkflowEngineClient : DaprWorkflowClient
```
halspang, line 299: Does moving the class to a different file cause any weird binary breaking changes? I appreciate the less code of removing the class and putting it here, just want to make sure we don't have any weirdness that happens.

halspang (commented):
Just some questions, overall seems good.

src/Dapr.Workflow/DaprWorkflowClient.cs
```diff
@@ -0,0 +1,310 @@
...
+
+        /// <summary>
+        /// Waits for a workflow to start running and returns a <see cref="WorkflowState"/> object that contains metadata
+        /// about the started workflow.
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// A "started" workflow instance is any instance not in the <see cref="WorkflowRuntimeStatus.Pending"/> state.
+        /// </para><para>
+        /// If a workflow instance is already running (or completed) when this method is called, the method will return immediately.
```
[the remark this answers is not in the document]
cgillum: I agree it's confusing. I'll reword it to something simpler. However, I don't want to list the states since that list is large and may change over time. Let me know if this rewording works for you.

src/Dapr.Workflow/DaprWorkflowClient.cs
cgillum: I think it's better to not add a comment here since the purge functionality will be available before we ship this SDK version (Ryan and I are actively working on it now). Otherwise, we'd need another PR in a week or two to remove the doc/comment.

src/Dapr.Workflow/DaprWorkflowClient.cs
```diff
@@ -0,0 +1,310 @@
...
+        /// </param>
+        /// <returns>
+        /// Returns a task that completes when the purge operation has completed. The value of this task will be
+        /// <c>true</c> if the workflow state was found and purged successfully; <c>false</c> otherwise.
+        /// </returns>
+        public async Task<bool> PurgeInstanceAsync(string instanceId, CancellationToken cancellation = default)
+        {
+            PurgeResult result = await this.innerClient.PurgeInstanceAsync(instanceId, cancellation);
+            return result.PurgedInstanceCount > 0;
        }
```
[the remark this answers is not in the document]
cgillum: When I reopened this file, VS told me that it found inconsistencies in the newlines. I accepted its offer to fix all of them, so hopefully this `^M` is gone in the current iteration.

src/Dapr.Workflow/DaprWorkflowClient.cs
cgillum: Moving to a different file doesn't have any impact on the built binaries. But I'll go ahead and move this to its own file anyways to avoid confusion.

cgillum (commented):
Thanks for the comments! I've updated the PR and responded to each comment.

src/Dapr.Workflow/DaprWorkflowClient.cs
```diff
@@ -0,0 +1,295 @@
...
+
+        /// <summary>
+        /// Waits for a workflow to start running and returns a <see cref="WorkflowState"/> object that contains metadata
+        /// about the started workflow.
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// A "started" workflow instance is any instance not in the <see cref="WorkflowRuntimeStatus.Pending"/> state.
+        /// </para><para>
+        /// This method will return a completed task if the workflow has already started running.
```
halspang, line 91: This is also true if a workflow is considered "Completed", right? I think just adding that would be good here.

cgillum: Added

halspang (approved):

PR #1077, opened 2023-04-21
Move config api to stable

shivamkm07:
# Description

This PR updates config API to stable. It updates the protos to include the stable apis. There has been some workflow changes causing the build to fail with latest dapr protos. It also fixes it

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


.github/workflows/itests.yml
```diff
@@ -41,21 +41,21 @@ jobs:
...
       DAPR_CLI_REF: ''
-      DAPR_REF: '82e097134cdc7494c5c3be71bf80f7904d6db9ea'
+      DAPR_REF: 'dbeea96710d38d08dcc5aaf6807914a7d697bed4'
     steps:
       - name: Set up Dapr CLI
         run: wget -q ${{ env.DAPR_INSTALL_URL }} -O - | /bin/bash -s ${{ env.DAPR_CLI_VER }}
       - name: Set up Go ${{ env.GOVER }}
         if: env.DAPR_REF != '' || env.DAPR_CLI_REF != ''
-        uses: actions/setup-go@v2
+        uses: actions/setup-go@v3
```
halspang, line 58: Can we do it like this instead? Then we don't need to keep updating the go version. You'll have to move this step to in between pulling and building dapr.

https://github.com/dapr/test-infra/blob/master/.github/workflows/dapr-longhaul-nightly.yml#L58

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -1526,7 +1526,7 @@ private Autogenerated.StateItem ToAutogeneratedStateItem(StateTransactionRequest
             {
                 InstanceId = instanceId,
                 WorkflowComponent = workflowComponent,
-                WorkflowType = workflowName //TODO: Change 'WorkflowType' to 'WorkflowName' once changes go through dapr/dapr
+                WorkflowName = workflowName
```
halspang, line 1529: This change and the one below it just come from updating the protos?

halspang (commented):
Two small comments.

src/Dapr.Client/DaprClientGrpc.cs
shivamkm07: Yes, updating the protos was failing the tests initially. On debugging I found i is due to these renaming in protos, so had to do here as well.

Issue #1081, opened 2023-05-02
Incorrect TaskFailedException is used when exceptions are thrown from workflow activities


marcduiker:
## Expected Behavior

When a `catch (DurableTask.Core.Exceptions.TaskFailedException) {...}` block is used in a Dapr workflow, I expect that it catches any exception that is thrown from an activity that the try/catch wraps.

## Actual Behavior

The catch block does not catch any exceptions.

It appears that the outer exception is of type `Microsoft.DurableTask.TaskFailedException` and not `DurableTask.Core.Exceptions.TaskFailedException`.

![Screenshot 2023-05-02 083404](https://user-images.githubusercontent.com/4191277/235598026-1ba8217e-2d15-49fe-b309-b4e7bc1156b9.png)

## Steps to Reproduce the Problem

Use the Dapr Workflow quickstart and comment out this part in the workflow:

```csharp
if (!result.Success)
{
   // End the workflow here since we don't have sufficient inventory
   await context.CallActivityAsync(
      nameof(NotifyActivity),
         new Notification($"Insufficient inventory for {order.Name}"));
     return new OrderResult(Processed: false);
}
```
Start the workflow with an order that has a higher quantity than the inventory, so the `InvalidOperationException` in the `UpdateInventoryActivity` will be thrown.

## Release Note

RELEASE NOTE:

cgillum:
Hi @marcduiker. Thanks for raising this. We're fixing this in v1.11 by actually changing the type of exception thrown when an activity or child-workflow task fails. Instead of `TaskFailedException` (which is a Durable Task-specific exception), you'll see `WorkflowTaskFailedException` (which is a Dapr-specific exception type). The PR is here: https://github.com/dapr/dotnet-sdk/pull/1086. This will be a breaking change, so you'll need to update your `catch` statements accordingly.

PR #1086, opened 2023-05-03
[Workflow] Add WorkflowTaskFailedException for workflow error handling

cgillum:
# Description

This change adds a new `WorkflowTaskFailedException` type for handling activity and child-workflow exceptions from inside a workflow. It's meant to replace the usage of `TaskFailedException` type defined in the Durable Task SDK, which we're trying to hide.

This PR also fixes a bug in the console app sample for workflows, plus cleaning up a few things.

This is a breaking change, but one that we were planning to make anyways since v1.10.

## Issue reference

Fixes https://github.com/dapr/dotnet-sdk/issues/1081

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

FYI @RyanLettieri @DeepanshuA @nyemade-uversky @marcduiker


halspang (changes requested):

src/Dapr.Workflow/WorkflowTaskFailedException.cs
```diff
@@ -0,0 +1,39 @@
...
+// ------------------------------------------------------------------------
+
+namespace Dapr.Workflow
+{
+    using System;
+
+    /// <summary>
+    /// Exception type for Dapr Workflow task failures.
+    /// </summary>
+    public class WorkflowTaskFailedException : Exception
```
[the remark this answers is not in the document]
cgillum: Yes, this is intentionally a breaking change. The TaskFailedException that Marc was trying to catch is actually not something we want to expose publicly since it's defined by the Durable Task Framework, and we're trying to hide those implementation details from the programming model. We solve the problem of leaking Durable types by creating our own exception type (`WorkflowTaskFailedException`) and asking users to catch that one instead.

examples/Workflow/WorkflowConsoleApp/Activities/UpdateInventoryActivity.cs
```diff
@@ -39,7 +39,7 @@ public override async Task<object> RunAsync(WorkflowActivityContext context, Pay
                 this.logger.LogInformation(
                     "Payment for request ID '{requestId}' could not be processed. Insufficient inventory.",
                     req.RequestId);
-                throw new InvalidOperationException();
+                throw new InvalidOperationException("Not enough inventory!");
```
RyanLettieri, line 42: Not a blocker for this review, but can we include the item that we are trying to purchase here as well as the number of items still available of that type for purchase?

cgillum: Agreed, that would have been better. It's actually really difficult to hit this case today since the ReserveInventoryActivity will do this check already before we get here, so I don't think users playing with the sample will ever hit this, but I'll update it if there are other blocking things that need to be done for this PR.

Issue #1027, opened 2023-02-08
[Workflow] Improvements to Dapr Client APIs for Workflow


cgillum:
## Describe the proposal

I'm proposing we make the following changes to the workflow management APIs in the .NET Dapr client. This is based on the feedback we got from the v1.10 alpha release of the workflow management APIs.

* In `StartWorkflowAsync`, make `instanceId` and `input` parameters optional.

```csharp
public abstract Task<WorkflowReference> StartWorkflowAsync(
    string workflowComponent,
    string workflowName,
    string? instanceId = null,
    object? input = null,
    IReadOnlyDictionary<string, string> workflowOptions = default,
    CancellationToken cancellationToken = default);
```

* Replace `GetWorkflowResponse` with `WorkflowState` from the authoring SDK
* Add `WorkflowFailureDetails` property to `WorkflowState` so that users can obtain more information about a workflow failure.

```csharp
public record WorkflowFailureDetails(
    string ErrorType,
    string ErrorMessage,
    string? StackTrace,
    WorkflowFailureDetails? InnerFailure)
```

* Add the following methods to `DaprClient`:

```csharp
// Waits for a workflow to start running and then returns the state associated with that workflow
public WorkflowState WaitForWorkflowStartAsync(
    string instanceId,
    bool fetchPayloads = false,
    CancellationToken cancellationToken = default);

// Wait for workflow the workflow to run to completion, be terminated, or fail
public WorkflowState WaitForWorkflowCompletionAsync(
    string instanceId,
    bool fetchPayloads = true,
    CancellationToken cancellationToken = default);
```

/cc @nyemade-uversky

cgillum:
As part of this work, we should also update the console app sample to use the updated Dapr client instead of workflow engine client. That will help us be confident that the new API surface area is sufficient for real use.

msfussell:
> 

I also added this issue in the Quickstarts repo https://github.com/dapr/quickstarts/issues/830

PR #1087, opened 2023-05-10
[Workflow] Improve management API usability

cgillum:
# Description

This PR contains breaking changes to the v1.10 workflow management client APIs. The point of these changes is to improve the usability of these core APIs. This includes adding a few additional convenience types and methods.

This PR also updates the console app sample, replacing the `DaprWorkflowClient` with `DaprClient`, which was always the intent.

## Issue reference

Resolves https://github.com/dapr/dotnet-sdk/issues/1027.

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

FYI @RyanLettieri


examples/Workflow/WorkflowConsoleApp/Program.cs
```diff
@@ -114,23 +112,31 @@
 
     // Start the workflow using the order ID as the workflow ID
     Console.WriteLine($"Starting order workflow '{orderId}' purchasing {amount} {itemName}");
-    await workflowClient.ScheduleNewWorkflowAsync(
-        name: nameof(OrderProcessingWorkflow),
+    await daprClient.StartWorkflowAsync(
+        WorkflowComponent,
```
RyanLettieri, line 116: Let's change this to be "workflowComponent: WorkflowComponent" for consistency.

src/Dapr.Client/GetWorkflowResponse.cs
```diff
@@ -11,25 +11,91 @@
...
+        /// </summary>
+        public WorkflowRuntimeStatus RuntimeStatus { get; init; }
+
+        /// <summary>
+        /// Gets the component-specific workflow properties.
+        /// </summary>
+        public IReadOnlyDictionary<string, string> Properties { get; init; }
+
+        /// <summary>
+        /// Gets the defaults associated with the workflow failure, if any.
```
RyanLettieri, line 61: "details" instead of "defaults"

RyanLettieri:
While you didn't touch the file, we should update the asserts inside WorkflowTest.cs to use the enumeration: "WorkflowRuntimeStatus" for the status of the runtime rather than hardcoded values such as "RUNNING", "SUSPENDED", etc.

cgillum:
> While you didn't touch the file, we should update the asserts inside WorkflowTest.cs to use the enumeration: "WorkflowRuntimeStatus" for the status of the runtime rather than hardcoded values such as "RUNNING", "SUSPENDED", etc.

Oh, interesting. I've never used FluentAssertions before, but it seems like a weakness in its design. Ideally the compiler would have caught this. I'll go ahead and fix that, along with your other suggestions.

src/Dapr.Client/DaprClient.cs
```diff
@@ -978,22 +978,95 @@ public async Task<StateEntry<TValue>> GetStateEntryAsync<TValue>(string storeNam
...
+        /// </returns>
+        [Obsolete("This API is currently not stable as it is in the Alpha stage. This attribute will be removed once it is stable.")]
+        public virtual async Task<GetWorkflowResponse> WaitForWorkflowStartAsync(
+            string instanceId,
+            string workflowComponent,
+            CancellationToken cancellationToken = default)
+        {
+            while (true)
+            {
+                GetWorkflowResponse response = await this.GetWorkflowAsync(instanceId, workflowComponent, cancellationToken);
```
halspang, line 1022: Nit: Any reason we don't use var? We're fairly consistent with it throughout the rest of the client.

src/Dapr.Client/DaprClient.cs
```diff
@@ -978,22 +978,95 @@ public async Task<StateEntry<TValue>> GetStateEntryAsync<TValue>(string storeNam
...
+        /// </para>
+        /// </remarks>
+        /// <param name="instanceId">The unique ID of the workflow instance to wait for.</param>
+        /// <param name="workflowComponent">The component to interface with.</param>
+        /// <param name="cancellationToken">A <see cref="CancellationToken"/> that can be used to cancel the wait operation.</param>
+        /// <returns>
+        /// Returns a <see cref="GetWorkflowResponse"/> record that describes the workflow instance and its execution status.
+        /// </returns>
+        [Obsolete("This API is currently not stable as it is in the Alpha stage. This attribute will be removed once it is stable.")]
+        public virtual async Task<GetWorkflowResponse> WaitForWorkflowStartAsync(
```
halspang, line 1015: Nit: Why are you defining these in the abstract class instead of the implementation? For things like `WaitForSidecarAsync` we declare it here and define it in `DaprClientGrpc`.

Generally, the only methods implemented in the base class are helpers (like generating the http client/grpc proxy) or ones that involve http in general.

src/Dapr.Client/GetWorkflowResponse.cs
```diff
@@ -11,25 +11,91 @@
...
 // ------------------------------------------------------------------------
 
-using System;
-using System.Collections.Generic;
-
 namespace Dapr.Client
 {
+    using System;
+    using System.Collections.Generic;
+    using System.Text.Json;
```
halspang, line 18: In general, we keep the `using` outside of the namespace.

halspang (commented):
Mostly nitpicks, overall seems fine.

src/Dapr.Client/DaprClient.cs
cgillum: I define it in the abstract class to make it simpler for new implementations and/or mocks to consume it - i.e. two less things to override. This is just a helper/wrapper around `GetWorkflowAsync`. I can move it if that's the preferred style for this project.

src/Dapr.Client/GetWorkflowResponse.cs
cgillum: I specifically checked for this and the random files I looked at had `using`s inside the namespace. I guess I checked the wrong random files. :) I can move them out.

cgillum: I'll also add this to the repo's .editorconfig file since it doesn't currently specify either way (and my machine defaults to inside).

halspang (approved):

PR #1090, opened 2023-05-16
[Workflow] Proper workflow retry support in Dapr SDK

cgillum:
# Description

The Dapr Workflows SDK didn't have its own types for managing workflow retry policies. Instead, it relied on types defined in the underlying Durable Task SDK. This was bad/confusing because we want to keep the Durable Task SDK an internal implementation detail as much as possible.

With this change, we can now properly document how to implement retry policies for workflows using the .NET SDK.

A few notes about the feature:

* Retry policies can be applied to activities or child workflows
* Workflow task retry policies are durable and survive process crashes. For example, if you have a policy to retry only 3 times, and your workflow crashes after the second retry, when it comes back up, it will do only one more retry before giving up.
* Workflow retry policies are not at all related to Dapr Resiliency policies, but you can compose the two together if you want.
* Retry policies allow you to specify the following:
  * Max number of attempts
  * First retry interval
  * Backoff coefficient
  * Max retry interval
  * Total retry timeout

FYI @nyemade-uversky @DeepanshuA @halspang 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/examples/Workflow/WorkflowConsoleApp/Activities/ReserveInventoryActivity.cs
+++ b/examples/Workflow/WorkflowConsoleApp/Activities/ReserveInventoryActivity.cs
@@ -1,5 +1,4 @@
-﻿using System.Threading.Tasks;
-using Dapr.Client;
+﻿using Dapr.Client;
 using Dapr.Workflow;
 using Microsoft.Extensions.Logging;
 using WorkflowConsoleApp.Models;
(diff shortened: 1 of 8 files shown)


cgillum:
Here's the docs PR which describes the feature: https://github.com/dapr/docs/pull/3419

halspang (approved):

PR #1102, opened 2023-06-02
Update fossa link to point from `contrib` to `dotnet`

JoshVanL:


halspang:
Thanks for opening this, I was just going in order on the pull requests and apparently Bernd's PR does the same thing right above it in the list #1105. I merged that one first so this one is redundant.

Issue #772, opened 2021-10-19
Actor: Get existing reminder by name


ssm3ll:
## Expected Behavior

As per official documentation https://docs.dapr.io/reference/api/actors_api/#get-actor-reminder it is possible to get existing reminder by its name. So Actor should contain method GetReminderAsync(string reminderName) to get existing reminder, that has been previously created using RegisterReminderAsync (string reminderName, byte[] state, TimeSpan dueTime, TimeSpan period) method

## Actual Behavior

Currently it's only possible to register or unregister reminder

## Steps to Reproduce the Problem

Nor Actor neither DaprHttpInteractor contain method to get existing actor reminder by it's name

rynowak:
Hi @ssm3ll,

You're right that we don't surface this through the Actors SDK. Do you have a use case in mind here? What do you plan to do with the data you will get? 

That information will help us prioritize this as well as determine how it should be designed.

tstojecki:
In the absence of this, is there any downside to registering a reminder with the same name with each request in a controller? We have a service that uses an actor to get/set state and want to have one instance of a reminder running at all times.

fabistb:
@halspang Is it possible to surface this api via the SDK?
Based on my tests the reminder API behavior changes when reminder partitioning is activated.

If reminder partitioning isn't activate _UnregisterReminder()_ returns 200 even if no reminder is currently active.
This makes it possible to call the method in an optimistic if it isn't clear if a reminder is currently active or not.

However if reminder partitioning is activate _UnregisterReminder()_ returns 500 if no reminder is currently active.
With this change the best way is probably to call the _Get actor reminder_ api validate the response and only if a reminder is active utilize the _UnregisterReminder()_ method. So it would be helpful if the _Get actor reminder_ api could also be surfaced via sdk.

shivamkm07:
@halspang This issue can be closed now.

fabistb:
@shivamkm07 thank you very much!
Just tried it and returned all the information I'm interested in.

ssm3ll:
> @fabistb @ssm3ll Get Actor Reminder API has been added in Dotnet sdk in 1.11. `GetActorReminder` returns `IActorReminder` Object if the reminder exists, otherwise null. Feel free to try this and let us know in case of any issues. P.S. I will be creating Docs PR soon for the change.

thank you. it works

olitomlinson:
[...]

PR #1103, opened 2023-06-03
adding get actor reminder API

shivamkm07:
# Description

It adds the get actor reminder API in dotnet-sdk

## Issue reference

Please reference the issue this PR will close: #772 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/src/Dapr.Actors/Runtime/Actor.cs
+++ b/src/Dapr.Actors/Runtime/Actor.cs
@@ -361,6 +361,18 @@ namespace Dapr.Actors.Runtime
         }
 
         /// <summary>
+        /// Gets a reminder previously registered using <see cref="Dapr.Actors.Runtime.Actor.RegisterReminderAsync(ActorReminderOptions)" />.
+        /// </summary>
+        /// <param name="reminderName">The name of the reminder to get.</param>
+        /// <returns>
+        /// Returns a task that represents the asynchronous get operation. The result of the task contains the reminder if it exists, otherwise null.
+        /// </returns>
+        protected async Task<IActorReminder> GetReminderAsync(string reminderName)
+        {
+            return await this.Host.TimerManager.GetReminderAsync(new ActorReminderToken(this.actorTypeName, this.Id, reminderName));
+        }
+
+        /// <summary>
         /// Unregisters a reminder previously registered using <see cref="Dapr.Actors.Runtime.Actor.RegisterReminderAsync(ActorReminderOptions)" />.
         /// </summary>
         /// <param name="reminder">The actor reminder to unregister.</param>
--- a/src/Dapr.Actors/Runtime/ActorTimerManager.cs
+++ b/src/Dapr.Actors/Runtime/ActorTimerManager.cs
@@ -28,6 +28,13 @@ namespace Dapr.Actors.Runtime
         public abstract Task RegisterReminderAsync(ActorReminder reminder);
 
         /// <summary>
+        /// Gets a reminder previously registered using
+        /// </summary>
+        /// <param name="reminder">The <see cref="ActorReminderToken" /> to unregister.</param>
+        /// <returns>A task which will complete when the operation completes.</returns>
+        public abstract Task<IActorReminder> GetReminderAsync(ActorReminderToken reminder);
+
+        /// <summary>
         /// Unregisters the provided reminder with the runtime.
         /// </summary>
         /// <param name="reminder">The <see cref="ActorReminderToken" /> to unregister.</param>
--- a/src/Dapr.Actors/Runtime/DefaultActorTimerManager.cs
+++ b/src/Dapr.Actors/Runtime/DefaultActorTimerManager.cs
@@ -98,14 +98,11 @@ namespace Dapr.Actors.Runtime
             {
                 throw new ArgumentNullException(nameof(stream));
             }
-            var streamContents = new byte[stream.Length];
-            stream.Read(streamContents);
-            if(Encoding.ASCII.GetString(streamContents) == "null")
+            var info = await ReminderInfo.DeserializeAsync(stream);
+            if(info == null)
             {
                 return null;
             }
-            stream.Seek(0, SeekOrigin.Begin);
-            var info = await ReminderInfo.DeserializeAsync(stream);
             var reminder = new ActorReminder(token.ActorType, token.ActorId, token.Name, info.Data, info.DueTime, 
                 info.Period);
             return reminder;
--- a/src/Dapr.Actors/Runtime/DefaultActorTimerManager.cs
+++ b/src/Dapr.Actors/Runtime/DefaultActorTimerManager.cs
@@ -14,6 +14,8 @@
 using System;
 using System.Text.Json;
 using System.Threading.Tasks;
+using System.IO;
+using System.Text;
 
 namespace Dapr.Actors.Runtime
 {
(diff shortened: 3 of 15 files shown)


src/Dapr.Actors/Runtime/ReminderInfo.cs
```diff
@@ -55,28 +55,29 @@ internal static async Task<ReminderInfo> DeserializeAsync(Stream stream)
...
-            if (json.TryGetProperty("data", out var dataProperty) && dataProperty.ValueKind != JsonValueKind.Null)
-            {
-                data = dataProperty.GetBytesFromBase64();
-            }
-
-            if (json.TryGetProperty("ttl", out var ttlProperty))
-            {
-                var ttlString = ttlProperty.GetString();
-                ttl = ConverterUtils.ConvertTimeSpanFromDaprFormat(ttlString);
+            if(json.ValueKind != JsonValueKind.Null){
```
philliphoff, line 58: Would it be more accurate to say?: `if (json.ValueKind == JsonValueKind.Object)`

philliphoff, line 58: `DeserializeAsync<T>()` returns `object?`; would it make sense to also test `json` for `null`?

src/Dapr.Actors/Runtime/Actor.cs
```diff
@@ -360,6 +360,18 @@ internal async Task<IActorReminder> RegisterReminderAsync(ActorReminderOptions o
...
         }
 
+        /// <summary>
+        /// Gets a reminder previously registered using <see cref="Dapr.Actors.Runtime.Actor.RegisterReminderAsync(ActorReminderOptions)" />.
+        /// </summary>
+        /// <param name="reminderName">The name of the reminder to get.</param>
+        /// <returns>
+        /// Returns a task that represents the asynchronous get operation. The result of the task contains the reminder if it exists, otherwise null.
+        /// </returns>
+        protected async Task<IActorReminder> GetReminderAsync(string reminderName)
```
philliphoff, line 370: This method states it returns `null` if no timer exists, but the underlying call to the `DaprHttpInteractor` appears to throw if the HTTP request does not return a successful response (and the docs say the Actor Get Reminder API returns only `200` and `500` responses.

src/Dapr.Actors/Runtime/ActorTimerManager.cs
```diff
@@ -27,6 +27,13 @@ public abstract class ActorTimerManager
         /// <returns>A task which will complete when the operation completes.</returns>
         public abstract Task RegisterReminderAsync(ActorReminder reminder);
 
+        /// <summary>
+        /// Gets a reminder previously registered using
+        /// </summary>
+        /// <param name="reminder">The <see cref="ActorReminderToken" /> to unregister.</param>
+        /// <returns>A task which will complete when the operation completes.</returns>
+        public abstract Task<IActorReminder> GetReminderAsync(ActorReminderToken reminder);
```
philliphoff, line 35: The underlying `IDaprInteractor` allows for cancellation, but the higher-level types don't seem to make use of it.  Should we start doing so for new APIs?  It'd likely be breaking to do so as a whole, but one has to start somewhere.

src/Dapr.Actors/Runtime/DefaultActorTimerManager.cs
```diff
@@ -77,5 +91,24 @@ private async ValueTask<string> SerializeReminderAsync(ActorReminder reminder)
...
+
+        private async ValueTask<ActorReminder> DeserializeReminderAsync(Stream stream, ActorReminderToken token)
+        {
+            if (stream == null)
+            {
+                throw new ArgumentNullException(nameof(stream));
+            }
+            var streamContents = new byte[stream.Length];
+            stream.Read(streamContents);
+            if(Encoding.ASCII.GetString(streamContents) == "null")
```
philliphoff, line 103: Would the JSON deserializer not catch this case?  In particular, can we assume `ASCII` encoding vs. `UTF8`?

src/Dapr.Actors/Runtime/DefaultActorTimerManager.cs
```diff
@@ -77,5 +91,24 @@ private async ValueTask<string> SerializeReminderAsync(ActorReminder reminder)
...
+            {
+                return null;
+            }
+            stream.Seek(0, SeekOrigin.Begin);
```
philliphoff, line 107: There are no guarantees that the `Stream` provided supports seeking; if we really need to do this, the capability should be verified and, if not, a copy of the stream (e.g. a `MemoryStream`) should be made.  If we were to know the JSON deserializer handles `null`, then it seems we would have no need for this.

src/Dapr.Actors/Runtime/ReminderInfo.cs
shivamkm07: Here we have been using `DeserializeAsync<JsonElement>` which returns `JsonElement` type, which is not a reference type and so can't be compared to null. It says this error on trying: `Operator '!=' cannot be applied to operands of type 'JsonElement' and '<null>'`.

src/Dapr.Actors/Runtime/Actor.cs
shivamkm07: In case the reminder doesn't exist, the HTTP call TO Dapr returns a successful status code with a response {"null"}. This is being handled in the `DeserializeReminderAsync` function in ActorTimerManager(https://github.com/dapr/dotnet-sdk/pull/1103/files#diff-89d1bcb06652638fdfeaa1b8146ada6ea2d8fe4ff18216de0928734712473cf4R103), in which we actually return a null object. Also the same scenario is added in E2E test as well, testing both cases when the reminder doesn't exist and when reminder exists.
The E2E test case for no reminder can be found here: https://github.com/dapr/dotnet-sdk/pull/1103/files#diff-6c279ec059b177b7a7e54e34eda712f299033fb2321fac66836f93a78c76b5d5R62

src/Dapr.Actors/Runtime/DefaultActorTimerManager.cs
shivamkm07: No the Json Deserializer actually checks for the values in the stream and if they are not there, it assigns a default value for some(e.g. TimeSpan,Period). Also note that, in case a reminder is not there, `stream` is not null, instead it's actually has a string "null", which is what Dapr returns when the reminder is not there. So this is required for handling the case when there is no reminder.

shivamkm07: Handled the case in JsonDeserializer only. That's a much better and cleaner way. Thanks!

src/Dapr.Actors/Runtime/DefaultActorTimerManager.cs
shivamkm07: Removed the Seeking code, and handled the case in JsonSerializer itself.

src/Dapr.Actors/Runtime/Actor.cs
shivamkm07: If the reminder doesn't exist, Dapr returns a suceessful status code with {null} response. That scenario is already handled in Json Serializer and so in case a reminder doesn't exist, this function returns a null value only.

Alos, the same has been tested in added E2E test as well and it is passing as well: https://github.com/dapr/dotnet-sdk/pull/1103/files#diff-6c279ec059b177b7a7e54e34eda712f299033fb2321fac66836f93a78c76b5d5R61

src/Dapr.Actors/Runtime/ReminderInfo.cs
shivamkm07: Handled null scenario differently.

shivamkm07: Handled null scenario differently

philliphoff (approved):
It might be good to more clearly document the behavior of Dapr when retrieving actor reminders that do not exist; a `200` response with a `null` body doesn't seem intuitive to me as opposed to, say, a `404` response.

yaron2 (approved):

artursouza:
Why is this feature needed? We don't have it in other SDKs.

shivamkm07:
> Why is this feature needed? We don't have it in other SDKs.

It was an ask from customers: https://github.com/dapr/dotnet-sdk/issues/772

PR #1119, opened 2023-07-03
[Workflow] Fix issue with ignored external event payload

cgillum:
# Description

In https://github.com/dapr/dapr/issues/6614, a user observed that the `RaiseEventAsync` method was always resulting in a null payload in the workflow code. This PR fixes the issue, which is isolated to the Dapr Workflow SDK.

I also made a couple other small changes as part of this PR:

* Bumped the Durable Task SDK dependency to the latest patch release.
* Changed the default `getInputsAndOutputs` to `true` for the workflow status query APIs (I decided to do this to improve usability after some frustration while setting up the local repro for the raise event issue)

I also updated the Workflow SDK nuget version to `0.3.0`.

## Issue reference

Closes https://github.com/dapr/dapr/issues/6614

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/src/Dapr.Workflow/DaprWorkflowClient.cs
+++ b/src/Dapr.Workflow/DaprWorkflowClient.cs
@@ -135,7 +135,7 @@ namespace Dapr.Workflow
         /// <inheritdoc cref="WaitForWorkflowStartAsync(string, bool, CancellationToken)"/>
         public async Task<WorkflowState> WaitForWorkflowCompletionAsync(
             string instanceId,
-            bool getInputsAndOutputs = false,
+            bool getInputsAndOutputs = true,
             CancellationToken cancellation = default)
         {
             OrchestrationMetadata metadata = await this.innerClient.WaitForInstanceCompletionAsync(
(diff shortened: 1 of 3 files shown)


src/Dapr.Workflow/DaprWorkflowClient.cs
```diff
@@ -70,9 +70,9 @@ public DaprWorkflowClient(DurableTaskClient innerClient)
         /// <param name="instanceId">The unique ID of the workflow instance to fetch.</param>
         /// <param name="getInputsAndOutputs">
         /// Specify <c>true</c> to fetch the workflow instance's inputs, outputs, and custom status, or <c>false</c> to
-        /// omit them. Defaults to false.
+        /// omit them. Defaults to true.
         /// </param>
-        public async Task<WorkflowState> GetWorkflowStateAsync(string instanceId, bool getInputsAndOutputs = false)
+        public async Task<WorkflowState> GetWorkflowStateAsync(string instanceId, bool getInputsAndOutputs = true)
```
halspang, line 75: Just checking, but I assume this doesn't change the return payload in some way that isn't backwards compatible? We'd just get additional info in it, yeah?

cgillum: Thanks for checking. Yes, this change is backwards compatible. The only difference is that we add new data which was previously left as null.

halspang:
@cgillum - You have test failures that I think are related to legacy steps in the itests workflow. Can you remove these steps and repush to see if that fixes things?

https://github.com/dapr/dotnet-sdk/blob/master/.github/workflows/itests.yml#L108-L116

halspang (approved):

Issue #1134, opened 2023-08-16
[Workflow] Workflow fails when ApiToken is passed into builder when constructing Dapr Client


RyanLettieri:
If you try to use Workflows with a Dapr instance that has API authentication enabled, you will fail since the workflow client can't pass the auth token.

Note that this an issue in the underlying durable task framework so it will need a fix in all SDKs for workflow.

## Expected Behavior

Expected to have the workflow run with no issues.

## Actual Behavior
App continously outputs the following failure statements:

== APP == fail: Microsoft.DurableTask[20]
== APP ==       Unexpected error in handling of instance ID ''.
== APP ==       Grpc.Core.RpcException: Status(StatusCode="Unauthenticated", Detail="Unauthorized")
== APP ==          at Microsoft.DurableTask.Worker.Grpc.GrpcDurableTaskWorker.Processor.ConnectAsync(CancellationToken cancellation)
== APP ==          at Microsoft.DurableTask.Worker.Grpc.GrpcDurableTaskWorker.Processor.ExecuteAsync(String target, CancellationToken cancellation)

## Steps to Reproduce the Problem

Open the example program for the workflow console app inside the .NET SDK repo.
Add the following to the line where we construct the daprClient (before .Build()): "UseDaprApiToken("apiToken")" (https://github.com/dapr/dotnet-sdk/blob/f4e02df980072438ad225b75412f779df67936a0/examples/Workflow/WorkflowConsoleApp/Program.cs#L50)
Run the example

## Release Note

RELEASE NOTE:

PR #1141, opened 2023-08-30
Updating workflow collection to allow for use of API Token validation

RyanLettieri:
# Description

Updated the service collection for workflow in the .NET SDK to allow for the use of api token authentication. Since the durable task framework allows for the passing in of a gRPC channel when configuring the gRPC client, we now create a gRPC channel object to have the API Token headers as well as the address we want to use and pass that through to the DTFx.

## Issue reference

The issue this PR will close: #_[[1134](https://github.com/dapr/dotnet-sdk/issues/1134)]_

## Checklist

* [x ] Code compiles correctly
* [ ] Created/updated tests


src/Dapr.Workflow/WorkflowServiceCollectionExtensions.cs
```diff
@@ -126,6 +152,40 @@ static bool TryGetGrpcAddress(out string address)
...
+        }
+    }
+
+    static class DaprDurableGrpcExtensions
+    {
+        const string ClientName = "durabletask-grpc";
+
+        public static IServiceCollection ConfigureDurableGrpcClient(this IServiceCollection services, string apiToken)
+        {
+            services.AddHttpClient(ClientName)
```
DeepanshuA, line 181: For my understanding: Why do we need to add HttpClient while configuring Grpc Client?

halspang: Under the hood, the gRPC client uses the HttpClient. It's all HTTP/2 for transport iirc.

examples/Workflow/WorkflowConsoleApp/Program.cs
```diff
@@ -47,7 +47,16 @@
...
+DaprClient daprClient;
+string apiToken = Environment.GetEnvironmentVariable("DAPR_API_TOKEN");
+if (!string.IsNullOrEmpty(apiToken))
+{
+    daprClient = new DaprClientBuilder().UseDaprApiToken(apiToken).Build();
+}
+else
+{
+    daprClient = new DaprClientBuilder().Build();
+}
```
halspang, line 59: We still need the `using`. After the initialization, you can just wrap the `while` in this:

```csharp
using (daprClient)
{
    // Loop here.
}
```

src/Dapr.Workflow/WorkflowServiceCollectionExtensions.cs
```diff
@@ -57,7 +59,19 @@ public static class WorkflowServiceCollectionExtensions
 
                 if (TryGetGrpcAddress(out string address))
                 {
-                    builder.UseGrpc(address);
+                    string? apiToken = Environment.GetEnvironmentVariable("DAPR_API_TOKEN");
```
halspang, line 62: For consistency, please use `var` where you can.

src/Dapr.Workflow/WorkflowServiceCollectionExtensions.cs
```diff
@@ -57,7 +59,19 @@ public static class WorkflowServiceCollectionExtensions
...
+                    if (!string.IsNullOrEmpty(apiToken))
+                    {
+                        serviceCollection.ConfigureDurableGrpcClient(apiToken);
+                        HttpClient client = new HttpClient();
+                        client.DefaultRequestHeaders.Add("Dapr-Api-Token", apiToken);
+                        builder.UseGrpc(CreateChannel(client));
```
halspang, line 68: This seems to repeat the code in the `ConfigureDurableGrpcClient` do we need both methods?

halspang (changes requested):
Can you update the example readme with how to use the API token?

src/Dapr.Workflow/WorkflowServiceCollectionExtensions.cs
```diff
@@ -57,7 +59,19 @@ public static class WorkflowServiceCollectionExtensions
...
                 if (TryGetGrpcAddress(out string address))
                 {
-                    builder.UseGrpc(address);
+                    var apiToken = Environment.GetEnvironmentVariable("DAPR_API_TOKEN");
+                    if (!string.IsNullOrEmpty(apiToken))
+                    {
+                        // serviceCollection.ConfigureDurableGrpcClient(apiToken);
+                        HttpClient client = new HttpClient();
+                        client.DefaultRequestHeaders.Add("Dapr-Api-Token", apiToken);
+                        builder.UseGrpc(CreateChannel(client));
```
yaron2, line 68: The `address` variable should be passed to the builder, otherwise if users configured Dapr to use a non-localhost endpoint, this will not be propagated correctly to the underlying Durable Task client. Good find by @alicejgibbons

yaron2 (changes requested):

src/Dapr.Workflow/WorkflowServiceCollectionExtensions.cs
```diff
@@ -57,7 +59,18 @@ public static class WorkflowServiceCollectionExtensions
 
                 if (TryGetGrpcAddress(out string address))
                 {
-                    builder.UseGrpc(address);
+                    var apiToken = Environment.GetEnvironmentVariable("DAPR_API_TOKEN");
```
famarting, line 62: there is this function https://github.com/dapr/dotnet-sdk/blob/master/src/Shared/DaprDefaults.cs#L29 which you can use to read the DAPR_API_TOKEN env var and avoid duplicating its name multiple times DAPR_API_TOKEN

src/Dapr.Workflow/WorkflowServiceCollectionExtensions.cs
```diff
@@ -14,16 +14,21 @@
...
     using Microsoft.Extensions.DependencyInjection.Extensions;
+    using System.Net.Http;
+    using Dapr;
 
     /// <summary>
     /// Contains extension methods for using Dapr Workflow with dependency injection.
     /// </summary>
     public static class WorkflowServiceCollectionExtensions
     {
+        private static string daprApiToken = "";
```
halspang, line 30: Why did we add this is a private static field? `DaprDefaults` does the lazy load for us so we have no need to create a static variable here.

src/Dapr.Workflow/WorkflowServiceCollectionExtensions.cs
```diff
@@ -57,7 +62,18 @@ public static class WorkflowServiceCollectionExtensions
 
                 if (TryGetGrpcAddress(out string address))
                 {
-                    builder.UseGrpc(address);
+                    daprApiToken = DaprDefaults.GetDefaultDaprApiToken();
+                    if (!string.IsNullOrEmpty(daprApiToken))
+                    {
+                        HttpClient client = new HttpClient();
```
halspang, line 68: Nit: `var` consistency.

src/Dapr.Workflow/WorkflowServiceCollectionExtensions.cs
```diff
@@ -85,7 +101,18 @@ public static IServiceCollection AddDaprWorkflowClient(this IServiceCollection s
             {
                 if (TryGetGrpcAddress(out string address))
                 {
-                    builder.UseGrpc(address);
+                    daprApiToken = DaprDefaults.GetDefaultDaprApiToken();
+                    if (!string.IsNullOrEmpty(daprApiToken))
+                    {
+                        HttpClient client = new HttpClient();
```
halspang, line 107: Nit: `var` consistency.

halspang (changes requested):
One question and a few nits. Did you ensure that the various places where we call the `DaprDefault` methods all handle an empty/null case instead of just a null case? It shouldn't be too many places to check.

halspang (approved):

yaron2 (approved):

PR #1148, opened 2023-09-21
Throw better exceptions when invalid actor interfaces are used.

### LINKED ISSUE #859: Cryptic TypeLoadException if actor proxy interface is non-public
Consider the following code snippet:

```csharp
    ActorId actorId = new("123");
    IMyActor proxy = ActorProxy.Create<IMyActor>(actorId, nameof(MyActor));

    // The Init method invokes the actor directly, which then decides whether to apply de-dupe logic.
    await proxy.MyMethod();
```

Also, assume `IMyActor` is an internal interface (non-public).

## Expected Behavior

Internal interfaces should be supported, OR there should be a friendly exception explaining why my interface is invalid.

## Actual Behavior

If `IMyActor` is internal, I get the following cryptic exception:

> System.TypeLoadException: Type 'MyNamespace.IMyActor_.actorV1.proxyIMyActoractorV1Proxy' from assembly 'MyNamespace.IMyActor_.actorV1.proxy, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' is attempting to implement an inaccessible interface.

It's not at all clear what needs to be done to fix this issue.

philliphoff:
# Description

Verifies that actor interfaces used to create actor proxies actually:

 - Inherit from `IActor`
 - Are publicly-visible

When interfaces are used that do not meet those requirements, a less cryptic error message is generated.

## Issue reference

Please reference the issue this PR will close: #859

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


src/Dapr.Actors/Client/ActorProxyFactory.cs
```diff
@@ -14,6 +14,7 @@
 namespace Dapr.Actors.Client
 {
     using System;
+    using System.Globalization;
```
halspang, line 17: Since we're here, do you mind moving these imports outside of the namespace?

philliphoff: Sure!

WhitWaldo:
As the `Dapr.Actors` package is in maintenance now in favor of the newer [`Dapr.Actors.Next`](https://github.com/dapr/dotnet-sdk/releases#release-v1.19.0-preview.2) package, I'm going to go ahead and close this. I added a PR recently to `Dapr.Actors.Next` that expands on the exceptions thrown to make them more specific and useful (as opposed to generic `InvalidDataOperationExceptions` and the like).

PR #1153, opened 2023-09-25
Support case insensitive cloudevent payloads and forward cloudevent props s headers

IliasP91:
# Description

Extend the existing CloudEvent Middleware to:
- Accept case insensitive cloudevent payloads
-  Extend its options to forward cloudevent props as HTTP Request.Headers

This will make the middleware accept JSON payloads that are Pascal cases, as the default ASP Net core options already do. 
And add the option to forward all the cloud event properties that are not automatically forwarded by the Dapr sidecar as headers. I had a use case where I wanted to read the type of the event (not just filter it) in my handler and currently there is no way of doing it as the middleware just unfolds the cloudevent envelope and discards it.

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation (NA in this case there's no section that describes the `UseCloudEvents()` method, happy to create one of you think its valuable. I have extended one of the examples though)


src/Dapr.AspNetCore/CloudEventsMiddleware.cs
```diff
@@ -83,17 +95,29 @@ private async Task ProcessBodyAsync(HttpContext httpContext, string charSet)
...
 
             // Check whether to use data or data_base64 as per https://github.com/cloudevents/spec/blob/v1.0.1/json-format.md#31-handling-of-data
-            var isDataSet = json.TryGetProperty("data", out var data);
-            var isBinaryDataSet = json.TryGetProperty("data_base64", out var binaryData);
+            // Get the property names by OrdinalIgnoreCase comparison to support case insensitive JSON as the Json Serializer for AspCore already supports it by default.
+            var jsonPropNames = json.EnumerateObject().ToArray();
+
+            var dataPropName = jsonPropNames
+                .Select(d => d.Name)
+                .FirstOrDefault(d => d.Equals("data", StringComparison.OrdinalIgnoreCase)) ?? "";
```
philliphoff, line 103: Would it make sense to make `"data"` and `"data_base64"` constants as they're used in multiple places?

src/Dapr.AspNetCore/CloudEventsMiddlewareOptions.cs
```diff
@@ -29,9 +29,20 @@ public class CloudEventsMiddlewareOptions
...
         /// </remarks>
         public bool SuppressJsonDecodingOfTextPayloads { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value that will determine whether the CloudEvent properties will be forwarded as Request Headers.
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// Setting this property to <c>true</c> will forward the CloudEvent properties as Request Headers in the following format.
+        /// ie. A CloudEvent property <c>"type": "Example.Type"</c> will be added as <c>"Cloudevent.type": "Example.Type"</c> request header.
```
philliphoff, line 43: Is it also worth noting that the header-forwarded property names will be lower-cased as request header names?

IliasP91: good spot i missed this after some debugging and refactoring

src/Dapr.AspNetCore/CloudEventsMiddleware.cs
IliasP91: yeah definitely, ill create a class with the constants or something like that

src/Dapr.AspNetCore/CloudEventsMiddleware.cs
```diff
@@ -52,7 +63,7 @@ public Task InvokeAsync(HttpContext httpContext)
             // The philosophy here is that we don't report an error for things we don't support, because
             // that would block someone from implementing their own support for it. We only report an error
             // when something we do support isn't correct.
-            if (!this.MatchesContentType(httpContext, out var charSet))
+            if (!MatchesContentType(httpContext, out var charSet))
```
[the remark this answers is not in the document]
IliasP91: no problem, my rider code cleanup configuration removes it automatically, but ill put it back to keep it consistent

src/Dapr.AspNetCore/CloudEventsMiddleware.cs
```diff
@@ -83,17 +95,29 @@ private async Task ProcessBodyAsync(HttpContext httpContext, string charSet)
...
+            // Get the property names by OrdinalIgnoreCase comparison to support case insensitive JSON as the Json Serializer for AspCore already supports it by default.
+            var jsonPropNames = json.EnumerateObject().ToArray();
+
+            var dataPropName = jsonPropNames
+                .Select(d => d.Name)
+                .FirstOrDefault(d => d.Equals("data", StringComparison.OrdinalIgnoreCase)) ?? "";
+
+            var dataBase64PropName = jsonPropNames
+                .Select(d => d.Name)
+                .FirstOrDefault(d => d.Equals("data_base64", StringComparison.OrdinalIgnoreCase)) ?? "";
```
philliphoff, line 107: AFAIK, JSON property names can be empty strings.  Is it really intended to use `""` as a property name should `"data_base64"` (or equivalent) not be found?  Should the empty string just be omitted and allow the property name to be `null`, in which case the following tests could be:

```csharp
var isDataSet = dataPropName is not null && json.TryGetProperty(dataPropName, out var data);
var isBinaryDataSet = dataBase64PropName is not null && json.TryGetProperty(dataBase64PropName, out var binaryData);
```

src/Dapr.AspNetCore/CloudEventsMiddleware.cs
```diff
@@ -27,6 +30,14 @@ namespace Dapr
     internal class CloudEventsMiddleware
     {
         private const string ContentType = "application/cloudevents+json";
+
+        // These cloudevent properties are either containing the body of the message or 
+        // are included in the headers by other components of Dapr earlier in the pipeline 
+        private static readonly string[] ExcludedPropertiesFromHeaders =
```
philliphoff, line 36: Might there be other properties that users want excluded for size/performance reasons?  Would it make sense for the options object to offer a property to set additional exclusions and/or included properties?

src/Dapr.AspNetCore/CloudEventsMiddleware.cs
```diff
@@ -148,16 +174,39 @@ private async Task ProcessBodyAsync(HttpContext httpContext, string charSet)
...
         {
+            if (!options.ForwardCloudEventPropertiesAsHeaders)
+            {
+                return;
+            }
+
+            foreach (var jsonProperty in jsonPropNames
+                         .Where(d => !ExcludedPropertiesFromHeaders.Contains(d.Name.ToLowerInvariant())))
+            {
+                httpContext.Request.Headers.TryAdd($"Cloudevent.{jsonProperty.Name.ToLowerInvariant()}",
```
philliphoff, line 189: Is it certain that all string-ified JSON values are valid HTTP header values?  (I.e. are there characters, for example, that might exist in JSON that are not allowed in headers?) If not, it might strengthen the argument for allowing more fine-grained forwarding.

src/Dapr.AspNetCore/CloudEventsMiddleware.cs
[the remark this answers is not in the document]
IliasP91: actually t his specific one was removed as the MatchesContentType is now a static method, and this. cant be used

src/Dapr.AspNetCore/CloudEventsMiddleware.cs
IliasP91: hmm this was under the assumption that the cloud event wont have any empty props but you are correct, there's nothing stopping the payload to have one

src/Dapr.AspNetCore/CloudEventsMiddleware.cs
IliasP91: that is an interesting idea, ill think about it for a bit
maybe two complementary optional lists with included and excluded property names will make it a lot more flexible
for cases where users might want to either forward a,b,c props only, or forward everything except a,b,c

src/Dapr.AspNetCore/CloudEventsMiddleware.cs
IliasP91: This is a very interesting observation.
In our services I have been using a similarly modified middleware to forward them, and it has been working just fine, but they were plain utf8 text, however after doing a quick search some special characters seem to not be accepted in HTTP headers.
I wonder if that validation happens earlier or later in the middleware pipeline, which could affect this or not
that said, I would be concerned if engineers posted unencoded special characters in JSON payloads in the first place, as some brokers like Kafka or RabbitMQ etc might not support them.

More fine grained control on which props we should forward is definitely a good idea, I'll refactor it tomorrow

IliasP91: @philliphoff 
How about this proposal:
I'm thinking of adding another flag to optionally URL encode the values that will be put in the headers, to allow people pass around non-ASCI characters if they need to.

IliasP91: After some more reading it seems that AspNetCore URL encodes/decodes automatically UTF-8 characters that are not ASCI, I will add some tests to try it out just in case we need to explicitly encode the values

IliasP91: I can confirm that non ASCI characters are forwarded just fine in the headers an the body without any changes required
image
However I think two optional filter lists could be beneficial so ill proceed and add them as well

src/Dapr.AspNetCore/CloudEventsMiddleware.cs
```diff
@@ -83,17 +96,43 @@ private async Task ProcessBodyAsync(HttpContext httpContext, string charSet)
...
+            var dataBase64PropName = jsonPropNames
+                .Select(d => d.Name)
+                .FirstOrDefault(d =>
+                    d.Equals(CloudEventPropertyNames.DataBase64, StringComparison.OrdinalIgnoreCase));
+
+            var isDataSet = false;
+            var isBinaryDataSet = false;
+            JsonElement data = default;
+
+            if (!string.IsNullOrWhiteSpace(dataPropName))
```
philliphoff, line 115: Nit: as the name of the property is hidden behind a constant, we don't known, here, for certain, that whitespace doesn't indicate a valid property.  Testing only for `null` or non-`null` seems more appropriate.

src/Dapr.AspNetCore/CloudEventsMiddleware.cs
```diff
@@ -148,16 +189,55 @@ private async Task ProcessBodyAsync(HttpContext httpContext, string charSet)
...
+        }
+
+        private static string GetDataContentType(JsonElement json, out bool isJson)
+        {
+            var dataContentTypePropName = json.EnumerateObject()
+                                              .Select(d => d.Name)
+                                              .FirstOrDefault(d =>
+                                                  d.Equals(CloudEventPropertyNames.DataContentType,
+                                                      StringComparison.OrdinalIgnoreCase))
+                                          ?? "";
```
philliphoff, line 231: Similar nit here as before; I don't see a need to use `""` as a default; just test whether `dataContentTypePropName` is non-`null` before calling `TryGetProperty()` in the following `if` statement.

src/Dapr.AspNetCore/CloudEventsMiddleware.cs
IliasP91: I can't say I agree on this one, first as you said its a known value stored in a constant, which is immutable, it cant be changed other than someone in a PR, which should be alarming to be honest.
And generally when dealing with strings, its a best practice to check against all the three possible "empty" values, null, empty string and whitespace character, which is what `IsNullOrWhiteSpace` does.
That said, it wont affect the functionality in this case so I wont die over this, but it will need to be updated when the project starts using nullable ref types, which is something that surprised me. I would expect it to already be using them since its targeting .NET 6+.

src/Dapr.AspNetCore/CloudEventsMiddleware.cs
IliasP91: On this one I agree, we don't want to pick the wrong JSON field, must have missed it when I changed the other ones above.
I'll change it to use `string.IsNullOrWhiteSpace` first thing in the if, based on the argument in the above comment.

src/Dapr.AspNetCore/CloudEventsMiddleware.cs
philliphoff: > And generally when dealing with strings, its a best practice to check against all the three possible "empty" values, null, empty string and whitespace character, which is what IsNullOrWhiteSpace does.

In this case, however, both the empty string and a white-space-only string are technically valid property names that could be returned by the prior search.  We know ourselves that those values are unlikely to be the actual names of the property being searched for but, as they're hidden behind the constant, the test should not make such assumptions.

> I would expect it to already be using them since its targeting .NET 6+.

Adding nullable annotations has been in the issues list for a while, but it's a pretty sizable engineering task.  The switch to .NET 6/7 is fairly recent so it will take time to start using the more modern language conveniences a large scale.

IliasP91: > > And generally when dealing with strings, its a best practice to check against all the three possible "empty" values, null, empty string and whitespace character, which is what IsNullOrWhiteSpace does.
> 
> In this case, however, both the empty string and a white-space-only string are technically valid property names that could be returned by the prior search. We know ourselves that those values are unlikely to be the actual names of the property being searched for but, as they're hidden behind the constant, the test should not make such assumptions.
> 
I agree on the principle, however the value being in the constant compared to being in a string visible right there makes no difference to me. In both cases it is hardcoded in the codebase and immutable in runtime. If a value has matched and returns it will be the same as the value in the constant, with the only difference on its casing, which is ignored. It cannot return an empty string or whitespace. The only difference is this way it can be reused in other places.

> > I would expect it to already be using them since its targeting .NET 6+.
> 
> Adding nullable annotations has been in the issues list for a while, but it's a pretty sizable engineering task. The switch to .NET 6/7 is fairly recent so it will take time to start using the more modern language conveniences a large scale.

Yeah that is a valid point, it does require quite a bit of refactoring to eliminate all the warnings. And there's quite a few possible NRE already in there.

IliasP91: Alright as I see that this is not treated as a Nit at all, I'll go ahead  and refactor it. However I will leave it here that the argument holds no ground as the purpose of constants is to reuse static values that do not mutate during runtime. And that the whole purpose of those two blocks of code is to grab either the data or data64 properties of the CloudEvent. If anything else returns for whatever reason, the middleware is critically broken.

IliasP91:
hi again, could someone have a look at this PR? I has been sitting there for a few months now...
@dapr/approvers-dotnet-sdk

WhitWaldo (approved):
LGTM!

marcduiker:
@holopin-bot @IliasP91 Thank you Ilias!

PR #1164, opened 2023-10-12
Actor State TTL

JoshVanL:
Dapr Actor state TTL is a feature that was added in Dapr 1.12.

Adds support for Actor State TTL. TTL of actor state is set by using the `ttlInSeconds` metadata option when saving actor state. The actor state cache uses the `ttlExpireTime` return metadata field to determine whether the key is still valid, which is important as the cache may become populated with `TTL` keys which were not created by the current manager.

SDK developers are encouraged to use TTLs when storing any actor state in order to prevent stale data accumulating in the actor state store over time.

Closes https://github.com/dapr/dotnet-sdk/pull/1164

### DIFF
--- a/src/Dapr.Actors/Communication/ActorStateResponse.cs
+++ b/src/Dapr.Actors/Communication/ActorStateResponse.cs
@@ -1,0 +1,50 @@
+﻿// ------------------------------------------------------------------------
+// Copyright 2023 The Dapr Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//     http://www.apache.org/licenses/LICENSE-2.0
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// ------------------------------------------------------------------------
+
+namespace Dapr.Actors.Communication
+{
+    using System;
+
+    /// <summary>
+    /// Represents a response from fetching an actor state key.
+    /// </summary>
+    internal class ActorStateResponse<T>
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ActorStateResponse{T}"/> class.
+        /// </summary>
+        /// <param name="value">The response value.</param>
+        /// <param name="ttlExpireTime">The time to live expiration time.</param>
+        public ActorStateResponse(T value, DateTime? ttlExpireTime)
+        {
+            this.Value = value;
+            this.TTLExpireTime = ttlExpireTime;
+        }
+
+        /// <summary>
+        /// Gets the response value as a string.
+        /// </summary>
+        /// <value>
+        /// The response value as a string.
+        /// </value>
+        public T Value { get; }
+
+        /// <summary>
+        /// Gets the time to live expiration time.
+        /// </summary>
+        /// <value>
+        /// The time to live expiration time.
+        /// </value>
+        public DateTime? TTLExpireTime { get; }
+    }
+}
--- a/src/Dapr.Actors/Communication/ActorStateResponse.cs
+++ b/src/Dapr.Actors/Communication/ActorStateResponse.cs
@@ -18,7 +18,7 @@ namespace Dapr.Actors.Communication
     /// <summary>
     /// Represents a response from fetching an actor state key.
     /// </summary>
-    class ActorStateResponse<T>
+    public class ActorStateResponse<T>
     {
         /// <summary>
         /// Initializes a new instance of the <see cref="ActorStateResponse{T}"/> class.
--- a/src/Dapr.Actors/Communication/ActorStateResponse.cs
+++ b/src/Dapr.Actors/Communication/ActorStateResponse.cs
@@ -25,7 +25,7 @@ namespace Dapr.Actors.Communication
         /// </summary>
         /// <param name="value">The response value.</param>
         /// <param name="ttlExpireTime">The time to live expiration time.</param>
-        public ActorStateResponse(T value, DateTime? ttlExpireTime)
+        public ActorStateResponse(T value, DateTimeOffset? ttlExpireTime)
         {
             this.Value = value;
             this.TTLExpireTime = ttlExpireTime;
@@ -45,6 +45,6 @@ namespace Dapr.Actors.Communication
         /// <value>
         /// The time to live expiration time.
         /// </value>
-        public DateTime? TTLExpireTime { get; }
+        public DateTimeOffset? TTLExpireTime { get; }
     }
 }
--- a/src/Dapr.Actors/Communication/ActorStateResponse.cs
+++ b/src/Dapr.Actors/Communication/ActorStateResponse.cs
@@ -18,7 +18,7 @@ namespace Dapr.Actors.Communication
     /// <summary>
     /// Represents a response from fetching an actor state key.
     /// </summary>
-    internal class ActorStateResponse<T>
+    class ActorStateResponse<T>
     {
         /// <summary>
         /// Initializes a new instance of the <see cref="ActorStateResponse{T}"/> class.
--- a/src/Dapr.Actors/DaprHttpInteractor.cs
+++ b/src/Dapr.Actors/DaprHttpInteractor.cs
@@ -57,7 +57,7 @@ namespace Dapr.Actors
             this.httpClient.Timeout = requestTimeout ?? this.httpClient.Timeout;
         }
 
-        public async Task<string> GetStateAsync(string actorType, string actorId, string keyName, CancellationToken cancellationToken = default)
+        public async Task<ActorStateResponse<string>> GetStateAsync(string actorType, string actorId, string keyName, CancellationToken cancellationToken = default)
         {
             var relativeUrl = string.Format(CultureInfo.InvariantCulture, Constants.ActorStateKeyRelativeUrlFormat, actorType, actorId, keyName);
 
@@ -72,7 +72,18 @@ namespace Dapr.Actors
 
             using var response = await this.SendAsync(RequestFunc, relativeUrl, cancellationToken);
             var stringResponse = await response.Content.ReadAsStringAsync();
-            return stringResponse;
+
+            var ttlExpireTime = new DateTime();
+            if (response.Headers.TryGetValues(Constants.TTLResponseHeaderName, out IEnumerable<string> headerValues))
+            {
+                var ttlExpireTimeString = headerValues.First();
+                if (!string.IsNullOrEmpty(ttlExpireTimeString))
+                {
+                    ttlExpireTime = DateTime.Parse(ttlExpireTimeString, CultureInfo.InvariantCulture);
+                }
+            }
+
+            return new ActorStateResponse<string>(stringResponse, ttlExpireTime);
         }
 
         public Task SaveStateTransactionallyAsync(string actorType, string actorId, string data, CancellationToken cancellationToken = default)
--- a/src/Dapr.Actors/DaprHttpInteractor.cs
+++ b/src/Dapr.Actors/DaprHttpInteractor.cs
@@ -73,7 +73,7 @@ namespace Dapr.Actors
             using var response = await this.SendAsync(RequestFunc, relativeUrl, cancellationToken);
             var stringResponse = await response.Content.ReadAsStringAsync();
 
-            var ttlExpireTime = new DateTime();
+            DateTimeOffset? ttlExpireTime = null;
             if (response.Headers.TryGetValues(Constants.TTLResponseHeaderName, out IEnumerable<string> headerValues))
             {
                 var ttlExpireTimeString = headerValues.First();
--- a/src/Dapr.Actors/Runtime/ActorStateChange.cs
+++ b/src/Dapr.Actors/Runtime/ActorStateChange.cs
@@ -28,7 +28,7 @@ namespace Dapr.Actors.Runtime
         /// <param name="value">The value associated with given actor state name.</param>
         /// <param name="changeKind">The kind of state change for given actor state name.</param>
         /// <param name="ttlExpireTime">The time to live for the state.</param>
-        public ActorStateChange(string stateName, Type type, object value, StateChangeKind changeKind, DateTime? ttlExpireTime)
+        public ActorStateChange(string stateName, Type type, object value, StateChangeKind changeKind, DateTimeOffset? ttlExpireTime)
         {
             ArgumentVerifier.ThrowIfNull(stateName, nameof(stateName));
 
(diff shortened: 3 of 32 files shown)


src/Dapr.Actors/DaprHttpInteractor.cs
```diff
@@ -57,7 +57,7 @@ internal class DaprHttpInteractor : IDaprInteractor
             this.httpClient.Timeout = requestTimeout ?? this.httpClient.Timeout;
         }
 
-        public async Task<string> GetStateAsync(string actorType, string actorId, string keyName, CancellationToken cancellationToken = default)
+        public async Task<ActorStateResponse<string>> GetStateAsync(string actorType, string actorId, string keyName, CancellationToken cancellationToken = default)
```
JoshVanL, line 60: We want to keep `ActorStateResponse` as an internal type. How do we keep complication happy with this constraint?

src/Dapr.Actors/Runtime/ActorStateManager.cs
```diff
@@ -35,17 +36,17 @@ internal ActorStateManager(Actor actor)
             this.defaultTracker = new Dictionary<string, StateMetadata>();
         }
 
-        public async Task AddStateAsync<T>(string stateName, T value, CancellationToken cancellationToken)
+        public async Task AddStateAsync<T>(string stateName, T value, CancellationToken cancellationToken, int? ttlInSeconds)
```
philliphoff, line 39: Would it make sense to instead add an overload for when the TTL is specified:
`AddStateAsync<T>(string stateName, T value, int ttlInSeconds)` rather than deviate from the .NET guidance that `CancellationToken` be the *last* parameter of a method declaration.  An exception exists to maintain backward compatibility with existing public APIs, but an overload would seem to provide the same effect while maintaining a consistent API.

Also, is it important for TTL to be specified in integer seconds as opposed to, say, a `TimeSpan`, which might be more usable for larger values?  (What do we expect the typical TTL to be?)

src/Dapr.Actors/Communication/ActorStateResponse.cs
```diff
@@ -0,0 +1,50 @@
...
+    /// Represents a response from fetching an actor state key.
+    /// </summary>
+    internal class ActorStateResponse<T>
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ActorStateResponse{T}"/> class.
+        /// </summary>
+        /// <param name="value">The response value.</param>
+        /// <param name="ttlExpireTime">The time to live expiration time.</param>
+        public ActorStateResponse(T value, DateTime? ttlExpireTime)
```
philliphoff, line 28: Consider using `DateTimeOffset` to ensure comparability of expiration times?

JoshVanL: Done :+1:

src/Dapr.Actors/Runtime/ActorStateChange.cs
```diff
@@ -68,5 +70,39 @@ public ActorStateChange(string stateName, Type type, object value, StateChangeKi
...
+        /// <summary>
+        /// Gets the time to live in seconds for the state.
+        /// </summary>
+        /// <value>
+        /// The time to live for the state.
+        /// </value>
+        /// <remarks>
+        /// If null, the state will not expire.
+        /// </remarks>
+        public int? TTLInSeconds {
```
philliphoff, line 94: Since this conversion to seconds is only used in a single place (for the specific circumstances of the runtime API), would it make sense to move this logic there and only expose `TTLExpireTime`?  Is there value in exposing both to users of `ActorStateChange`? (Especially since this property is a constantly changing value, being relative to the current date/time, which users may not expect from a property.)

src/Dapr.Actors/Runtime/ActorStateManager.cs
philliphoff: Right. I was think more of the actual values used for TTLs.  For example, the samples specify a TTL of `360`.  Would that be better expressed as `TimeSpan.FromMinutes(3)` so that users need not do the math (and potentially get it wrong), specially if they have much larger TTLs (e.g. hours/days).

JoshVanL: Gotcha, yes that makes sense I'll add that in.

src/Dapr.Actors/Runtime/ActorStateChange.cs
JoshVanL: Thanks @philliphoff, I've updated the PR to only have hold the `TTLExpireTime` value which gets converted to int seconds when saving the actor state.

src/Dapr.Actors/DaprHttpInteractor.cs
JoshVanL: @philliphoff any ideas here?

philliphoff: `ActorStateResponse` is already `internal`, no, or are you referring to the fact that this method is `public`? `IDaprHttpInteractor` and `DaprrHttpInteractor` are already `internal` types, so nothing of them can be seen outside the SDK.  The `public` on the method only applies to other types within the SDK.  (Method visibility is always masked by visibility of the underlying type.)

src/Dapr.Actors/Runtime/ActorStateManager.cs
```diff
@@ -102,8 +144,8 @@ public async Task<ConditionalValue<T>> TryGetStateAsync<T>(string stateName, Can
             {
                 var stateMetadata = stateChangeTracker[stateName];
 
-                // Check if the property was marked as remove in the cache
-                if (stateMetadata.ChangeKind == StateChangeKind.Remove)
+                // Check if the property was marked as remove in the cache or is expired
+                if (stateMetadata.ChangeKind == StateChangeKind.Remove || stateMetadata.TTLExpireTime <= DateTime.UtcNow)
```
philliphoff, line 148: Nit: `DateTimeOffset.UtcNow`?

I'm virtually certain that the built-in coersion between `DateTime` and `DateTimeOffset` make it equivalent, but perhaps best to use `DateTimeOffset` as consistently as possible.

src/Dapr.Actors/Runtime/ActorStateManager.cs
```diff
@@ -405,9 +551,11 @@ private StateMetadata(object value, Type type, StateChangeKind changeKind)
 
             public Type Type { get; }
 
-            public static StateMetadata Create<T>(T value, StateChangeKind changeKind)
+            public DateTimeOffset? TTLExpireTime { get; set; }
+
+            public static StateMetadata Create<T>(T value, StateChangeKind changeKind, DateTimeOffset? ttlExpireTime = null, TimeSpan? ttl = null)
```
philliphoff, line 556: Is the intent to have a method which allows a TTL, expiry time, or neither?  If so, consider three methods, one which doesn't accept either, one that accepts the offset, and a third that accepts the time span.  That would help ensure proper usage.

src/Dapr.Actors/Runtime/ActorStateManager.cs
```diff
@@ -392,11 +529,20 @@ private void EnsureStateProviderInitialized()
...
             {
                 this.Value = value;
                 this.Type = type;
                 this.ChangeKind = changeKind;
+
+                if (ttlExpireTime.HasValue && ttl.HasValue) {
+                    throw new ArgumentException("Cannot specify both TTLExpireTime and TTL");
+                }
+                if (ttl.HasValue) {
+                    this.TTLExpireTime = DateTime.UtcNow.Add(ttl.Value);
```
philliphoff, line 542: `DateTimeOffset.UtcNow`?  (Doesn't really *need* to be specified in UTC, as sometimes keeping the local time is useful for logging).

src/Dapr.Actors/Runtime/ActorStateManager.cs
philliphoff: Also might be useful to keep a watch on .NET 8 which (finally) adds a [means of abstracting time](https://learn.microsoft.com/en-us/dotnet/api/system.timeprovider?view=net-8.0) to simplify testing.

src/Dapr.Actors/Runtime/DaprStateProvider.cs
```diff
@@ -132,6 +133,12 @@ private async Task DoStateChangesTransactionallyAsync(string actorType, string a
                             writer.WritePropertyName("value");
                             JsonSerializer.Serialize(writer, stateChange.Value, stateChange.Type, jsonSerializerOptions);
                         }
+
+                        if (stateChange.TTLExpireTime.HasValue) {
+                            var ttl = (int)Math.Ceiling((stateChange.TTLExpireTime.Value - DateTime.UtcNow).TotalSeconds);
```
philliphoff, line 138: Just a question, but is it spec'd to use the ceiling (vs. the floor or rounding) for TTLs?

src/Dapr.Actors/Runtime/ActorStateManager.cs
JoshVanL: Thanks! Done

src/Dapr.Actors/Runtime/ActorStateManager.cs
JoshVanL: Yes, that makes sense- they are intended to be mutually exclusive. I've moved to using 3 functions.

src/Dapr.Actors/Runtime/DaprStateProvider.cs
JoshVanL: The spec is that it needs to be an integer value. It makes sense to me to be consistent in how we translate a fractional seconds. Rounding up a fractional second ensures that a value has a TTL of 1 second, rather than 0.

JoshVanL:
@philliphoff do you understand why the integration tests are failing? I'm not too sure what the problem is.

philliphoff:
> @philliphoff do you understand why the integration tests are failing? I'm not too sure what the problem is.

Dunno.  Maybe retry the job in case it was just a transient issue?

JoshVanL:
@philliphoff Doesn't seem to be the case. Struggling to understand what is going wrong here.. at the limit of my .NET knowledge :grimacing:

src/Dapr.Actors/DaprHttpInteractor.cs
```diff
@@ -72,7 +72,18 @@ HttpRequestMessage RequestFunc()
 
             using var response = await this.SendAsync(RequestFunc, relativeUrl, cancellationToken);
             var stringResponse = await response.Content.ReadAsStringAsync();
-            return stringResponse;
+
+            var ttlExpireTime = new DateTime();
```
philliphoff, line 76:
```suggestion
            DateTimeOffset? ttlExpireTime = null;
```

src/Dapr.Actors/Runtime/ActorStateManager.cs
```diff
@@ -102,8 +144,8 @@ public async Task<ConditionalValue<T>> TryGetStateAsync<T>(string stateName, Can
             {
                 var stateMetadata = stateChangeTracker[stateName];
 
-                // Check if the property was marked as remove in the cache
-                if (stateMetadata.ChangeKind == StateChangeKind.Remove)
+                // Check if the property was marked as remove in the cache or is expired
+                if (stateMetadata.ChangeKind == StateChangeKind.Remove || stateMetadata.TTLExpireTime <= DateTimeOffset.UtcNow)
```
philliphoff, line 148:
```suggestion
                if (stateMetadata.ChangeKind == StateChangeKind.Remove || (stateMetadata.TTLExpireTime.HasValue && stateMetadata.TTLExpireTime <= DateTimeOffset.UtcNow))
```

src/Dapr.Actors/Runtime/ActorStateManager.cs
```diff
@@ -114,10 +156,10 @@ public async Task<ConditionalValue<T>> TryGetStateAsync<T>(string stateName, Can
             var conditionalResult = await this.TryGetStateFromStateProviderAsync<T>(stateName, cancellationToken);
             if (conditionalResult.HasValue)
             {
-                stateChangeTracker.Add(stateName, StateMetadata.Create(conditionalResult.Value, StateChangeKind.None));
+                stateChangeTracker.Add(stateName, StateMetadata.Create(conditionalResult.Value.Value, StateChangeKind.None, ttlExpireTime: conditionalResult.Value.TTLExpireTime));
```
philliphoff, line 159:
```suggestion
                stateChangeTracker.Add(stateName, StateMetadata.Create(conditionalResult.Value.Value, StateChangeKind.None, ttlExpireTime: conditionalResult.Value.TTLExpireTime));

                return new ConditionalValue<T>(true, conditionalResult.Value.Value);
```

src/Dapr.Actors/Runtime/DaprStateProvider.cs
```diff
@@ -43,36 +44,36 @@ public DaprStateProvider(IDaprInteractor daprInteractor, JsonSerializerOptions j
...
-        public async Task<ConditionalValue<T>> TryLoadStateAsync<T>(string actorType, string actorId, string stateName, CancellationToken cancellationToken = default)
+        public async Task<ConditionalValue<ActorStateResponse<T>>> TryLoadStateAsync<T>(string actorType, string actorId, string stateName, CancellationToken cancellationToken = default)
         {
-            var result = new ConditionalValue<T>(false, default);
-            var stringResult = await this.daprInteractor.GetStateAsync(actorType, actorId, stateName, cancellationToken);
+            var result = new ConditionalValue<ActorStateResponse<T>>(false, default);
+            var response = await this.daprInteractor.GetStateAsync(actorType, actorId, stateName, cancellationToken);
 
-            if (stringResult.Length != 0)
+            if (response.Value.Length != 0 && response.TTLExpireTime > DateTimeOffset.UtcNow)
```
philliphoff, line 52:
```suggestion
            if (response.Value.Length != 0 && (!response.TTLExpireTime.HasValue || response.TTLExpireTime > DateTimeOffset.UtcNow))
```

philliphoff:
> @philliphoff Doesn't seem to be the case. Struggling to understand what is going wrong here.. at the limit of my .NET knowledge 😬

Poked around this morning and found a couple of issues where existing state was ignored in favor of default values (see above suggestions), which were causing the E2E to fail.  It might be good to add some unit tests for `DaprHttpInteractor`, `ActorStateManager`, and `DaprStateProvider`, if reasonably possible, to ensure that all the cases are covered (e.g. when values are cached vs. non-cached), pre- and post-expiration, etc.).

philliphoff (changes requested):
Needs some changes (added as suggestion comments) in order to resolve E2E failure.

JoshVanL:
Thank you @philliphoff, really appreciate your help with this!

I've gone ahead and added those fixes, expanded the E2E state tests and added unit tests for those classed. Please take another look :slightly_smiling_face:

JoshVanL:
Hi @philliphoff, please let me know what you think when you have time

philliphoff (approved):

halspang (approved):

Issue #1172, opened 2023-10-23
GetBulkStateAsync should have typed overload that performs bulk deserialization


WhitWaldo:
## Describe the feature
When calling `DaprClient.GetStateAsync<T>` one can specify a type parameter to have the SDK deserialize the response to. This is nice as it prevents the user from having to maintain their own serialization/deserialization options somewhere and rely on whatever the SDK is configured to use.

However, when using `DaprClient.GetBulkStateAsync`, this does not accept a type parameter, so one ultimately retrieves an `IReadOnlyList<BulkStateItem>` wherein the `BulkStateItem` contains ETag, Key and Value strings. This leaves deserialization to the user (who again, may have relied on Dapr to perform the initial serialization and thus may not be clued into precisely how it's serialized).

I would propose that for historical reasons, the existing approach be maintained, but that an overload be added to both `GetBulkStateAsync` and `BulkStateItem` that each accept a type parameter and which deserialize the returned result from the former as `IReadOnlyList<BulkStateItem<T>>` wherein the `BulkStateItem` contains a string property for both Key and ETag, but a type T value, wherein deserialization has already been performed, like that of `GetStateAsync`.

## Release Note
A type parameter overload has been added to `BulkStateItem` so items retrieved from state using `GetBulkStateItem<T>` will now have their values deserialize to T instead of exclusively returning a serailized string.

RELEASE NOTE:
**ADD** A type parameter overload has been added to `BulkStateItem` so the value of items retrieved from state using `GetBulkStateItem<T>` will now deserialize to T instead of exclusively returning a serialized string.

PR #1173, opened 2023-10-23
Add overload to deserialize GetBulkStateAsync item values

WhitWaldo:
Adds overload to BulkStateItem and GetBulkStateAsync to perform SDK-based deserialization of returned values instead of strictly returning serialized strings.

Signed-off-by: Whit Waldo <whit.waldo@innovian.net>

# Description
Explained in more detail at #1172 , but `GetBulkItemAsync` returns a `BulkStateItem` with a Key, ETag and Value wherein each is typed as a string and the user is expected to deserialize the Value. Because the state might have been originally added one by one, relying on Dapr and its serialization options to serialize everything, this can cause downstream issues deserializing values not originally serialized by the developer.

As such, just as `GetStateAsync` allows a type parameter to deserialize each returned value, I've added the same overload to `GetBulkItemAsync` to deserialize using the same mechanism as `GetStateAsync` to deserialize each of the returned items as part of an overloaded and generically-typed `BulkItemState`.

## Issue reference

As it's a simple addition, I didn't think it really merited a lot of discussion as the functionality is already available for single items (and this just adds the same for multiple items). 

Please reference the issue this PR will close: #1172

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests

I'll work on finding relevant documentation and updating that right now.


src/Dapr.Client/DaprClient.cs
```diff
@@ -714,6 +714,17 @@ public HttpRequestMessage CreateInvokeMethodRequest<TRequest>(string appId, stri
         /// <returns>A <see cref="Task{IReadOnlyList}" /> that will return the list of values when the operation has completed.</returns>
         public abstract Task<IReadOnlyList<BulkStateItem>> GetBulkStateAsync(string storeName, IReadOnlyList<string> keys, int? parallelism, IReadOnlyDictionary<string, string> metadata = default, CancellationToken cancellationToken = default);
 
+        /// <summary>
+        /// Gets a list of deserialized values associated with the <paramref name="keys" /> from the Dapr state store.
```
philliphoff, line 718: I might add some remarks to be clear about the assumption that all keys in the bulk operation are expected to be of the specified type (and maybe refer to the original method for cases where mixed types might be retrieved).

src/Dapr.Client/BulkStateItem.cs
```diff
@@ -49,4 +49,41 @@ public BulkStateItem(string key, string value, string etag)
...
+
+        /// <summary>
+        /// Gets the state key.
+        /// </summary>
+        public string Key { get; }
+
+        /// <summary>
+        /// Gets the value.
+        /// </summary>
+        public TValue Value { get; }
```
philliphoff, line 82: I might add remarks to the original `BulkStateItem.Value` docs about it being (possibly) serialized and might need deserialization.

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -640,6 +640,55 @@ public override async Task<IReadOnlyList<BulkStateItem>> GetBulkStateAsync(strin
             return bulkResponse;
         }
 
+        /// <inheritdoc/> 
+        public override async Task<IReadOnlyList<BulkStateItem<TValue>>> GetBulkStateAsync<TValue>(string storeName,
+            IReadOnlyList<string> keys, int? parallelism, IReadOnlyDictionary<string, string> metadata = default,
+            CancellationToken cancellationToken = default)
+        {
+            ArgumentVerifier.ThrowIfNullOrEmpty(storeName, nameof(storeName));
```
philliphoff, line 648: Would it make sense to consolidate the setup/execution of the bulk operation, as the only real difference is in handling of the response, in order to ensure these methods don't ever vary?

WhitWaldo: Are you ok if I rewrite both methods to use a private shared method? I took this approach to avoid deserializing the value from the JSON byte string twice and to avoid impacting the original method, but I agree - for forward maintainability, this could be improved. I'd propose a shared private method that returns the Key, ETag and ByteString values and leave it to either public method to then deserialize once in the manner most appropriate to either.

src/Dapr.Client/BulkStateItem.cs
WhitWaldo: Updated to provide a little more context here.

src/Dapr.Client/DaprClient.cs
WhitWaldo: Added more context to the comment per your suggestion.

src/Dapr.Client/DaprClientGrpc.cs
WhitWaldo: Updated to use the shared private method so each of the public-facing methods perform the deserialization against the ByteString only once.

philliphoff (approved):

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -597,16 +597,54 @@ public override async Task<TResponse> InvokeMethodGrpcAsync<TResponse>(string ap
...
+                bulkResponse.Add(new BulkStateItem(item.Key, item.Value.ToStringUtf8(), item.Etag));
+            }
+
+            return bulkResponse;
+        }
+        
+        /// <inheritdoc/> 
+        public override async Task<IReadOnlyList<BulkStateItem<TValue>>> GetBulkStateAsync<TValue>(string storeName,
+            IReadOnlyList<string> keys, int? parallelism, IReadOnlyDictionary<string, string> metadata = default,
+            CancellationToken cancellationToken = default)
```
halspang, line 617: Nit: If we've started to separate the parameters onto separate lines, our convention is to have each parameter on their own line.

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -597,16 +597,54 @@ public override async Task<TResponse> InvokeMethodGrpcAsync<TResponse>(string ap
...
+        }
+
+        /// <summary>
+        /// Retrieves the bulk state data, but rather than deserializing the values, leaves the specific handling
+        /// to the public callers of this method to avoid duplicate deserialization.
+        /// </summary>
+        private async Task<IReadOnlyList<(string Key, string Etag, ByteString Value)>> GetBulkStateRawAsync(
+            string storeName,
+            IReadOnlyList<string> keys, int? parallelism, IReadOnlyDictionary<string, string> metadata = default,
+            CancellationToken cancellationToken = default)
```
halspang, line 639: Same here.

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -597,16 +597,54 @@ public override async Task<TResponse> InvokeMethodGrpcAsync<TResponse>(string ap
...
         {
             ArgumentVerifier.ThrowIfNullOrEmpty(storeName, nameof(storeName));
             if (keys.Count == 0)
                 throw new ArgumentException("keys do not contain any elements");
 
             var envelope = new Autogenerated.GetBulkStateRequest()
             {
-                StoreName = storeName,
-                Parallelism = parallelism ?? default
+                StoreName = storeName, Parallelism = parallelism ?? default
```
halspang, line 647: Nit: These should be on their own line as before.

halspang (changes requested):
Looks good overall, just some formatting nitpicks.

src/Dapr.Client/DaprClientGrpc.cs
WhitWaldo: I've updated my editor settings so it honors this going forward. Corrections made in latest commit.

src/Dapr.Client/DaprClientGrpc.cs
WhitWaldo: Updated accordingly

src/Dapr.Client/DaprClientGrpc.cs
WhitWaldo: Updated accordingly

philliphoff:
@hhunter-ms @halspang If Hal is ok with the last set to changes he requested (and the required merge is clean), I say merge it.

halspang (approved):

Issue #1170, opened 2023-10-20
Increase Logging Metrics for Workflow


RyanLettieri:
## Describe the feature
Related to the following issue in runtime: https://github.com/dapr/dapr/issues/7065 this request is to increase the current logging that we have present for workflow.

## Release Note

RELEASE NOTE:

PR #1176, opened 2023-10-25
Initial implementation for workflow log tracing

RyanLettieri:
# Description

Added in basic logging for workflow to log the registered activities and registered workflows.

## Issue reference

Please reference the issue this PR will close: [ #_[1170]_](https://github.com/dapr/dotnet-sdk/issues/1170)

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/src/Dapr.Workflow/WorkflowLoggingService.cs
+++ b/src/Dapr.Workflow/WorkflowLoggingService.cs
@@ -39,17 +39,17 @@ namespace Dapr.Workflow
         {
             this.logger.Log(LogLevel.Information, "WorkflowLoggingService started");
 
-                this.logger.Log(LogLevel.Information, "List of registered workflows");
-                foreach (string item in registeredWorkflows)
-                {
-                    this.logger.Log(LogLevel.Information, item);
-                }
+            this.logger.Log(LogLevel.Information, "List of registered workflows");
+            foreach (string item in registeredWorkflows)
+            {
+                this.logger.Log(LogLevel.Information, item);
+            }
 
-                this.logger.Log(LogLevel.Information, "List of registered activities:");
-                foreach (string item in registeredActivities)
-                {
-                    this.logger.Log(LogLevel.Information, item);
-                }
+            this.logger.Log(LogLevel.Information, "List of registered activities:");
+            foreach (string item in registeredActivities)
+            {
+                this.logger.Log(LogLevel.Information, item);
+            }
 
             return Task.CompletedTask;
         }
--- a/src/Dapr.Workflow/WorkflowLoggingService.cs
+++ b/src/Dapr.Workflow/WorkflowLoggingService.cs
@@ -1,0 +1,88 @@
+// ------------------------------------------------------------------------
+// Copyright 2022 The Dapr Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//     http://www.apache.org/licenses/LICENSE-2.0
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// ------------------------------------------------------------------------
+
+namespace Dapr.Workflow
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.Extensions.Hosting;
+    using Microsoft.Extensions.Logging;
+
+    /// <summary>
+    /// Defines runtime options for workflows.
+    /// </summary>
+    internal sealed class WorkflowLoggingService : IHostedService
+    {
+        private readonly ILogger<WorkflowLoggingService> logger;
+        private static HashSet<string>? registeredWorkflows;
+        private static HashSet<string>? registeredActivities;
+        private LogLevel logLevel = LogLevel.Debug;
+
+        public WorkflowLoggingService(ILogger<WorkflowLoggingService> logger)
+        {
+            var value = Environment.GetEnvironmentVariable("DAPR_LOG_LEVEL");
+            logLevel = string.IsNullOrEmpty(value) ? LogLevel.Debug : (LogLevel)Enum.Parse(typeof(LogLevel), value);
+            this.logger = logger;
+            registeredActivities = new HashSet<string>();
+            registeredWorkflows = new HashSet<string>();
+        }
+        public Task StartAsync(CancellationToken cancellationToken)
+        {
+            this.logger.Log(logLevel, "WorkflowLoggingService started");
+    
+            if (registeredWorkflows != null)
+            {
+                this.logger.Log(logLevel, "List of registered workflows");
+                foreach (string item in registeredWorkflows)
+                {
+                    this.logger.Log(logLevel, item);
+                }
+            }
+
+            if (registeredActivities != null)
+            {
+                this.logger.Log(logLevel, "List of registered activities:");
+                foreach (string item in registeredActivities)
+                {
+                    this.logger.Log(logLevel, item);
+                }
+            }
+            return Task.CompletedTask;
+        }
+
+        public Task StopAsync(CancellationToken cancellationToken)
+        {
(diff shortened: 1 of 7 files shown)


src/Dapr.Workflow/WorkflowLoggingService.cs
```diff
@@ -0,0 +1,88 @@
...
+    using Microsoft.Extensions.Hosting;
+    using Microsoft.Extensions.Logging;
+
+    /// <summary>
+    /// Defines runtime options for workflows.
+    /// </summary>
+    internal sealed class WorkflowLoggingService : IHostedService
+    {
+        private readonly ILogger<WorkflowLoggingService> logger;
+        private static HashSet<string>? registeredWorkflows;
```
philliphoff, line 29: Suggest: `private static readonly HashSet<string> registeredWorkflows = new();`

There's no reason *not* to just create an (and ensure only one) instance.  Then you needn't check for `null` later on.  (Waiting until the service instance is created to create the `static` collections actually creates a race condition.)

src/Dapr.Workflow/WorkflowLoggingService.cs
```diff
@@ -0,0 +1,88 @@
...
+        private static HashSet<string>? registeredActivities;
+        private LogLevel logLevel = LogLevel.Debug;
+
+        public WorkflowLoggingService(ILogger<WorkflowLoggingService> logger)
+        {
+            var value = Environment.GetEnvironmentVariable("DAPR_LOG_LEVEL");
```
philliphoff, line 35: I'd suggest pulling such settings from the `IConfiguration` service which, by default in ASP.NET, aggregates environment variables as well as a number of other sources of configuration values (e.g. JSON files, cloud settings providers, etc.).

src/Dapr.Workflow/WorkflowLoggingService.cs
```diff
@@ -0,0 +1,88 @@
...
+            logLevel = string.IsNullOrEmpty(value) ? LogLevel.Debug : (LogLevel)Enum.Parse(typeof(LogLevel), value);
```
philliphoff, line 36: Rather than making the logging level of its entires configurable, I'd suggest using a level that makes the most sense for the information being displayed (information?), and then rely on MS logging's mechanisms for configuring which logs are written, which would seem more in line with ASP.NET logging.

src/Dapr.Workflow/WorkflowLoggingService.cs
philliphoff, line 35: I'd also note that the `LogLevel` enumeration's values don't perfectly align with the Dapr runtime's levels, which could be confusing in a variable somewhat generically named `DAPR_LOG_LEVEL`.  (I don't think the runtime uses such a variable at the moment, but if needed it'd probably be better to make it more .NET SDK and workflow specific.)

src/Dapr.Workflow/WorkflowLoggingService.cs
```diff
@@ -0,0 +1,86 @@
...
+    internal sealed class WorkflowLoggingService : IHostedService
+    {
+        private readonly ILogger<WorkflowLoggingService> logger;
+        private static readonly HashSet<string> registeredWorkflows = new();
+        private static readonly HashSet<string> registeredActivities = new();
+        private LogLevel logLevel = LogLevel.Information;
+
+        public WorkflowLoggingService(ILogger<WorkflowLoggingService> logger, IConfiguration configuration)
+        {
+            logLevel = string.IsNullOrEmpty(configuration["DAPR_LOG_LEVEL"]) ? LogLevel.Information : ConvertLogLevel(configuration["DAPR_LOG_LEVEL"]);
```
philliphoff, line 36: With the changes to make log entries `Information`, is any of this needed any longer?

test/Dapr.E2E.Test.App/Program.cs
```diff
@@ -1,23 +1,25 @@
...
 namespace Dapr.E2E.Test
 {
     using Microsoft.AspNetCore.Hosting;
     using Microsoft.Extensions.Hosting;
+    using Serilog;
     public class Program
     {
         public static void Main(string[] args)
         {
+            Log.Logger = new LoggerConfiguration().WriteTo.File("log.txt").CreateLogger();
```
philliphoff, line 22: Should this be *both* here and in `Startup.ConfigureServices()`?  (I would think just the latter.)

test/Dapr.E2E.Test/Workflows/WorkflowTest.cs
```diff
@@ -11,25 +11,76 @@
...
+                            }
+                        }
+                        allLogsFound = logStrings.All(k => k.Value);
+                        if (allLogsFound)
+                        {
+                            break;
+                        }
+                    }
+                }
+            }, cts.Token);
```
philliphoff, line 68: AFAIK, this `CancellationToken` is only used if the `Task` has not yet already started, which it most likely has, so it's possible that the task never ends (in the case of not all expected logs being found).

You could use the `async` method `Stream.ReadLineAsync(CancellationToken)` to provide a means of abandoning the search when time expires.  That would also allow you to avoid calling `Task.WaitAll()`, which is a synchronously-blocking method.

test/Dapr.E2E.Test/Workflows/WorkflowTest.cs
```diff
@@ -11,25 +11,76 @@
...
+
+            if (!Task.WaitAll(new Task[] {searchTask}, timeout))
```
philliphoff, line 70: Nit: maybe hold off testing the result until after the log file has been deleted (rather than two deletion paths).

src/Dapr.Workflow/WorkflowLoggingService.cs
```diff
@@ -0,0 +1,75 @@
...
+                foreach (string item in registeredWorkflows)
+                {
+                    this.logger.Log(LogLevel.Information, item);
+                }
+
+                this.logger.Log(LogLevel.Information, "List of registered activities:");
+                foreach (string item in registeredActivities)
+                {
+                    this.logger.Log(LogLevel.Information, item);
+                }
```
halspang, line 52: Nit: Indentation is weird here.

src/Dapr.Workflow/WorkflowLoggingService.cs
```diff
@@ -0,0 +1,75 @@
...
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.Extensions.Hosting;
+    using Microsoft.Extensions.Logging;
+    using Microsoft.Extensions.Configuration;
+
+    /// <summary>
+    /// Defines runtime options for workflows.
+    /// </summary>
+    internal sealed class WorkflowLoggingService : IHostedService
```
halspang, line 27: I'm not sure if hosted service is the right way to go here. I was thinking more that this would be injected into the various workflow runtime mechanics, like the context.

halspang (commented):
Can you give some examples of what logs are output when we run a workflow?

test/Dapr.E2E.Test/Workflows/WorkflowTest.cs
```diff
@@ -11,25 +11,81 @@
...
+            using var daprClient = new DaprClientBuilder().UseGrpcEndpoint(this.GrpcEndpoint).UseHttpEndpoint(this.HttpEndpoint).Build();
+            var health = await daprClient.CheckHealthAsync();
+            health.Should().Be(true, "DaprClient is not healthy");
+
+            var searchTask = Task.Run(async() =>
+            {
+                using (StreamReader reader = new StreamReader(logFilePath))
+                {
+                    string line;
+                    while ((line = await reader.ReadLineAsync()) != null)
```
philliphoff, line 51: Need to pass the cancellation token to `ReadLineAsync()` so it will stop reading after the timeout.

test/Dapr.E2E.Test/Workflows/WorkflowTest.cs
```diff
@@ -11,25 +11,81 @@
...
+            }, cts.Token);
+
+            try
+            {
+                await searchTask;
+            }
+            finally
+            {
+                File.Delete(logFilePath);
+                if (!allLogsFound)
```
philliphoff, line 76: The `allLogsFound` test should probably be outside/after the `finally` block; it doesn't need to be executed in the case of an exception (e.g. timeout) as the exception will bubble up and fail the test.

philliphoff (approved):

halspang (approved):

PR #1185, opened 2023-11-13
Updates Dapr to 1.12 in GitHub actions itest

JoshVanL:


.github/workflows/itests.yml
```diff
@@ -37,11 +37,11 @@ jobs:
...
       GOPROXY: https://proxy.golang.org
-      DAPR_CLI_VER: 1.9.1
-      DAPR_RUNTIME_VER: 1.10.5
-      DAPR_INSTALL_URL: https://raw.githubusercontent.com/dapr/cli/3dacfb672d55f1436c249057aaebbe597e1066f3/install/install.sh
+      DAPR_CLI_VER: 1.12.0
+      DAPR_RUNTIME_VER: 1.12.0
+      DAPR_INSTALL_URL: https://raw.githubusercontent.com/dapr/cli/release-1.12/install/install.sh
       DAPR_CLI_REF: ''
-      DAPR_REF: '4181de0edc65fc98a836ae7abc6042c575c8fae5'
+      DAPR_REF: '2149fca96cdf11627c387bda26dcc027d1c47354'
```
halspang, line 44: The Dapr ref set's a specific commit and is redundant to the 1.12 flag. I'd prefer we just set 1.12 and leave that blank.

halspang (changes requested):
Let's remove the commit ref as that's for updating the dapr version for a new feature (mid release).

halspang:
Yeah, that sure does. @RyanLettieri, can you take a look at this workflow failure?

RyanLettieri:
@JoshVanL The error looks like it is coming from an assert statement after we purge an instance and attempt to get the workflow instance back. The current statement we use to check for this is the following:
`ex.InnerException.Message.Should().Contain("No such instance exists", $"Instance {instanceId} was not correctly purged");`

Due to Should().Contain() being case-sensitive, the error is probably coming from a change in the capitalization of the error message. From the error logs:

> Error: Dapr.E2E.Test.E2ETests.TestWorkflows: Expected string "Status(StatusCode="NotFound", Detail="unable to find workflow with the provided instance ID: testInstanceId%!(EXTRA *fmt.wrapError=Failed to fetch orchestration metadata: no such instance exists)")" to contain "No such instance exists" because Instance testInstanceId was not correctly purged.

It appears we check for a capital N on the error whereas the error in 1.12 seems to use a lower-case N. I'd recommend changing that line and re-running the test

JoshVanL:
Thanks @RyanLettieri! Really appreciate the detailed breakdown- I'm still getting used to dotnet so thanks for the help.

halspang (approved):

Issue #679, opened 2021-05-16
Can't use C# 9.0 Record for actor method return type


LukePammant:
## Expected Behavior

I would like to use a record as a return type for an actor method.

## Actual Behavior

> ActorInvokeException: Type 'Chronos.Actors.EquipmentAssignments' cannot be serialized. Consider marking it with the DataContractAttribute attribute, and marking all of its members you want serialized with the DataMemberAttribute attribute. Alternatively, you can ensure that the type is public and has a parameterless constructor - all public members of the type will then be serialized, and no attributes will be required.

Here is the record definition:

``` C#
    public record EquipmentActorState(Guid Id, Guid SolarSystemId, string OwnerTypeId, EquipmentAssignments EquipmentAssignments, List<EquipmentSwapRequest> SwapRequests);
    public record EquipmentAssignments(Guid Id, string ShipTypeId, Dictionary<int, EquipmentSlot> EquipmentSlots, Dictionary<Guid, EquipmentReference> AvailableEquipment);
    public record EquipmentSlot(EquipmentSlotType Type, Guid? CurrentlyAssignedEquipment);
    public record EquipmentReference(Guid Id, string Type, EquipmentSlotType SlotType, int? AssignedSlotId);

//... Inside the actor class definition
    public async Task<EquipmentAssignments?> GetEquipmentDetails()
    {
         var state = await StateManager.GetStateAsync<EquipmentActorState>("state");
         return state.EquipmentAssignments; // <-- this cannot be serialized?
    }
```

It fails on this line in the dotnet-sdk:
https://github.com/dapr/dotnet-sdk/blob/master/src/Dapr.Actors/Communication/ActorMessageBodyDataContractSerializationProvider.cs#L206

Which throws this exception:
```
Exception has occurred: CLR/System.Runtime.Serialization.InvalidDataContractException
Exception thrown: 'System.Runtime.Serialization.InvalidDataContractException' in System.Private.DataContractSerialization.dll: 'Type 'Chronos.Actors.EquipmentAssignments' cannot be serialized. Consider marking it with the DataContractAttribute attribute, and marking all of its members you want serialized with the DataMemberAttribute attribute. Alternatively, you can ensure that the type is public and has a parameterless constructor - all public members of the type will then be serialized, and no attributes will be required.'
   at System.Runtime.Serialization.DataContract.DataContractCriticalHelper.ThrowInvalidDataContractException(String message, Type type)
[...]

PR #1199, opened 2023-12-01
Docs: Actor serialization documentation

WhitWaldo:
# Description
This is my first draft of a new page in the documentation detailing how serialization works using the Dapr .NET Actors SDK today. It's my response to the issue raised [here](https://github.com/dapr/dotnet-sdk/issues/679) which is marked as a bug, but reflects a misunderstanding of how the serialization engine handles records.  My goal with adding this page is to provide an upfront resource for developers using actors to understand why serialization may not be working as expected out of the box and provide not only a solution, but an explanation.

I detail how serialization is performed to illustrate the benefits to either going the parameterless constructor route or decorating types and members with the applicable attributes and then share my recommendation to use attributes and why. I dive through several examples documenting expectations in classes, structs and records and then speak a bit to other related scenarios (collections, enums, polymorphic types) towards the end pointing back to the official documentation to keep this document from getting _too_ verbose.

This is a documentation update only and changes no code in the SDK.

## Issue reference

Please reference the issue this PR will close: #679 

## Checklist

~* [ ] Code compiles correctly~
~* [ ] Created/updated tests~

### DIFF
--- a/daprdocs/content/en/dotnet-sdk-docs/dotnet-actors/dotnet-actors-serialization.md
+++ b/daprdocs/content/en/dotnet-sdk-docs/dotnet-actors/dotnet-actors-serialization.md
@@ -1,0 +1,315 @@
+---
+type: docs
+title: "Actor serialization in the .NET SDK"
+linkTitle: "Actor serialization"
+weight: 300000
+description: Necessary steps to serialize your types using remoted Actors in .NET
+---
+
+The Dapr actor package enables you to use Dapr virtual actors within a .NET application with strongly-typed remoting, but if you intend to send and receive strongly-typed data from your methods, there are a few key ground rules to understand. In this guide, you will learn how to configure your classes and records so they are properly serialized and deserialized at runtime.
+
+# Data Contract Serialization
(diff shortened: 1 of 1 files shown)


daprdocs/content/en/dotnet-sdk-docs/dotnet-actors/dotnet-actors-serialization.md
```diff
@@ -0,0 +1,315 @@
...
+
+    public Guid Id { get; set; }
+    public string Name { get; set; }
+    public int Count { get; set; }
+}
+```
+
+In the above example, we don't need to also use the [DataMemberAttribute](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.serialization.datamemberattribute) attributes because again, we're using [built-in primitives](###supported-primitive-types) that the serializer supports. But, we do get more flexibility if we use the attributes. From the DataContractAttribute attribute, we can specify our own XML namespace with the Namespace argument and, via the Name argument, change the name of the type as used when serialized into the XML document. 
+
+It's a recommended practive to append the DataContractAttribute attribute to the type and the DataMemberAttribute attributes to all the members you want to serialize anyway - if they're not necessary and you're not changing the default values, they'll just be ignored, but they give you a mechanism to opt into serializing members that wouldn't otherwise have been included such as those marked as private or that are themselves complex types or collections.
```
philliphoff, line 119: Spelling: `practice`

WhitWaldo: Spell check is an excellent idea. Fixed a few additional misspellings as well.

WhitWaldo:
> Overall a fantastic summary of `DataContract` based serialization. The only thought I had was whether to further emphasize the official .NET docs at the beginning as "the final word" in serialization, as much of this is not Dapr-specific and subject to change with the .NET Framework rather than Dapr.

I think that's an excellent addition. I've called that out separately and linked to the implementation in the .NET Github repo.

halspang:
@WhitWaldo - Thanks for your contribution! Can you sign-off on your commit so DCO will pass?

WhitWaldo:
@halspang As far as I can tell, both commits include the sign-off.

Issue #1203, opened 2023-12-07
Add more workflow samples


MregXN:
## Describe the proposal
Python and Java SDKs have provided sample codes based on the [Dapr workflow pattern](https://docs.dapr.io/developing-applications/building-blocks/workflow/workflow-patterns/) as mentioned in the official documents. However, dotnet SDK is yet to provide corresponding samples.

It is suggested to first add the following sample according to the documentation:
- [Task chaining](https://docs.dapr.io/developing-applications/building-blocks/workflow/workflow-patterns/#task-chaining)
- [Fan-out/fan-in](https://docs.dapr.io/developing-applications/building-blocks/workflow/workflow-patterns/#fan-outfan-in)
- [Monitor](https://docs.dapr.io/developing-applications/building-blocks/workflow/workflow-patterns/#monitor)
- [External system interaction](https://docs.dapr.io/developing-applications/building-blocks/workflow/workflow-patterns/#external-system-interaction)

cgillum:
Looks like the work has been mostly completed and just needs to be reviewed/merged.

WhitWaldo:
@MregXN I'm showing a merge conflict in the solution file for each of the referenced issues in this PR. If you can resolve that, I'd be happy to review these and get them merged in.

PR #1204, opened 2023-12-08
Add workflow sample fanout/fanin

MregXN:
# Description

Please reference the issue #1203


WhitWaldo:
@MregXN First of all, thank you very much for your contributions!

I'd like to merge this and your other PRs, but it looks as though they're using the since-deprecated methods on the `DaprClient` instead of those from the separate `Dapr.Workflow` client.

Each client had the same general layout, so I created a matching review on each and suggested a few typo corrections to the README. If you're able to get these changes made, I don't see any other blockers to merging each of these into 1.15 barring a brief review following your updates.

PR #1188, opened 2023-11-14
.NET 8 Support

philliphoff:
# Description

Add explicit .NET 8 targets to Dapr .NET SDK NuGet packages.

Currently, the Dapr .NET SDK NuGet packages target only .NET 6 (except for Workflow, which also targets .NET 7).  With this change, the packages will additionally target .NET 8.

## Issue reference

We strive to have all PR being opened based on an issue, where the problem or feature have been discussed prior to implementation.

Please reference the issue this PR will close: #1187 

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [x] Code compiles correctly
* [x] Created/updated tests
* [ ] Extended the documentation

PR #1217, opened 2023-12-22
Implementing Cryptography building block in .NET

### LINKED ISSUE #1072: [Crypto] Implement support for EncryptAlpha1/DecryptAlpha1 APIs
dapr/dapr#6197 implements two new APIs in Dapr:

- `EncryptAlpha1`
- `DecryptAlpha1`

Both APIs are stream-based, so the SDK is expected to read input as a stream, send it to Dapr using a streamed gRPC in chunks, and in parallel read the response from Dapr (in chunks sent over the same gRPC stream) and return a readable stream to the caller.

The reference implementation is in the Go SDK: dapr/go-sdk#387

WhitWaldo:
# Description
This adds support for Cryptography building block (alpha) in .NET supporting both the encrypt and decrypt streaming APIs.

Examples work and I've updated the README to reflect use with Azure Key Vault (RBAC, environment variable install, etc.).

## Issue reference

This closes #1072 - per recommendations in Discord, this streams the requests in configurable blocks (defaulting to 4 KB), sending only the request metadata payload in the first message and sending the plaintext/ciphertext blocks in subsequent messages. I manually updated the relevant sections of the dapr.proto pulling from the latest in dapr/dapr.

Submitted update for docs at [https://github.com/dapr/docs/pull/3928](https://github.com/dapr/docs/pull/3928).
Also needs a Quickstart to show how this is used https://github.com/dapr/quickstarts/issues/969

I still need to figure out how to write any tests for this, but I confirmed that my example (properly configured) runs without issue locally.

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests (some tests written - presumably room for more)


src/Dapr.Client/EnumExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        /// <summary>
+        /// Reads the value of an enum out of the attached <see cref="EnumMemberAttribute"/> attribute.
+        /// </summary>
+        /// <typeparam name="T">The enum.</typeparam>
+        /// <param name="value">The value of the enum to pull the value for.</param>
+        /// <returns></returns>
+        public static string GetValueFromEnumMember<T>(this T value) where T : Enum
+        {
+            var memberInfo = typeof(T).GetMember(value.ToString());
```
philliphoff, line 29: Consider constraining the types of members returned to be *only* the static/public/declared fields of that enumerated type (e.g. not base types, such as `Object`).

src/Dapr.Client/DaprClient.cs
```diff
@@ -939,6 +940,293 @@ public async Task<StateEntry<TValue>> GetStateEntryAsync<TValue>(string storeNam
...
+        /// </summary>
+        /// <param name="vaultResourceName">The name of the vault resource used by the operation.</param>
+        /// <param name="plainTextBytes">The bytes of the plaintext value to encrypt.</param>
+        /// <param name="algorithm">The name of the algorithm used to wrap the encryption key.</param>
+        /// <param name="keyName">The name of the key to use from the Vault for the encryption operation.</param>
+        /// <param name="dataEncryptionCipher">The name of the cipher to use for the encryption operation.</param>
+        /// <param name="cancellationToken">A <see cref="CancellationToken"/> that can be used to cancel the operation.</param>
+        /// <returns>An array of encrypted bytes.</returns>
+        [Obsolete("The API is currently not stable as it is in the Alpha stage. This attribute will be removed once it is stable.")]
+        public abstract Task<byte[]> EncryptAsync(string vaultResourceName, byte[] plainTextBytes,
```
philliphoff, line 956: The number of arguments for these methods is rather high, which can make learning/use intimidating and error prone (i.e. inconsistencies from one call to the next).  I wonder if there'd be any benefit to moving some of the arguments to an "options" object to make the methods easier to digest as well as to promote the aggregation of related values (e.g. the algorithm, cipher, etc.) rather than being handed one-by-one to each encrypt call.

src/Dapr.Client/DaprClient.cs
```diff
@@ -939,6 +940,293 @@ public async Task<StateEntry<TValue>> GetStateEntryAsync<TValue>(string storeNam
...
+        /// <param name="plainTextBytes">The bytes of the plaintext value to encrypt.</param>
+        /// <param name="algorithm">The name of the algorithm used to wrap the encryption key.</param>
+        /// <param name="keyName">The name of the key to use from the Vault for the encryption operation.</param>
+        /// <param name="decryptionKeyName">The name (and optionally version) of the decryption key to specify should be used.</param>
+        /// <param name="dataEncryptionCipher">The name of the cipher to use for the encryption operation.</param>
+        /// <param name="cancellationToken">A <see cref="CancellationToken"/> that can be used to cancel the operation.</param>
+        /// <returns>An array of encrypted bytes.</returns>
+        [Obsolete("The API is currently not stable as it is in the Alpha stage. This attribute will be removed once it is stable.")]
+        public abstract Task<byte[]> EncryptAsync(string vaultResourceName, byte[] plainTextBytes,
+            KeyWrapAlgorithm algorithm, string keyName, string decryptionKeyName, DataEncryptionCipher dataEncryptionCipher = DataEncryptionCipher.AesGcm,
```
philliphoff, line 973: Should there be an explicit endeavour to make the default cipher's consistent across SDKs?  (Is this consistent with other SDKs today?)

src/Dapr.Client/DaprClient.cs
philliphoff, line 956: Consider accepting `ReadOnlyMemory<byte>` instead of `byte[]` (as the latter should be implicitly convertible to the former)?

philliphoff, line 956: Is "plaintext" intended in the "common term for arbitrary bytes to be encrypted" sense or is there an expectation that the value actually represent text?  If the former, perhaps a different argument name (e.g. `source` or `sourceBytes` or even `bytesToBeEncrypted`)?

src/Dapr.Client/DaprClient.cs
```diff
@@ -939,6 +940,293 @@ public async Task<StateEntry<TValue>> GetStateEntryAsync<TValue>(string storeNam
...
+        /// </summary>
+        /// <param name="vaultResourceName">The name of the vault resource used by the operation.</param>
+        /// <param name="plainTextStream">The stream containing the bytes of the plaintext value to encrypt.</param>
+        /// <param name="algorithm">The name of the algorithm used to wrap the encryption key.</param>
+        /// <param name="keyName">The name of the key to use from the Vault for the encryption operation.</param>
+        /// <param name="dataEncryptionCipher">The name of the cipher to use for the encryption operation.</param>
+        /// <param name="cancellationToken">A <see cref="CancellationToken"/> that can be used to cancel the operation.</param>
+        /// <returns>An array of encrypted bytes.</returns>
+        [Obsolete("The API is currently not stable as it is in the Alpha stage. This attribute will be removed once it is stable.")]
+        public abstract Task<byte[]> EncryptAsync(string vaultResourceName, Stream plainTextStream,
```
philliphoff, line 987: I wonder if it'd make sense to offer a streamed response (e.g. based on `IAsyncEnumerable`) if provided an input stream?

src/Dapr.Client/DaprClient.cs
```diff
@@ -939,6 +940,293 @@ public async Task<StateEntry<TValue>> GetStateEntryAsync<TValue>(string storeNam
...
+        /// Encrypts an array of bytes using the Dapr Cryptography encryption functionality.
+        /// </summary>
+        /// <param name="vaultResourceName">The name of the vault resource used by the operation.</param>
+        /// <param name="plainTextBytes">The bytes of the plaintext value to encrypt.</param>
+        /// <param name="algorithm">The name of the algorithm used to wrap the encryption key.</param>
+        /// <param name="keyName">The name of the key to use from the Vault for the encryption operation.</param>
+        /// <param name="dataEncryptionCipher">The name of the cipher to use for the encryption operation.</param>
+        /// <param name="cancellationToken">A <see cref="CancellationToken"/> that can be used to cancel the operation.</param>
+        /// <returns>An array of encrypted bytes.</returns>
+        [Obsolete("The API is currently not stable as it is in the Alpha stage. This attribute will be removed once it is stable.")]
```
philliphoff, line 955: Not necessarily for this PR, but I wonder if we should be using the newer .NET 6 (and possibly more appropriate)[`RequiresPreviewFeaturesAttribute`](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.versioning.requirespreviewfeaturesattribute?view=net-6.0) instead of `ObsoleteAttribute` for our alpha APIs.

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -1373,6 +1375,537 @@ private Autogenerated.StateItem ToAutogeneratedStateItem(StateTransactionRequest
...
+
+        #region Cryptography
+
+        private async Task<byte[]> EncryptStreamAsync(string vaultResourceName, Stream stream,
+            KeyWrapAlgorithm algorithm, string keyName, DataEncryptionCipher dataEncryptionCipher,
+            string decryptionKeyName, bool omitDecryptionKey, CancellationToken cancellationToken = default)
+        {
+            ArgumentVerifier.ThrowIfNullOrEmpty(vaultResourceName, nameof(vaultResourceName));
+            ArgumentVerifier.ThrowIfNullOrEmpty(keyName, nameof(keyName));
+
```
philliphoff, line 1389: Nit: verify a non-null stream?

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -1373,6 +1375,537 @@ private Autogenerated.StateItem ToAutogeneratedStateItem(StateTransactionRequest
...
+            var options = CreateCallOptions(headers: null, cancellationToken);
+            var duplexStream = client.EncryptAlpha1(options);
+
+            //Start with passing the metadata about the encryption request itself in the first message
+            await duplexStream.RequestStream.WriteAsync(
+                new Autogenerated.EncryptRequest { Options = encryptRequestOptions }, cancellationToken);
+
+            //Send the plaintext bytes in 4 kb blocks in subsequent messages
+            const int blockSize = 4 * 1024; // 4 kb
```
philliphoff, line 1414: Should the block size be an option to the encrypt method?

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -1373,6 +1375,537 @@ private Autogenerated.StateItem ToAutogeneratedStateItem(StateTransactionRequest
...
+                    //Increment the sequence number
+                    sequenceNumber++;
+                }
+            }
+            
+            //Send the completion message
+            await duplexStream.RequestStream.CompleteAsync();
+
+            //Set up how the response is handled since we know it's of finite length
+            var returnedBytes = new List<byte>();
```
philliphoff, line 1440: Is this significantly different than writing to a `MemoryStream` and then returning its underlying buffer?

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -1373,6 +1375,537 @@ private Autogenerated.StateItem ToAutogeneratedStateItem(StateTransactionRequest
...
+        [Obsolete("The API is currently not stable as it is in the Alpha stage. This attribute will be removed once it is stable.")]
+        public override Task<byte[]> EncryptAsync(
+            string vaultResourceName, Stream plainTextStream, KeyWrapAlgorithm algorithm,
+            string keyName, string decryptionKeyName,
+            DataEncryptionCipher dataEncryptionCipher = DataEncryptionCipher.AesGcm,
+            CancellationToken cancellationToken = default) => EncryptStreamAsync(vaultResourceName, plainTextStream,
+            algorithm, keyName, dataEncryptionCipher,
+            decryptionKeyName, false, cancellationToken);
+
+        private async Task<byte[]> EncryptByteArrayAsync(string vaultResourceName, byte[] plainTextBytes,
```
philliphoff, line 1467: Could we not just wrap `plainTextBytes` in a `MemoryStream` and then use a single stream-based encrypt implementation?

examples/Client/Cryptography/Cryptography.csproj
```diff
@@ -0,0 +1,21 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <OutputType>Exe</OutputType>
+        <TargetFramework>net8.0</TargetFramework>
```
philliphoff, line 5: I don't know if there's any official guidance, but it looks like most of our examples target `net6`.

src/Dapr.Client/DaprClient.cs
WhitWaldo: This sort of fits into the next review note as well, but one of the goals stated for the reduced crypto interface footprint was that they wanted to simplify the approach as the "subtle" approach is too complex for the average person simply wanting to encrypt/decrypt something. What I was trying to do then was make it easy for anyone to pass in the barest number of arguments necessary to achieve that goal while also expose the few additional options available in the existing API for those mildly more advanced users.

The downside to the options approach is it simply shifts the complexity from the method arguments to options constructor overloads so that all the appropriately paired combinations are properly set. Yes, it'd limit the methods to two (one with options, one assuming some reasonable defaults), but it doesn't necessarily make using the options less intimidating.

That said, especially if there's interest in expanding the crypto building block to do more than the existing encrypt/decrypt (e.g. more of the "subtle" methods like wrapping keys or signing and verifying), I think separate options for each method would make a lot of sense to keep from ballooning the method space too broadly.

src/Dapr.Client/DaprClient.cs
WhitWaldo: This is the default per the comments in the [dapr.proto](https://github.com/dapr/dapr/blob/b99cf8b5476a503f655d74ce478e681a1f9a37f0/dapr/proto/runtime/v1/dapr.proto#L976) so I simply selected it, but in case there was a reason for users to use the alternative (right now just "chacha20-poly1305"), I wanted to leave users the option to select the other one. That said, [this issue](https://github.com/dapr/dapr/issues/6027) suggested that a third should be added and used as the default (perhaps completely replacing the alternatives) and by setting it as the default it would promote better security.

src/Dapr.Client/DaprClient.cs
WhitWaldo: I think I added the byte[] overload after the string variant, so it was just a renaming of the existing parameter ("plaintext") to make it match the type. That said, yes, "ciphertext" is a commonly used term in cryptography to describe the encrypted version of data and "plaintext" is the original or since-decrypted version of that data. The two are the common opposite of one another, so it would make sense to refer to this just as "plaintext" or "plaintextBytes" (the T shouldn't have been capitalized) with a clarifying description in the XML to make it clear to the developer that these are the bytes to be encrypted.

src/Dapr.Client/DaprClientGrpc.cs
WhitWaldo: Good catch - added.

src/Dapr.Client/DaprClient.cs
WhitWaldo: I would handily agree, but I wanted to opt for consistency with how the methods were marked for the distributed lock building block, so I just copied the approach from there.

src/Dapr.Client/DaprClient.cs
WhitWaldo: I mentioned this in one of the [comments](https://github.com/dapr/dotnet-sdk/blob/0e92eae547e5800e5501732df65f6a525a99e1d9/src/Dapr.Client/DaprClientGrpc.cs#L1519), but the only reason I didn't add one here is because of the [conversation](https://discord.com/channels/778680217417809931/1133104673558114439/1187858033091563591) I had with @ItalyPaleAle in the Crypto channel on Discord in which they said that "gRPC doesn't allow continuous streams of data" which clarified that the data being encrypted is bounded. As such, as part of the implementation in DaprClientGrpc, I read this back out to the byte array and return it.

The idea was that by only exposing methods with a bounded length, we don't encourage developers to build something that would enable unbounded streams since those aren't actually supported at the sidecar level (though it's not obvious from the .NET SDK itself). However, if you think that's something better addressed through documentation, I'm not otherwise opposed (especially since I don't know the background of this limitation - if it's something being worked on, perhaps this overload should be added when that work is completed).

src/Dapr.Client/DaprClientGrpc.cs
WhitWaldo: This is another relic of the [discussion](https://discord.com/channels/778680217417809931/1133104673558114439/1187856221529722910) on Discord in which the recommendation was to use 2-4 KB chunks with no more than 64KB. I'm entirely in favor of using this as a default and letting the developer pick a specific value, but we should resolve first whether we'd prefer to use method overloads to do that or switch to an options block (and its various constructor overloads) instead.

src/Dapr.Client/DaprClientGrpc.cs
WhitWaldo: Why, yes.. Yes we could. Added in the latest commit.

examples/Client/Cryptography/Cryptography.csproj
WhitWaldo: I switched to .NET 8 after seeing your [PR for the version bump](https://github.com/dapr/dotnet-sdk/pull/1188) get merged for the SDK itself. Seems if the SDK is using the latest, the examples probably should too, but that can be another PR.

src/Dapr.Client/EnumExtensions.cs
WhitWaldo: Done

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -1373,6 +1375,475 @@ private Autogenerated.StateItem ToAutogeneratedStateItem(StateTransactionRequest
             var resp = await client.UnsubscribeConfigurationAsync(request, options);
             return new UnsubscribeConfigurationResponse(resp.Ok, resp.Message);
         }
+
+        #endregion
+
+        #region Cryptography
+
+        private async Task<byte[]> EncryptStreamAsync(string vaultResourceName, Stream stream,
```
ItalyPaleAle, line 1383: Ideally in this method you would want to return a stream too, or the response is still buffered in-memory.

At least there should be a variant of this method that returns a stream so it can be written to file (or somewhere else) without buffering

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -1373,6 +1375,475 @@ private Autogenerated.StateItem ToAutogeneratedStateItem(StateTransactionRequest
...
+            var options = CreateCallOptions(headers: null, cancellationToken);
+            var duplexStream = client.EncryptAlpha1(options);
+
+            //Start with passing the metadata about the encryption request itself in the first message
+            await duplexStream.RequestStream.WriteAsync(
+                new Autogenerated.EncryptRequest { Options = encryptRequestOptions }, cancellationToken);
+
+            //Send the plaintext bytes in 4 kb blocks in subsequent messages
+            const int blockSize = 4 * 1024; // 4 kb
+            await using (var bufferedStream = new BufferedStream(stream, blockSize))
```
ItalyPaleAle, line 1415: Maybe this is because my knowledge of C# is limited, but if I understand correctly here you first send the entire data to Dapr in chunks, and then (when the data has been fully sent), you start reading the response.

Ideally, you'd be reading the response _at the same time_ as you're sending the data to Dapr.

The idea is that you can send a chunk to Dapr, and then Dapr sends it back to you encrypted/decrypted as soon as it's ready

_(However, don't wait for each chunk; just write and read in parallel)_

ItalyPaleAle (commented):
This is awesome, thanks for writing this as I know it wasn't the simplest code :) I mentioned to you before that I'm really bad at C#, especially when streaming is involved, so please take my comments with a grain of salt!

src/Dapr.Client/DaprClient.cs
WhitWaldo: Done

examples/Client/Cryptography/Cryptography.csproj
philliphoff: The NuGet packages now ship *both* .NET 6 *and* .NET 8 targets (as there are some minor optimizations allowed for .NET 7/8).  Given both are LTS versions, I can see valid arguments either way for the sample target versions.  I'd say stick with .NET 6 for now, for the sake of consistency with the others.  In November, when .NET LTS ends for .NET 6, it'd make sense to move the samples as a whole to .NET 8.

src/Dapr.Client/CryptographyEnums.cs
```diff
@@ -0,0 +1,76 @@
...
+    /// <summary>
+    /// The algorithm used for key wrapping cryptographic operations.
+    /// </summary>
+    public enum KeyWrapAlgorithm
+    {
+        /// <summary>
+        /// Represents the AES key wrap algorithm.
+        /// </summary>
+        [EnumMember(Value="AES")]
+        Aes,
```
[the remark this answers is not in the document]
philliphoff: Yes, it's a .NET/C# convention; acronyms longer than 2-letters are to have only the first letter capitalized (e.g. `Json` and `Xml`, but `IO`).

src/Dapr.Client/DaprClientGrpc.cs
philliphoff: This was a factor in the suggestion for an options object; I think it reasonable to allow users to override the block size default, but I also think it would be too messy to try to do that via method overloads.

src/Dapr.Client/DaprClientGrpc.cs
philliphoff: This is a good catch, as otherwise the request is getting implicitly buffered in the runtime.  I recall having to do something similar in the Dapr .NET pluggable components SDK to support the pub-sub and binding streaming calls.

src/Dapr.Client/DaprClientGrpc.cs
WhitWaldo: Added to the options with some notes in the XML documentation about maxing at 64. Typed as a uint to limit any additional validation needs.

ItalyPaleAle: I don't see many situations where users may need to change the chunk size personally

examples/Client/Cryptography/Cryptography.csproj
WhitWaldo: Downgraded to .NET 6

src/Dapr.Client/DaprClientGrpc.cs
WhitWaldo: Either one results in an allocation via `ToArray()` ultimately and a MemoryStream would have a slightly higher overhead than a list. As either is just a buffer for the bytes, I think there's little value in swapping over to a MemoryStream, but it was a worthwhile learning experience to give it a shot.

src/Dapr.Client/DaprClient.cs
WhitWaldo: Updated all uses of "plainText" and "cipherText" to "plaintext" and "ciphertext" for consistency.

WhitWaldo: I gave the options approach a whirl and I think you're right - it lends to a cleaner shape that can be augmented further later should more options emerge. Also added an option to configure the number of bytes to stream to the sidecar in each chunk.

src/Dapr.Client/CryptographyEnums.cs
```diff
@@ -0,0 +1,76 @@
...
+        /// <summary>
+        /// An alias for the AES key wrap algorithm.
+        /// </summary>
+        [EnumMember(Value="AES")]
```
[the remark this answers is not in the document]
WhitWaldo: Updated both that one and RSA -> "RSA-OAEP-256" per the [documented options](https://github.com/dapr/dapr/blob/b99cf8b5476a503f655d74ce478e681a1f9a37f0/dapr/proto/runtime/v1/dapr.proto#L974C27-L974C27) in the dapr.proto.

src/Dapr.Client/DaprClientGrpc.cs
WhitWaldo: @philliphoff Mentioned this in another review note earlier, but the idea here was based on our conversation in Discord where you called out that the gRPC implementation doesn't support unbounded streams. Given that there's a specific length to the data, while the sidecar can work on this in chunks, I was concerned that by exposing a Stream response method, there might be developers that misunderstand that to create perpetual streams. By instead accepting a Stream input and returning a (now `ReadOnlyMemory<byte>`), it makes clear that there's a boundary to the data given and received and the SDK handles buffering the response to take that off the developer's plate.

I'm not opposed to adding an overload that returns a stream, but that was my reasoning in the first pass. Thoughts?

src/Dapr.Client/DaprClientGrpc.cs
WhitWaldo: Great catch - updated in the latest commit for both encrypt and decrypt.

src/Dapr.Client/DaprClientGrpc.cs
ItalyPaleAle: I'm not sure I understand what you mean with the concerns about unbounded streams.

The API is designed to use bi-di streams just as a way to be able to get back the data that is encrypted (or decrypted) as soon as it's processed. In the Go and JS implementations, we are able to encrypt files that are multiple GBs in size without needing to buffer more than a few KBs in memory at any time.

The idea would be to return a readable stream to the caller. I am not sure how that is implemented in C#, but the caller would receive a stream and can attach a reader to that to consume the data as it comes in.

The non-streaming implementation would just be a wrapper around the method above, which buffers the data in memory and returns a byte array.

WhitWaldo: Sure, all that makes sense to me. I've got no problem with adding another method to return a streamed response, but I fear it's going to get messy. Try as I might, I couldn't find a way to use the generated prototypes to make a generic implementation, so we're going to instead get several duplicated private methods for each. I'll add another commit in the next couple of hours.

WhitWaldo: This now has encrypt and decrypt methods that return both `ReadOnlyMemory<byte>` and `IEnumerableAsync<byte[]>`. I refactored the implementations for each to minimize any duplicate code, so this should be in a better place now for byte-sized amounts of data or large streams.

src/Dapr.Client/DaprClient.cs
philliphoff: Proposed moving to `RequiresPreviewFeaturesAttribute` in #1219.

src/Dapr.Client/DaprClient.cs
philliphoff: I think there are plenty of uses in .NET of APIs that make use of values that are *potentially* unbounded (think `IEnumerable`) but which are generally expected to be bounded, and there generally isn't a problem.  Good, explicit documentation can help there.

src/Dapr.Client/CryptographyOptions.cs
```diff
@@ -0,0 +1,52 @@
...
+        /// </summary>
+        public KeyWrapAlgorithm KeyWrapAlgorithm { get; set; }
+
+        /// <summary>
+        /// The size of the block in bytes used to send data to the sidecar for cryptography operations.
+        /// </summary>
+        /// <remarks>
+        /// This defaults to 4KB and generally should not exceed 64KB.
+        /// </remarks>
+        public uint StreamingBlockSizeInBytes { get; set; } = 4 * 1024;
```
philliphoff, line 29: I generally prefer to apply defaults in the implementation rather than in the options object (i.e. `public uint? StreamingBlockSizeInBytes { get; set }`), else it can be difficult to know whether a value was explicitly set or not which is sometimes helpful in the implementation.

philliphoff: It also helps enable the API to allow omitting an options object altogether for simple cases.

src/Dapr.Client/DaprClient.cs
```diff
@@ -939,6 +940,286 @@ public async Task<StateEntry<TValue>> GetStateEntryAsync<TValue>(string storeNam
...
+        /// Decrypts the specified ciphertext bytes using the Dapr Cryptography encryption functionality.
+        /// </summary>
+        /// <param name="vaultResourceName">The name of the vault resource used by the operation.</param>
+        /// <param name="ciphertextStream">The stream containing the bytes of the ciphertext value to decrypt.</param>
+        /// <param name="keyName">The name of the key to use from the Vault for the decryption operation.</param>
+        /// <param name="options">Options informing how the decryption operation should be configured.</param>
+        /// <param name="cancellationToken">A <see cref="CancellationToken"/> that can be used to cancel the operation.</param>
+        /// <returns>An array of decrypted bytes.</returns>
+        [Obsolete("The API is currently not stable as it is in the Alpha stage. This attribute will be removed once it is stable.")]
+        public abstract Task<ReadOnlyMemory<byte>> DecryptAsync(string vaultResourceName, Stream ciphertextStream, string keyName, DecryptionOptions options,
```
philliphoff, line 1009: I wonder if we could start with just two basic forms of each method, one which accepts a stream and returns an async enumerator (the "efficient" case) and one which accepts a whole buffer and returns a whole buffer, and see how that goes with users.  I find the additional "accepts a stream but returns a whole buffer" a bit unusual and I think makes the naming complicated.

src/Dapr.Client/DaprClient.cs
```diff
@@ -939,6 +940,286 @@ public async Task<StateEntry<TValue>> GetStateEntryAsync<TValue>(string storeNam
...
+        /// Encrypts a stream using the Dapr Cryptography encryption functionality.
+        /// </summary>
+        /// <param name="vaultResourceName">The name of the vault resource used by the operation.</param>
+        /// <param name="plaintextStream">The stream containing the bytes of the plaintext value to encrypt.</param>
+        /// <param name="keyName">The name of the key to use from the Vault for the encryption operation.</param>
+        /// <param name="encryptionOptions">Options informing how the encryption operation should be configured.</param>
+        /// <param name="cancellationToken">A <see cref="CancellationToken"/> that can be used to cancel the operation.</param>
+        /// <returns>An array of encrypted bytes.</returns>
+        [Obsolete("The API is currently not stable as it is in the Alpha stage. This attribute will be removed once it is stable.")]
+        public abstract IAsyncEnumerable<byte[]> EncryptStreamAsync(string vaultResourceName, Stream plaintextStream, string keyName,
```
philliphoff, line 983: `IAsyncEnumerable<ReadOnlyMemory<byte>>`?

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -1373,6 +1375,563 @@ private Autogenerated.StateItem ToAutogeneratedStateItem(StateTransactionRequest
...
+                yield return decryptResponse.Payload.Data.ToByteArray();
+            }
+        }
+        
+        /// <inheritdoc />
+        [Obsolete("The API is currently not stable as it is in the Alpha stage. This attribute will be removed once it is stable.")]
+        public override async Task<ReadOnlyMemory<byte>> DecryptAsync(string vaultResourceName,
+            ReadOnlyMemory<byte> ciphertextBytes, string keyName, DecryptionOptions decryptionOptions,
+            CancellationToken cancellationToken = default) =>
+            await DecryptAsync(vaultResourceName, new MemoryStream(ciphertextBytes.ToArray()), keyName,
```
philliphoff, line 1648: `ReadOnlyMemory<byte>` should by implicitly convertible to `byte[]`, no?  `ToArray()` will make a copy which is probably not what we want.

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -1373,6 +1375,563 @@ private Autogenerated.StateItem ToAutogeneratedStateItem(StateTransactionRequest
...
+        /// Retrieves the decrypted bytes from the decryption operation on the sidecar and returns as an enumerable stream.
+        /// </summary>
+        private async IAsyncEnumerable<byte[]> RetrieveDecryptedStreamAsync(
+            AsyncDuplexStreamingCall<Autogenerated.DecryptRequest, Autogenerated.DecryptResponse> duplexStream,
+            [EnumeratorCancellation] CancellationToken cancellationToken)
+        {
+            await foreach (var decryptResponse in duplexStream.ResponseStream.ReadAllAsync(cancellationToken)
+                               .ConfigureAwait(false))
+            {
+                yield return decryptResponse.Payload.Data.ToByteArray();
```
philliphoff, line 1639: If the method yielded `ReadOnlyMemory<byte>`, we could eliminate copying the payload data.

src/Dapr.Client/DaprClient.cs
ItalyPaleAle: I think this is a good suggestion. That's what we do in the JS SDK too.

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -1373,6 +1375,563 @@ private Autogenerated.StateItem ToAutogeneratedStateItem(StateTransactionRequest
...
+            var decryptRequestOptions = new Autogenerated.DecryptRequestOptions
+            {
+                ComponentName = vaultResourceName, KeyName = keyName
+            };
+
+            var options = CreateCallOptions(headers: null, cancellationToken);
+            var duplexStream = client.DecryptAlpha1(options);
+
+            //Run both operations at the same time, but return the output of the streaming values coming from the operation
+            var tasks = new Task<IAsyncEnumerable<byte[]>>[]
```
philliphoff, line 1570: I feel like this might be more complicated than it need to (in particular, requiring `SendCiphertextStreamAsync()` return an async enumerable just so that it can be grouped together with the parallel task, and then having to iterate and re-yield the retrieved async enumerable).  Might something like this work:

```csharp
var sendTask = SendCiphertextStreamAsync(ciphertextStream, (int)decryptionOptions.StreamingBlockSizeInBytes, duplexStream, decryptRequestOptions, cancellationToken);

var retrieveTask = RetrieveDecryptedStreamAsync(duplexStream, cancellationToken);

await Task.WhenAll(sendTask, retrieveTask)

return await retrieveTask;
```

While it's technically not necessary to `await` the send task, as the retrieve task shouldn't effectively complete unless the send tasks completes, it still valid to do so (and may avoid Roslyn warnings about tasks not being awaited).

src/Dapr.Client/DaprClient.cs
philliphoff: A `Stream` works from an API perspective but, AFAIK, there's no built-in `Stream` implementation that provides the needed semantics (having a reader read from the stream while another task writes chunks to the stream).  (There are other .NET mechanisms, such as blocks/channels, that provide those semantics but they typically don't find their ways into common or third-party APIs.)  Returning a `Stream` would mean implementing a new type and I don't know whether it's worth the effort.

The difference in usage would be something like:

```charp

await using var inputStream = ...;
await using var outputStream = ...;

// Stream-returning API

await using var encryptedDataStream = await daprClient.EncryptStreamAsync(..., inputStream, ...);

await encryptedDataStream.CopyToAsync(outputStream);

// Async-enumerable-returning API

var encryptedData = await daprClient.EncryptStreamAsync(..., inputStream, ...);

await foreach (var chunk in encryptedData)
{
  await outputStream.WriteAsync(chunk);
}

```

ItalyPaleAle: Yeah, we didn't have a suitable built-in Stream implementation in JS either.

The possible advantage of using something that implements the Stream API is that it allows composition. For example in JS (and Go) there are built-in methods to pipe streams (for example, open a file in write mode, and then pipe from the output stream into the file directly), etc. It allows for more idiomatic APIs.

I also don't know if this is the case in C#, but in Go and JS using the stream APIs allows managing backpressure too, all the way to the gRPC channel.

src/Dapr.Client/DaprClient.cs
WhitWaldo: It's a fair complaint - I left it in there since the implementation method itself accepts a `Stream` but returns a `ReadOnlyMemory<byte>` and it felt weird to have a method that did precisely that but which wasn't available to the developer to use themselves. Nonetheless, removed from the Dapr client and swapped the visibility in the gRPC implementation to private for both encrypt and decrypt.

WhitWaldo:
@philliphoff  Would appreciate your thoughts on this:  Because of the method signature conflict, I renamed the encrypt/decrypt methods that took a stream and returned an `IAsyncEnumerable<byte[]>` to `EncryptStreamAsync` and `DecryptStreamAsync`. 

Now that the Stream -> ReadOnlyMemory<byte> method is out of the picture, do you think it makes the API more accessible to keep these different names or instead expose just the EncryptAsync and DecryptAsync methods again with their various overloads?

src/Dapr.Client/DaprClient.cs
```diff
@@ -939,6 +940,286 @@ public async Task<StateEntry<TValue>> GetStateEntryAsync<TValue>(string storeNam
...
+        /// Decrypts the specified ciphertext bytes using the Dapr Cryptography encryption functionality.
+        /// </summary>
+        /// <param name="vaultResourceName">The name of the vault resource used by the operation.</param>
+        /// <param name="ciphertextBytes">The bytes of the ciphertext value to decrypt.</param>
+        /// <param name="keyName">The name of the key to use from the Vault for the decryption operation.</param>
+        /// <param name="options">Options informing how the decryption operation should be configured.</param>
+        /// <param name="cancellationToken">A <see cref="CancellationToken"/> that can be used to cancel the operation.</param>
+        /// <returns>An array of decrypted bytes.</returns>
+        [Obsolete("The API is currently not stable as it is in the Alpha stage. This attribute will be removed once it is stable.")]
+        public abstract Task<ReadOnlyMemory<byte>> DecryptAsync(string vaultResourceName, ReadOnlyMemory<byte> ciphertextBytes, string keyName, DecryptionOptions options,
```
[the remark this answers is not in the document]
WhitWaldo: Added an overload that removes DecryptionOptions.

philliphoff:
> Now that the Stream -> ReadOnlyMemory method is out of the picture, do you think it makes the API more accessible to keep these different names or instead expose just the EncryptAsync and DecryptAsync methods again with their various overloads?

I would try to use just a single method name (`EncryptAsync()` and `DecryptAsync()`), if possible, as that keeps the API completion lists shorter and groups related methods together better.

src/Dapr.Client/CryptographyOptions.cs
WhitWaldo: See, I've typically taken the other route. For a simple method like these, only letting the method be configurable through an options object, itself the sole source of default values and the variations of constructors and validations ensuring the right combination of properties are set for any given configuration, it makes the options class the sole source of truth instead of hunting around in the implementations.

Putting the default values in the implementation then to be overridden by anything coming from the options provides the opportunity for unverified conflicts and only serves to lengthen the implementation which should be focused on whatever it's doing instead of refereeing which options to use.

In my approach, the options object is easily omitted by just adding an overload without it and in the override method, create a new instance of it (knowing all the defaults are set there as expected).

If it's a Dapr convention to put the defaults in the implementation instead of the options, I'll of course abide by it, but I'd otherwise rather leave it as-is.

src/Dapr.Client/DaprClientGrpc.cs
WhitWaldo: So, I tried going down this rabbit hole and the short answer is that the ByteString exposed by protobuf only exposes a `ReadOnlySpan<byte>`. This isn't an issue in .NET 7 and up as `Memory<T>` has a new constructor that takes a `Span<T>` as a parameter thus letting us convert to a `ReadOnlyMemory<T>` without copying the payload. Unfortunately, as long as .NET 6 is supported, the only route I'm seeing that works is to copy to a byte array, so while this could yield a `ReadOnlyMemory<byte>`, it's still going to incur the allocation somewhere to get at the bytes in the payload.

I'd propose this be revisited once .NET 6 is EOL'ed later this year.

src/Dapr.Client/DaprClientGrpc.cs
WhitWaldo: Whoops, don't code tired - forgot to even apply the `Task.WhenAll` in there so this never would have even sent anything. Updated and simplified according to your suggestion.

WhitWaldo: Whoops - introduced a regression here since I wasn't even awaiting the tasks I put in the array. Fixed and simplified per your suggestion.

src/Dapr.Client/DaprClient.cs
WhitWaldo: There's little reason the developer can't enjoy the relative magic of `await foreach` and drop the result into a stream (or buffer or yield return out to something else) as they'd like. I just updated the examples to reflect encrypting data from a `FileStream` to an in-memory buffer via a `List<byte>` and then decrypting via a `MemoryStream` from that list back into another `FileStream`.

ItalyPaleAle: Ok. Make sure that handles backpressure too please!

WhitWaldo: Updated methods to use `IAsyncEnumerable<ReadOnlyMemory<byte>>` along with a refresh to the examples.

src/Dapr.Client/DaprClientGrpc.cs
WhitWaldo: `byte[]` is implicitly convertible to a `ReadOnlyMemory<byte>`, but not vice versa. The best alternative I can find is using a MemoryMarshal to create a reference to the byte array and pass that in instead of copying it, but as this is exposed via `TryGetArray`, it begs the new question of what the exception message should be if it fails for some reason. Is there any documented guidance on the shape of exceptions for the SDK?

src/Dapr.Client/CryptographyOptions.cs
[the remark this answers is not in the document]
WhitWaldo: Updated

src/Dapr.Client/DaprClient.cs
WhitWaldo: I'm sold - updated the implementation last night accordingly

src/Dapr.Client/DaprClientGrpc.cs
philliphoff: Huh, interesting, as the [docs](https://cloud.google.com/dotnet/docs/reference/Google.Protobuf/latest/Google.Protobuf.ByteString#Google_Protobuf_ByteString_Memory) show `ByteString` having a `ReadOnlyMemory<byte> Memory` property in addition to a `ReadOnlySpan<byte> Span` property.  It's possible that that is the case only in more recent versions of the gRPC NuGet packages, though.

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -1373,6 +1376,582 @@ private Autogenerated.StateItem ToAutogeneratedStateItem(StateTransactionRequest
...
+            await Task.WhenAll(
+                //Stream the ciphertext data to the sidecar in chunks
+                Task.FromResult(SendCiphertextStreamAsync(ciphertextStream, decryptionOptions.StreamingBlockSizeInBytes,
+                    duplexStream, decryptRequestOptions, cancellationToken)),
+                //At the same time, retrieve the decrypted response from the sidecar
+                receiveResult
+            );
+            
+            //Return only the result of the `RetrieveEncryptedStreamAsync` method
+            await foreach (var item in receiveResult.Result.WithCancellation(cancellationToken))
```
philliphoff, line 1580: It occurs to me that, if we're re-iterating and re-yielding the `receiveResult`, do we need to wait for either task above?  Alternatively, can `receiveResult.Result` just be returned without the iteration?

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -1373,6 +1376,582 @@ private Autogenerated.StateItem ToAutogeneratedStateItem(StateTransactionRequest
...
+            {
+                yield return item;
+            }
+        }
+
+        /// <summary>
+        /// Sends the plaintext bytes in chunks to the sidecar to be encrypted.
+        /// </summary>
+        private async IAsyncEnumerable<byte[]> SendPlaintextStreamAsync(Stream plaintextStream,
```
philliphoff, line 1495: Can this be made to return `Task` rather than an (unused) async enumerable?

src/Dapr.Client/DaprClientGrpc.cs
philliphoff: It looks like the SDK's existing dependency version has the `Memory` property as well.

src/Dapr.Client/DaprClientGrpc.cs
philliphoff: I see. Ach!  The purpose of accepting `ReadOnlyMemory<byte>` was to be more flexible for callers, but if we have to essentially restrict it to just array-like memory (or else make a copy), I'm wondering if that flexibility is really worth it.  (Note that there *is* a performant `Stream` implementation over `ReadOnlyMemory<byte>` in the [.NET "high efficiency" community toolkit](https://learn.microsoft.com/en-gb/dotnet/communitytoolkit/high-performance/introduction).  That said, I don't think we want to go down that path of taking on a new dependency or having our own implementation just for this purpose.)

I'm starting to feel like accepting (for now) `byte[]` as the better approach for now.  (And also returning `byte[]` to keep the symmetry of the method.)

src/Dapr.Client/DaprClientGrpc.cs
WhitWaldo: You're right - I don't know how I missed that yesterday. I've swapped this out to remove the `ToByteArray()` method (and its allocation) to instead point to its `Memory` property, then updated the `EncryptDataAsync` and `DecryptDataAsync` methods to both replace the `List<byte>` buffers with `ArrayBufferWriter<byte>` so it can similarly avoid any array copies.

src/Dapr.Client/DaprClientGrpc.cs
WhitWaldo: It certainly can. Done.

src/Dapr.Client/DaprClientGrpc.cs
WhitWaldo: It looks like [their implementation](https://github.com/CommunityToolkit/dotnet/blob/7b53ae23dfc6a7fb12d0fc058b89b6e948f48448/src/CommunityToolkit.HighPerformance/Streams/MemoryStream.cs#L32) to make a memory stream from a `ReadOnlyMemory<byte>` is awfully similar to what I've got, using the MemoryMarshal to access the byte array. 

I did copy over their argument exception text as it's richer than what I had.

As such, I vote leaving this as-is. It's about as high-performant as it'll probably get with this version of .NET and there's nothing that stops the developer from simply passing in a `byte[]` to the method if that's what they have or a `ReadOnlyMemory<byte>` if they happen to have that as well (as opposed to making this the developer's problem by reverting to only accepting `byte[]` and leaving them having to do this same research to convert their own `ReadOnlyMemory<byte>` accordingly).

src/Dapr.Client/DaprClientGrpc.cs
WhitWaldo: Updated accordingly - removed the second (and unnecessary) `await foreach` and updated the example.

philliphoff (approved):
LGTM

msfussell:
To support this code, best to also add a quickstart example at the same. I have raised the issue here https://github.com/dapr/quickstarts/issues/969

WhitWaldo:
@philliphoff Apparently I somehow dismissed your review. Mind giving it another pass?

philliphoff (approved):

philliphoff:
> Apparently I somehow dismissed your review. Mind giving it another pass?

@WhitWaldo Re-approved.  Not sure why the DCO check isn't passing; might try to re-sign the commits?

WhitWaldo:
Looks like commit #4 (2e1e6b335cc1421dc71556fea48de2c0e133011f) is missing the DCO. I know how to revert the last commit and amend it, but this is 41 commits in the past. Any chance you know the commands necessary to amend that commit?

philliphoff:
@WhitWaldo What I've done in the past is rebase and squash the changes into a single commit, then amend *that* commit with the signoff.  Or, rebase and amend each commit with the signoff (if needed). Then force push the change.  I don't know which way you'd find easier.

WhitWaldo:
If the terminal only supported multi-line commit messages, that would have been much easier to deal with. As it is, I believe I've properly merged this and the DCO check is validating as it should. @philliphoff

philliphoff:
@WhitWaldo Would you be willing to look at the differences in code coverage and see if there are some tests that could be added to bring it back to parity with the baseline?

WhitWaldo:
@philliphoff  I can give it a look, but I don't know that there's a whole lot that can be added on the .NET side since it's largely just shuffling the bytes out and vice versa, as all the magic happens on the Dapr sidecar. Give me a few days and I'll see what I can add to it that I don't already have in there. What's the baseline we're aiming for?

philliphoff:
> What's the baseline we're aiming for?
@WhitWaldo Ideally we'd end up close to the same test coverages as the baseline (i.e. prior to the changes).  The coverage maps show a good portion of the new logic that isn't covered.  This may just be a false-negative, but something to look at to see if that's the case or, if not, how easy it is to get it covered.

philliphoff (approved):

philliphoff:
Merging this to ensure it's in 1.13; more test coverage would be nice prior to the APIs being considered stable but can wait.

PR #1051, opened 2023-03-01
Enable vault name mapping and error suppression

yash-nisar:
# Description

Enable vault name mapping and error suppression

## Issue reference

We strive to have all PR being opened based on an issue, where the problem or feature have been discussed prior to implementation.

Please reference the issue this PR will close: https://github.com/dapr/dotnet-sdk/issues/887

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [x] Code compiles correctly
* [x] Created/updated tests
* [ ] Extended the documentation

PR #1231, opened 2024-02-09
Enable vault name mapping and error suppression

jamesmcroft:
# Description

Enable vault name mapping and error suppression. This PR supersedes #1051 including the original changes from @yash-nisar 

## Issue reference

Please reference the issue this PR will close: #887 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


src/Dapr.Extensions.Configuration/DaprSecretDescriptor.cs
```diff
@@ -31,20 +31,51 @@ public class DaprSecretDescriptor
...
 
         /// <summary>
-        /// Secret Descriptor Construcutor
+        /// This flag indicates if this field's existence should trigger an exception. Setting it to "false"
+        /// will suppress the exception whereas setting it to "true" will not suppress it.
+        /// </summary>
+        public bool IsRequired { get; }
+
+        /// <summary>
+        /// SecretKey value is mapping value to Vault Name. If The application's Secret Name and Secret in the
```
philliphoff, line 40: Nit: the description could be more agnostic as to the backend service.  That is, it may be something other than Azure Key Vault.  The description for `SecretName` could be improved to better highlight the differences between the two.

src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationProvider.cs
```diff
@@ -197,16 +201,39 @@ private async Task LoadAsync()
...
+
+                        result = new Dictionary<string, string>();
+                    }
 
                     foreach (var key in result.Keys)
                     {
                         if (data.ContainsKey(key))
                         {
-                            throw new InvalidOperationException($"A duplicate key '{key}' was found in the secret store '{store}'. Please remove any duplicates from your secret store.");
+                            throw new InvalidOperationException($"A duplicate key '{key}' was found in the " +
```
philliphoff, line 230: Nit: was there a specific reason to split this up aside from, perhaps, overall line length (which I wouldn't be concerned with for a string)?

src/Dapr.Extensions.Configuration/DaprSecretDescriptor.cs
```diff
@@ -31,20 +31,51 @@ public class DaprSecretDescriptor
...
+        public DaprSecretDescriptor(string secretName, IReadOnlyDictionary<string, string> metadata, bool isRequired) :
+            this(secretName, metadata, isRequired, secretName)
+        {
+        }
+
+        /// <summary>
+        /// Secret Descriptor Constructor
         /// </summary>
-        public DaprSecretDescriptor(string secretName, IReadOnlyDictionary<string, string> metadata)
+        public DaprSecretDescriptor(string secretName, IReadOnlyDictionary<string, string> metadata, bool isRequired, string secretKey)
```
philliphoff, line 73: Consider making `isRequired` and `secretKey` both optional arguments, to help reduce the explosion of constructor overloads as well as, say, allow users to *only* specify the `secretKey` but not have to then care about the `isRequired` argument.  We might also consider doing the same for `metadata`.

Ideally these would all (save for `secretName`) bet `init` properties on the class, but that's a larger breaking change (depending on how many people *actually* specify metdata with each secret).

philliphoff (changes requested):
Overall the approach is sound, but I had some feedback on the additional constructors and console statements.

jamesmcroft:
Thanks for the review @philliphoff. I will review your comments and make any amendments

src/Dapr.Extensions.Configuration/DaprSecretDescriptor.cs
jamesmcroft: Removed the constructor overload bloat and consolidated this change into the existing 2 constructors using optional arguments.

src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationProvider.cs
jamesmcroft: I have reverted this change, not sure how this occurred.

src/Dapr.Extensions.Configuration/DaprSecretStoreConfigurationProvider.cs
```diff
@@ -197,16 +201,39 @@ private async Task LoadAsync()
...
+                            .ConfigureAwait(false);
+                    }
+                    catch (DaprException)
+                    {
+                        if (secretDescriptor.IsRequired)
+                        {
+                            throw;
+                        }
+                            
+                        Console.WriteLine($"'{secretDescriptor.SecretName}' doesn't exists in Key Vault but because " +
```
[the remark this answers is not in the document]
jamesmcroft: I agree with your second comment, this `Console.WriteLine` is redundant and I have removed this.

src/Dapr.Extensions.Configuration/DaprSecretDescriptor.cs
jamesmcroft: I've updated the descriptions on the properties in the descriptor to be more reflective of their usage. Please let me know if this now makes sense.

philliphoff (approved):

Issue #937, opened 2022-08-25
HttpClient created by DaprClient.CreateInvokeHttpClient can't find the target service to invoke


bledbet:
Per the recommendation here:
https://docs.microsoft.com/en-us/dotnet/architecture/dapr-for-net-developers/service-invocation#invoke-http-services-using-httpclient

I am trying to use the HttpClient that is created by DaprClient.CreateInvokeHttpClient() to make a service invocation call in a self-hosted dapr environment.  I was successful in modifying the "service_invocation" quickstart by modifying Program.cs in the checkout app, replacing the whole section before the 'for' loop that creates and configures an HttpClient with just

```
var client = DaprClient.CreateInvokeHttpClient("orderProcessor");
```

and that worked fine.

Now I am trying to add service invocation to a larger demo that already has some other working api endpoints, state store, and pubsub with working subscriber services, and I can't seem to get it to work with the HttpClient that is produced by tDaprClient.CreateInvokeHttpClient().

In my larger demo, if I build the HttpClient as the original "service_involcation" quickstart did, like so:

```
    private static HttpClient CreateHttpClientForDaprApp(string appId)
    {
        var daprBaseUrl = (Environment.GetEnvironmentVariable("BASE_URL") ?? "http://localhost") + ":" + (Environment.GetEnvironmentVariable("DAPR_HTTP_PORT") ?? "3500");

        var httpClient = new HttpClient();
        httpClient.BaseAddress = new Uri(daprBaseUrl);
        httpClient.DefaultRequestHeaders.Add("dapr-app-id", appId);
        httpClient.DefaultRequestHeaders.Accept.Add(new System.Net.Http.Headers.MediaTypeWithQualityHeaderValue("application/json"));

        return httpClient;
    }
```

my service invocation call works fine, but if I replace this with a call to DaprClient.CreateInvokeHttpClient(appId), I get 500 error responses like  

{"errorCode":"ERR_DIRECT_INVOKE","message":"fail to invoke, id: orderapi, err: couldn't find service: orderapi"}

and 

{"errorCode":"ERR_DIRECT_INVOKE","message":"fail to invoke, id: orderapi, err: timeout waiting for address for app id orderapi"}

The service that is trying to make the invocation call is a subscriber to an event, and what I've commonly seen is that the first time it tries to process the event it gets that first error response message above, and upon retry it gets the second response message.

The command being used to launch the Order API service is 

```
[...]

PR #1233, opened 2024-02-09
Handle the case where appid contains at least one upperletter

TWEESTY:
# Description

The .NET SDK Client for Dapr does not handle the case where the app id contains an upper character for HTPP service invocation, because it uses the property Host of the URI, which is case insensitive (always in lower case), inside the class InvocationHandler.

The change is to use the appid sent when creating the HTTPClient or if there is no appId, use the host issued by the original string of the URI to get the correct app id (i.e. case sensitive).

The DaprClient can always be used for another appid if the user put an absolute URI.

There is no breaking change.

## Issue reference

#_[#937]_

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [ Not relevant] Extended the documentation


src/Dapr.Client/InvocationHandler.cs
```diff
@@ -128,17 +136,56 @@ internal bool TryRewriteUri(Uri? uri, [NotNullWhen(true)] out Uri? rewritten)
...
+            string host;
+
+            // It is just for optimization, to not add some overhead time
+            if (this.DefaultAppId is not null && uri.Host.Equals(this.DefaultAppId, StringComparison.InvariantCultureIgnoreCase))
+            {
+                host = this.DefaultAppId;
+            }
+            else
+            {
+                host = this.GetOriginalHostFromUri(uri);
```
philliphoff, line 148: I wonder if we really need to be concerned about this case; it's only if the caller of `DaprClient.CreateInvokeHttpClient()` does *not* provide an `appId`, right?  That use case is to have a "generic" HTTP client that can call any Dapr application and I wonder how often that's done vs. having a HTTP client that's specific to a specific Dapr application.

If the `appId` is omitted, then they would get the prior behavior (with attempts to use case-sensitive app IDs in those URIs failing).  Requiring callers to pass `appId` if case-sensitivity is an issue seems like a reasonable starting point; if users find they need the case-sensitivity with the more generic case, then we could look at supporting that.  (An alternative is, always, to *not* use case-sensitive app IDs in the first place.)

All that to say, I'd prefer to avoid all the URL parsing can of worms just to pick out the original host, if there's no significant demand for that use case.

philliphoff (changes requested):
I'd prefer to avoid the URL parsing if the principal scenario is unblocked without it.

TWEESTY:
> I'd prefer to avoid the URL parsing if the principal scenario is unblocked without it.
 
Alright I've updated the code in this way?

philliphoff (approved):

Issue #1235, opened 2024-02-13
AddDaprClient incompatible with keyed services


onionhammer:
## Expected Behavior
When adding dapr client to service builder, this should not error

https://weblogs.asp.net/ricardoperes/net-8-dependency-injection-changes-keyed-services

```C#
// Add dapr client
builder.Services.AddDaprClient(client =>
{
    client.UseJsonSerializationOptions(JsonExtensions.JsonOptions);
});
```

## Actual Behavior

When adding dapr to your serviceBuilder, if there are keyed services present, adding dapr fils with exception:

```
Unhandled exception. System.InvalidOperationException: This service descriptor is keyed. Your service provider may not support keyed services.
```

## Steps to Reproduce the Problem
Add a dapr client to an application with keyed services defined, i.e. with Orleans:

```C#
builder.Host.UseOrleans(siloBuilder =>
{
    siloBuilder.AddActivityPropagation();
});
```

artyom-p:
Same here, related to [this](https://github.com/dotnet/runtime/issues/95789) issue.

![image](https://github.com/dapr/dotnet-sdk/assets/3993463/18a09c54-92b6-4b3b-b3fa-b9c537663e20)

Sen-Gupta:
I'm not sure it this works for you, I placed AddDapr above all calls in service registration and it works.

Additionally, I downgraded Assemblies to 8.0.0 for DependencyInjection.

Sent from Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

rliberoff:
Thank you @Sen-Gupta.

The issue is not with `AddDapr` but with `AddDaprClient`.

This happens with the newest .NET Aspire Preview 3.

Sen-Gupta:
Apologies, I meant that only.

I attached the file, I would be able to share more details tomorrow.

It is midnight in my time zone.

Regards
Sen


Sent from Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

rliberoff:
From further investigation, we found out that the offending dependency is with library `Microsoft.Extensions.Http.Resilience` when updated from version `8.1.0` to `8.2.0`. Hope it helps.

Sen-Gupta:
@rliberoff 

Double Confirmation!

That's correct. I did upgrade Microsoft.Extensions.Http.Resilience to 8.2.0.

PR #1238, opened 2024-02-15
Register Dapr services in keyed-compatible manner

philliphoff:
# Description

.NET 8 adds the ability to register "keyed" services with a service collection. The implementation of this feature, however, breaks how many existing components look through that service collection to determine whether they've been previously registered (a common if, perhaps, now an anti-pattern with the advent of `TryAddSingleton()`).  If a component registers a keyed service *before* another component performs such a search, that search cause an exception to be thrown.

The Dapr .NET SDK uses this "old" style of registering "marker" services to help prevent multiple registrations, which triggers this new behavior in .NET 8.  This change switches to the newer `TryAddSingleton()` approach to avoid that problem.  

## Issue reference

Please reference the issue this PR will close: #1235 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


philliphoff:
@halspang Would appreciate a quick review when you have a chance, thanks!

rliberoff:
Hello!

I hope this helps: from further investigation, we found out that the offending dependency is with library `Microsoft.Extensions.Http.Resilience` when updated from version `8.1.0` to `8.2.0`. Hope it helps.

When using that library, the `AddDaprClient` throws the exception that is mentioned in the issue.

My Team and I really appreciete your effor guys for providing an update for this. Thank you.

halspang (approved):

onionhammer:
When is this getting released (non-preview)?

philliphoff:
> When is this getting released (non-preview)?

@onionhammer It will be released as part of the larger Dapr 1.13 release, coming very soon.

Issue #1255, opened 2024-03-20
GetBulkStateAsync returns empty items for missing keys


adrianhr91:
I'm using the Dapr State Store functionality to save some records that later will be retrieved from the store. When I'm retrieving data based on a key, it won't always find an associated record. My expectation was that the `GetBulkStateAsync()` method will return only a list of the records in the store with matching keys. Instead, even for records that are not in the store, it returns a key with a value of `null`

![image](https://github.com/dapr/dotnet-sdk/assets/2922498/ee540932-007d-403b-a7b2-3656e393d622)

I found this confusing since the `BulkStateItem.Value` is not nullable, the compiler did not complain and I didn't realise this is a problem until runtime.

Am I missing something about the expected default behaviour? And even then, should the `BulkStateItem.Value` be nullable?

WhitWaldo:
I think providing a null response for each key is more appropriate than pretending that you never asked for the key in the first place, particularly since the SDK is just operating off the information available to it from the Dapr sidecar (which itself is returning null values for missing keys rather than more explicitly indicating they're not found). I agree that the documentation could be updated to reflect this possibility and leave a more informed developer of the possibilities that a null value indicates either:
1) that the key wasn't found in the state store; or 
2) that the key was found, but had a value of null

To that end, I think it makes sense to mark the `BulkStateItem.Value` as nullable and update the documentation, but still include the requested key in the response (with that null value). 

What do you think @adrianhr91 ?

adrianhr91:
Yeah, I think as a minimum the value should be nullable as otherwise you just don't get any compile time warning. 👍

WhitWaldo:
@adrianhr91 What're your thoughts on exposing a `ConditionalValue<T>` as the return type for each value that indicates whether the value exists or not for each key instead of simply returning a null?

adrianhr91:
I'd need to write up some example code but it feels like it should be possible to use some sort of c# language feature to force return types to be nullable without having to use a custom type to represent that.

I appreciate some of this is how the actor state is implemented so perhaps better to go with it for consistency.

Either way, I agree that conceptually it makes sense to force users of the SDK to check if there is a value or not since otherwise you only discover this problem after it has occurred.

WhitWaldo:
[...]

PR #1259, opened 2024-04-03
Marked bulk state `Value` property as nullable

WhitWaldo:
# Description

If a request is made for bulk state for a set of keys for which there isn't matching data, a null value is returned for that key in the result. An issue was raised that indicated that the value isn't marked as nullable, so the author was surprised to find that the value for an unregistered key returned null given that so many other packages are marked with such annotations these days.

As such, I marked the file (not the larger project) with nullable annotations and indicated that the `Value` property is nullable as well. This requires no changes on the Dapr sidecar itself and simply updates the C# package to reflect what's already returned so developers are up to speed on expectations.

## Issue reference

Please reference the issue this PR will close: #1255 

## Checklist

* [X] Code compiles correctly
* [ ] Created/updated tests (not applicable)
* [ ] Extended the documentation (not applicable)


philliphoff (approved):
Adding null-ability on a case-by-case basis is probably the best we can do at the moment as, despite being tremendously useful (in my opinion), a SDK-wide review is unlikely to rise to the top of any priority stack.

Thanks for taking a look at this!

marcduiker:
@holopin-bot @WhitWaldo Thank you!

Issue #96, opened 2019-10-14
Samples - Add k8s deployment yaml to demoactor sample


yaron2:
It'd be helpful to create a Dockerfile and detail how one goes about building the C# actor into a Docker image and then deploying it to Kubernetes with Dapr.

m3nax:
@yaron2 done

philliphoff:
Closed with #1308.

PR #1308, opened 2024-06-17
Samples - Add k8s deployment yaml to DemoActor sample

m3nax:
# Description

Add Dockerfile to build DemoActor example and a manifest to deploy it inside k8s cluster.

## Issue reference

- #96 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


m3nax:
image
 
I can't see which test failed

examples/Actor/DemoActor/Dockerfile
```diff
@@ -0,0 +1,28 @@
+﻿FROM mcr.microsoft.com/dotnet/aspnet:6.0-jammy-chiseled-extra AS base
```
philliphoff, line 1: It'd be nice to avoid having to maintain a `Dockerfile` going forward (or taking on VS dependencies just for the purpose of creating containers).  I'd suggest using the [built-in .NET CLI support for publishing applications as containers](https://learn.microsoft.com/en-us/dotnet/core/docker/publish-as-container?pivots=dotnet-8-0#publish-net-app), which should work even for our .NET 6 based example/sample projects.

examples/Actor/DemoActor/build-container.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
```
philliphoff, line 1: Given the commands within the scripts are the same across platforms, I'd just add the commands directly to the `README` to avoid having to maintain them.

philliphoff (changes requested):
Generally looks ok to me, but would like to keep the repo out of the container generation business as much as possible.

examples/Actor/DemoActor/Dockerfile
m3nax: I removed the Dockerfile and used the integrated container support, i also updated the readme to reflect the changes

examples/Actor/DemoActor/build-container.sh
m3nax: Done

examples/Actor/DemoActor/DemoActor.csproj
```diff
@@ -1,13 +1,24 @@
...
-  </ItemGroup>
+    <PropertyGroup>
+        <TargetFramework>net6</TargetFramework>
+    </PropertyGroup>
+
+    <PropertyGroup>
+        <IsPublishable>true</IsPublishable>
+        <EnableSdkContainerSupport>true</EnableSdkContainerSupport>
+        <ContainerRepository>demo-actor</ContainerRepository>
+        <ContainerBaseImage>mcr.microsoft.com/dotnet/runtime:6.0-jammy-chiseled-extra</ContainerBaseImage>
```
philliphoff, line 11: Nit: can we get away without specifying a specific base image?  Given it's a sample, we shouldn't be particularly concerned about image size, etc.  I'd rather not have to remember to update this as the target framework changes.

m3nax: Done

m3nax:
@philliphoff DCO is really annoying

philliphoff:
> DCO is really annoying

@m3nax Yes, quite.

m3nax:
@philliphoff All check passed, DCO signed

philliphoff (approved):

PR #1309, opened 2024-06-19
Remove unused stuff

RafaelJCamara:
# Description

Removed unused references, and variables.
Simplified namespaces.
Adjusts in tests, because their build was failing because of not awaiting async method calls.

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


src/Dapr.Actors/ActorReference.cs
```diff
@@ -15,7 +15,7 @@ namespace Dapr.Actors
 {
     using System;
     using System.Runtime.Serialization;
-    using Dapr.Actors.Client;
+    using Client;
```
philliphoff, line 18: I'm personally not a fan of abbreviating namespaces in this way, even if it's technically allowed.  For `Dapr.Client`, in particular, there's just too many other "client" types and I'd prefer to be explicit.  (If anything, I would propose moving `using` statements to be above the `namespace` declaration and use the `namespace <ns>;` form rather than `namespace <ns> { /* ... */ }`.  That's a significant refactor, however, so it's only been done on a piecemeal basis when working within individual files.)

philliphoff (changes requested):
While I appreciate the effort to cleanup the codebase--and there's plenty to do there--this is a hefty PR.  While the changes all appear individually trivial, it would be easy to miss a change that might have unintended consequences.  I would prefer to split this up into several PRs that each address a particular thing (e.g. removal of unused `using` statements, another that cleaned up unused fields/values, etc.) which will make the changes easier to review.

RafaelJCamara:
> While I appreciate the effort to cleanup the codebase--and there's plenty to do there--this is a hefty PR. While the changes all appear individually trivial, it would be easy to miss a change that might have unintended consequences. I would prefer to split this up into several PRs that each address a particular thing (e.g. removal of unused `using` statements, another that cleaned up unused fields/values, etc.) which will make the changes easier to review.

I'll split that in multiple PRs! Thank you so much for the feedback!

Issue #1303, opened 2024-06-13
Passing query string parameters on method invocation


WhitWaldo:
The [service invocation documentation](https://docs.dapr.io/reference/api/service_invocation_api/) speaks to a body-based payload (via POST) and doesn't mention query strings anywhere.

I was looking to use Dapr to invoke a method on an API controller that required query string parameters. On a whim, I jammed the query string to the end of the request URL to see what it'd do and to my surprise, it worked precisely as expected. To that end, I thought I'd file this ticket to confirm this is happening as it was expected to and if so, see if there's any interest in a PR that adds some overloads to allow query strings to be more readily appended to the URI.

In the short term, here's what I'm currently doing that's working:
```cs
//API - MyController.cs
[ApiController]
[Route("[controller]")]
public class MyController() : ControllerBase
{
  [HttpGet]
  public IEnumerable<MyObject> Get([FromQuery(Name="c")] int count)
  {
    //Return something
  }
}

// Extensions
    public static void AddQueryParameters(
        this HttpRequestMessage? message,
        Dictionary<string, string>? queryStringParameters)
    {
        ArgumentNullException.ThrowIfNull((object)message, nameof(message));
        if (queryStringParameters == null)
            return;
        UriBuilder uriBuilder = new UriBuilder(message.RequestUri);
        string str = uriBuilder.Query;
        foreach (KeyValuePair<string, string> queryStringParameter in queryStringParameters)
        {
            if (!string.IsNullOrEmpty(str))
                str += "&";
            str = str + Uri.EscapeDataString(queryStringParameter.Key) + "=" + Uri.EscapeDataString(queryStringParameter.Value);
        }
        uriBuilder.Query = str;
        message.RequestUri = uriBuilder.Uri;
    }

//Service - Program.cs
app.MapGet("/api/data", async (DaprClient client) => 
{
  var req = daprClient.CreateInvokeMethodRequest(HttpMethod.Get, "api-service", "Get");
  req.AddQueryParameters(new Dictionary<string,string>
  {
    {"c", "2"}
  });

  return await client.InvokeMethodAsync<int[]>(req);  
});
```

Can anyone confirm this is working as fully expected?

philliphoff:
@WhitWaldo As far as I know, query parameters were intended to "just work" with Dapr service invocation and simply "forwarded" to the application.  You might submit an issue in the [docs](https://github.com/dapr/docs) repo to have that be made explicit.

[...]

PR #1310, opened 2024-06-19
Added overload to support SDK supplying query string on invoked URL

WhitWaldo:
# Description
When a method is invoked via the Dapr service invocation building block, a query string can be passed through on the invocation URL without issue. However, when using the .NET SDK, this behavior isn't exposed anywhere. Existing overloads allow the app ID to be specified, the method named, the payload to be provided, but that's it - it's up to the developer to realize that a query string is supported, then augment the `RequestUri` property of the returned `HttpRequestMessage` to append their query string values, and proceed with the invocation.

This PR provides overloads to the invocation methods to allow an `IReadOnlyCollection<KeyValuePair<string,string>>` to be provided as the collection of key/value pairs to build/supplement the query string with. This uses a static extension method I've contributed (and added tests for) that appends the encoded key/value pairs provided in the collection to build a standards-matching query string so developers don't have to maintain the same on their own.

Why an `IReadOnlyCollection` instead of a `Dictionary<string,string>`? Because identical keys [can exist more than once in a query string](https://url.spec.whatwg.org/#example-searchparams-sort) with different values, so this needs to be more of a collection and less of a mapping. And because at the end of the day, this is being written to a string-based URI, these methods expect string-based keys and values, leaving it to the developer to figure out how to serialize accordingly before calling this.

Finally, I had the existing concrete implementations point to the new overloads and pass empty collections so as to avoid code duplication.

## Issue reference
Please reference the issue this PR will close: #1303 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests

Documentation PR extended here, though generically and not in a C#-specific way: [https://github.com/dapr/docs/pull/4217](https://github.com/dapr/docs/pull/4217)

### DIFF
--- a/src/Dapr.Client/Extensions/HttpExtensions.cs
+++ b/src/Dapr.Client/Extensions/HttpExtensions.cs
@@ -14,7 +14,6 @@
 #nullable enable
 using System;
 using System.Collections.Generic;
-using System.Net.Http;
 using System.Text;
 
 namespace Dapr.Client
@@ -22,21 +21,21 @@ namespace Dapr.Client
     /// <summary>
     /// Provides extensions specific to HTTP types.
     /// </summary>
-    public static class HttpExtensions
+    internal static class HttpExtensions
     {
         /// <summary>
         /// Appends key/value pairs to the query string on an HttpRequestMessage.
         /// </summary>
-        /// <param name="message">The message to append the query string parameters to.</param>
+        /// <param name="uri">The uri to append the query string parameters to.</param>
         /// <param name="queryStringParameters">The key/value pairs to populate the query string with.</param>
-        public static void AddQueryParameters(this HttpRequestMessage? message,
+        public static Uri AddQueryParameters(this Uri? uri,
             IReadOnlyCollection<KeyValuePair<string, string>>? queryStringParameters)
         {
-            ArgumentNullException.ThrowIfNull(message, nameof(message));
-            if (queryStringParameters is null || message.RequestUri is null)
-                return;
+            ArgumentNullException.ThrowIfNull(uri, nameof(uri));
+            if (queryStringParameters is null)
+                return uri;
 
-            var uriBuilder = new UriBuilder(message.RequestUri);
+            var uriBuilder = new UriBuilder(uri);
             var qsBuilder = new StringBuilder(uriBuilder.Query);
             foreach (var kvParam in queryStringParameters)
             {
@@ -46,7 +45,7 @@ namespace Dapr.Client
             }
 
             uriBuilder.Query = qsBuilder.ToString();
-            message.RequestUri = uriBuilder.Uri;
+            return uriBuilder.Uri;
         }
     }
 }
--- a/src/Dapr.Client/Extensions/HttpExtensions.cs
+++ b/src/Dapr.Client/Extensions/HttpExtensions.cs
@@ -1,0 +1,52 @@
+﻿// ------------------------------------------------------------------------
+// Copyright 2024 The Dapr Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//     http://www.apache.org/licenses/LICENSE-2.0
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// ------------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


src/Dapr.Client/Extensions/HttpExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+    /// Provides extensions specific to HTTP types.
+    /// </summary>
+    public static class HttpExtensions
+    {
+        /// <summary>
+        /// Appends key/value pairs to the query string on an HttpRequestMessage.
+        /// </summary>
+        /// <param name="message">The message to append the query string parameters to.</param>
+        /// <param name="queryStringParameters">The key/value pairs to populate the query string with.</param>
+        public static void AddQueryParameters(this HttpRequestMessage? message,
```
philliphoff, line 32: A couple of comments:

 - I feel like `HttpRequestMessage` is too high-level for this; IMO it'd be better to accept/return a `Uri`, which also avoids mutation of the message, which I find a little odd for a method that's not otherwise part of a "builder" pattern.
 - I'd make this class/method `internal` if not made part of `DaprClientGrpc` itself, lest developers start using it as a general-purpose query management utility function and commit us to support the API into the future.

WhitWaldo: Fair on both counts. Updated the extension, the method and the unit tests.

philliphoff (approved):

philliphoff (changes requested):
@WhitWaldo  It looks like the integration test failures are due to something in this PR (as they succeed in `master`). Can you take a look and see what's going on?

WhitWaldo:
@philliphoff I've updated my branch and resynced my local branch to master. I'm running the tests locally and everything I contributed here is testing fine. I'm showing errors when testing locally that appear entirely unrelated to my contribution:

- InvokeMethodAsync_VoidResponseNoHttpMethod_Success, InvokeMethodAsync_VoidresponseWithHttpMethod_Success, InvokeMethodAsync_VoidVoidNotHttpMethod_Success and InvokeMethodAsync_VoidVoidWithHttpMethod_Success are all failing because of "The client has 1 or more incomplete requests. Use `request.Dismiss()` if the test is uninterested in the response." which suggests some sort of shared client issue?
- Seeing an error with TestWorkflowLogging that's failing because the file is in use by another process

I don't necessarily mind tackling either of these, but I handling them in the context of this PR may be out of scope. I'll keep poking at it and see what my changes might have to do with the error.

philliphoff:
@WhitWaldo I agree that I don't see an immediate connection to your changes, but since the tests only seem to fail in this branch, we can't commit it until that's sorted out.

src/Dapr.Client/Extensions/HttpExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// </summary>
+    internal static class HttpExtensions
+    {
+        /// <summary>
+        /// Appends key/value pairs to the query string on an HttpRequestMessage.
+        /// </summary>
+        /// <param name="uri">The uri to append the query string parameters to.</param>
+        /// <param name="queryStringParameters">The key/value pairs to populate the query string with.</param>
+        public static Uri AddQueryParameters(this Uri? uri,
+            IReadOnlyCollection<KeyValuePair<string, string>>? queryStringParameters)
```
frankbuckley, line 32: Curious why not [NameValueCollection](https://learn.microsoft.com/en-us/dotnet/api/system.collections.specialized.namevaluecollection?view=net-8.0#remarks)?

WhitWaldo: Fundamentally, I was looking to ensure that a read-only collection is passed. Conceptually, they wind up operationally working the same way. I don't see why an overload couldn't be added for a `NameValueCollection` as well.

WhitWaldo:
@philliphoff I show that all the tests are passing as of the last commit. Figured out the issue - there was some ambiguity about which method should be invoked given all the overloads available. One of them, added as part of this issue, added:
```cs
public override HttpRequestMessage CreateInvokeMethodRequest(HttpMethod httpMethod, string appId, string methodName,
    IReadOnlyCollection<KeyValuePair<string, string>> queryStringParameters)
```

The existing overload was:
```cs
public override HttpRequestMessage CreateInvokeMethodRequest<TRequest>(HttpMethod httpMethod, string appId, string methodName, TRequest data)
```

And at runtime then, it was taking the query string parameters and instead dropping them in as the request payload. I've remedied by changing the latter overload signatures to also accept the `IReadOnlyCollection<KeyValuePair<string, string>>` and simply pass it an empty collection when it's not provided and both locally and in the CI/CD, everything looks to be working now.

philliphoff (approved):

Issue #1316, opened 2024-07-03
Implementing .NET Dapr SDK-wide nullability annotations


WhitWaldo:
This is just an issue to track the larger initiative to implement nullability annotations across the whole of the .NET SDK. It's a large enough undertaking that it's worth splitting into individual PRs for each SDK rather than trying to handle across all the SDKs in a single swoop.

Target APIs:
* [ ] Dapr.Actors
* [ ] Dapr.Actors.AspNetCore
* [ ] Dapr.AspNetCore
* [ ] Dapr.Client
* [ ] Dapr.Extensions.Configuration
* [ ] Dapr.Workflow - tentatively already completed, but should verify
* [ ] Samples/Examples
* [ ] Quickstarts - will likely need to wait until all this is completed so the NuGet imports will lack the annotations

WhitWaldo:
Following up with a note here that this issue will not be completed in the short term. Rather, it'll be part of a larger effort to migrate out all the existing functionality to standalone packages by purpose. Those packages themselves will be nullability annotated from the start, conform to best practices and start from a place of rigorous unit and integration testing as part of the transition.

Issue #1004, opened 2023-01-04
Review and update NuGet package and assembly metadata


philliphoff:
## Describe the proposal
There are a number of instances, in the SDK's NuGet package and assembly metadata, that still refer to Microsoft as authors or copyright holder. With Dapr managed by the foundation, that metadata should be reviewed and updated, if necessary.

See the common assembly/package [copyright](https://github.com/dapr/dotnet-sdk/blob/ab5403ed015d7ba46099cf829186e2ea7d6264a5/properties/dapr_managed_netcore.props#L44) for an example.

PR #1320, opened 2024-07-06
Added .NET client for new Dapr Jobs API support

WhitWaldo:
# Description

Added a Dapr Jobs client for the new [Jobs API](https://v1-14.docs.dapr.io/reference/api/jobs_api/#schedule-a-job)

## Issue reference

Please reference the issue this PR will close: #1321 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests

This isn't yet ready to be merged as I need to figure out how the sidecar signals job triggers back to the app.


WhitWaldo:
@philliphoff If you could prioritize review of this PR so it can make it into 1.15, that'd be great. I only just learned of this new API this week, so it's a bit last-minute, but reflects all the available documentation on the [preview docs](https://v1-14.docs.dapr.io/reference/api/jobs_api/).

I'll also be working on writing some documentation in a separate PR for it this evening.

src/Dapr.Jobs/Dapr.Jobs.csproj
```diff
@@ -0,0 +1,23 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+      <TargetFrameworks>net6;net7;net8</TargetFrameworks>
```
philliphoff, line 4: Nit: only the Workflow library explicitly targets `net7` (for some historical reason). I would remove this so it gets picked up by the common build properties.  (The same goes for the version-related properties.)

src/Dapr.Jobs/DaprJobsClient.cs
```diff
@@ -0,0 +1,106 @@
...
+/// </summary>
+public class DaprJobsClient
+{
+    private readonly HttpClient httpClient;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="DaprJobsClient"/> class.
+    /// </summary>
+    /// <param name="httpClient">The <see cref="HttpClient"/> used to communicate with the Dapr sidecar.</param>
+    public DaprJobsClient(HttpClient httpClient)
```
philliphoff, line 30: We should be using gRPC to talk to the sidecar, when possible.  This brings up the question of how best to share creation of the underlying gRPC client, probably by internally exposing/sharing parts of the `DaprClientBuilder`.

src/Dapr.Jobs/JobsServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,50 @@
...
+/// <summary>
+/// Contains extension methods for using Dapr Jobs with dependency injection.
+/// </summary>
+public static class JobsServiceCollectionExtensions
+{
+    /// <summary>
+    /// Adds Dapr Jobs support to the service collection.
+    /// </summary>
+    /// <param name="serviceCollection">The <see cref="IServiceCollection"/>.</param>
+    public static IServiceCollection AddDaprJobs(this IServiceCollection serviceCollection)
```
philliphoff, line 27: `AddDaprJobsClient()`?

src/Dapr.Jobs/MalformedJobException.cs
```diff
@@ -0,0 +1,38 @@
...
+// limitations under the License.
+// ------------------------------------------------------------------------
+
+namespace Dapr.Jobs;
+
+/// <summary>
+/// The exception type thrown when an malformed request is made to the Dapr Jobs service.
+/// </summary>
+[Serializable]
+public class MalformedJobException : Exception
```
philliphoff, line 20: I would prefix these exceptions with `Dapr`.

src/Dapr.Jobs/DaprJobsClient.cs
```diff
@@ -0,0 +1,106 @@
...
+                return;
+        }
+    }
+
+    /// <summary>
+    /// Gets a job from its name.
+    /// </summary>
+    /// <param name="name">The name of the scheduled job being retrieved.</param>
+    /// <returns>The job data.</returns>
+    public async Task<string> GetJobDataAsync(string name)
```
philliphoff, line 68: The Dapr runtime returns both the job data and its related metadata (e.g. due time). The latter should be returned as well, and in a formatted way?

src/Dapr.Jobs/DaprJobsClient.cs
```diff
@@ -0,0 +1,106 @@
...
+        this.httpClient = httpClient;
+    }
+
+    /// <summary>
+    /// Schedules a job with a name.
+    /// </summary>
+    /// <param name="name">The name of the job being scheduled.</param>
+    /// <param name="jsonSerializableData">A string value providing any related content. Content is returned when the reminder expires.</param>
+    /// <param name="dueTime">Specifies the time after which this job is invoked.</param>
+    public async Task ScheduleJobAsync(string name, object jsonSerializableData, TimeSpan dueTime)
```
philliphoff, line 41: Nit: the APIs should all accept an optional `CancellationToken`.

src/Dapr.Jobs/TimeSpanExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+namespace Dapr.Jobs;
+
+/// <summary>
+/// Provides extension methods used with <see cref="TimeSpan"/>.
+/// </summary>
+internal static class TimeSpanExtensions
+{
+    /// <summary>
+    /// Creates a duration string that matches the specification at https://pkg.go.dev/time#ParseDuration per the
+    /// Jobs API specification https://v1-14.docs.dapr.io/reference/api/jobs_api/#schedule-a-job.
```
philliphoff, line 25: I find it a bit odd that the runtime didn't go with an ISO standard duration format and just used Golang's apparently arbitrary one.  I suppose it's a bit late to complain, though.

philliphoff: Oh, wait, nevermind, the original spec says that the RFC and ISO durations formats should also be supported.

src/Dapr.Jobs/DaprJobsClient.cs
WhitWaldo: Yeah, after writing this version, I was digging into the gRPC prototypes and found that what's presented there is quite different from the HTTP API (which is why I deleted the comment asking for your urgent review). I'm working on prioritizing support for the gRPC side in lieu of HTTP (much as we do with Crypto).

src/Dapr.Jobs/JobsServiceCollectionExtensions.cs
WhitWaldo: Sure

src/Dapr.Jobs/DaprJobsClient.cs
WhitWaldo: This is a quirk of the HTTP API and one I'm doing away with as I switch to the gRPC protos. They accept the metadata as a string and return a string, so my thinking was that I'd leave it to the developer to handle (de)serialization, but we have more flexibility with the gRPC side, so the latest iteration of this is generically typed.

src/Dapr.Jobs/ScheduleJobOptions.cs
```diff
@@ -0,0 +1,31 @@
...
+internal record ScheduleJobOptions([property: JsonPropertyName("job")] ScheduleJobInnerOptions JobData);
+
+/// <summary>
+/// The payload used to schedule a new job.
+/// </summary>
+/// <param name="JsonSerializableData">The JSON serializable data payload.</param>
+/// <param name="DueTime">The </param>
+internal record ScheduleJobInnerOptions(
+    [property: JsonPropertyName("data")] object JsonSerializableData, 
+    [property: JsonPropertyName("dueTime")] string DueTime);
```
philliphoff, line 31: Ideally the `DueTime` should be a `TimeSpan` (and converted into whatever format is used by Dapr).  There are also additional properties (such as recurrence) which appear to be supported in the gRPC API but not mentioneed in the HTTP-based Dapr preview docs.  We'll want to incorporate those as well.

philliphoff: The protos still appear to use a `string` instead of the common gRPC `Duration` type, which I find odd. In any case, we should attempt to *not* require users of this .NET API to understand curiosities of the Golang runtime.

philliphoff: There may be enough parameters on the gRPC interface to justify an "options" argument type rather than try to have them all as individual arguments.

WhitWaldo: I was thinking of just using a mix of overloads as jobs can be scheduled with either a Cron string or a TimeSpan. I've got the extension method to handle their golang TimeSpan expression already done and tested, but by using arguments, we could more easily guide users to using one or the other and avoiding a situation where they populate both and.. probably something undesirable happens.

philliphoff: Let's see how that ends up.  Another option would be to abstract the notion of a duration into a special type (e.g. `Duration`) and then have converters for the various formats (e.g. `Duration.FromTimeSpan(TimeSpan ts)` and `Duration.FromCronString(string cs)`).

yaron2:
Amazing to see this added!

WhitWaldo:
Trying to think through a preferred way of handling the Watch method, but this is otherwise done and nearly ready for review.

philliphoff:
@WhitWaldo Do I understand things right in that applications need to implement a gRPC "app callback" service in order to receive job events (rather than just specifying an HTTP route)?

WhitWaldo:
@philliphoff That's what it looks like to me - it appears to expect that there's some long-running service that subscribes to receive notifications (perhaps partitioned by app ID and namespace since that's passed along). When it receives job triggers, it's supposed to invoke them, then return an ID so it can be marked as completed.

Going in, I had assumed this would just repurpose pubsub in some manner, but that doesn't appear to be the case.

yaron2:
> @WhitWaldo Do I understand things right in that applications need to implement a gRPC "app callback" service in order to receive job events (rather than just specifying an HTTP route to be invoked a la pub-sub)?

HTTP triggering of apps is also supported in addition to a gRPC appcallback

WhitWaldo:
> > @WhitWaldo Do I understand things right in that applications need to implement a gRPC "app callback" service in order to receive job events (rather than just specifying an HTTP route to be invoked a la pub-sub)?
> 
> HTTP triggering of apps is also supported in addition to a gRPC appcallback

Can you point me to any interface that suggests how to do this? I don't see this on any of the protos.

yaron2:
> > > @WhitWaldo Do I understand things right in that applications need to implement a gRPC "app callback" service in order to receive job events (rather than just specifying an HTTP route to be invoked a la pub-sub)?
> > 
> > 
> > HTTP triggering of apps is also supported in addition to a gRPC appcallback
> 
> Can you point me to any interface that suggests how to do this? I don't see this on any of the protos.
> 
> The [job definition](https://github.com/dapr/dapr/blob/7ea7016c0150e6a7b7bc7c30e34d56f0001c5e90/dapr/proto/scheduler/v1/scheduler.proto#L21) only provides for the schedule and payload.
> 
> The [job metadata](https://github.com/dapr/dapr/blob/7ea7016c0150e6a7b7bc7c30e34d56f0001c5e90/dapr/proto/scheduler/v1/scheduler.proto#L63) suggests that it's just identifying the requestor app based on the comments.
> 
> The [schedule job request](https://github.com/dapr/dapr/blob/7ea7016c0150e6a7b7bc7c30e34d56f0001c5e90/dapr/proto/scheduler/v1/scheduler.proto#L118) only includes those two properties and a job name.

It's not part of the protos because these are the interfaces for gRPC. We are updating the docs tomorrow to add the entire specification for HTTP.

WhitWaldo:
> > > > @WhitWaldo Do I understand things right in that applications need to implement a gRPC "app callback" service in order to receive job events (rather than just specifying an HTTP route to be invoked a la pub-sub)?
> > > 
> > > 
> > > HTTP triggering of apps is also supported in addition to a gRPC appcallback
> > 
> > 
> > Can you point me to any interface that suggests how to do this? I don't see this on any of the protos.
> > The [job definition](https://github.com/dapr/dapr/blob/7ea7016c0150e6a7b7bc7c30e34d56f0001c5e90/dapr/proto/scheduler/v1/scheduler.proto#L21) only provides for the schedule and payload.
> > The [job metadata](https://github.com/dapr/dapr/blob/7ea7016c0150e6a7b7bc7c30e34d56f0001c5e90/dapr/proto/scheduler/v1/scheduler.proto#L63) suggests that it's just identifying the requestor app based on the comments.
> > The [schedule job request](https://github.com/dapr/dapr/blob/7ea7016c0150e6a7b7bc7c30e34d56f0001c5e90/dapr/proto/scheduler/v1/scheduler.proto#L118) only includes those two properties and a job name.
> 
> It's not part of the protos because these are the interfaces for gRPC. We are updating the docs tomorrow to add the entire specification for HTTP.

Could you speak at all to how the app callback is envisioned to work on the client?

yaron2:
> > > > > @WhitWaldo Do I understand things right in that applications need to implement a gRPC "app callback" service in order to receive job events (rather than just specifying an HTTP route to be invoked a la pub-sub)?
> > > > 
> > > > 
> > > > HTTP triggering of apps is also supported in addition to a gRPC appcallback
> > > 
> > > 
> > > Can you point me to any interface that suggests how to do this? I don't see this on any of the protos.
> > > The [job definition](https://github.com/dapr/dapr/blob/7ea7016c0150e6a7b7bc7c30e34d56f0001c5e90/dapr/proto/scheduler/v1/scheduler.proto#L21) only provides for the schedule and payload.
> > > The [job metadata](https://github.com/dapr/dapr/blob/7ea7016c0150e6a7b7bc7c30e34d56f0001c5e90/dapr/proto/scheduler/v1/scheduler.proto#L63) suggests that it's just identifying the requestor app based on the comments.
> > > The [schedule job request](https://github.com/dapr/dapr/blob/7ea7016c0150e6a7b7bc7c30e34d56f0001c5e90/dapr/proto/scheduler/v1/scheduler.proto#L118) only includes those two properties and a job name.
> > 
> > 
> > It's not part of the protos because these are the interfaces for gRPC. We are updating the docs tomorrow to add the entire specification for HTTP.
> 
> Could you speak at all to how the app callback is envisioned to work on the client?

Yes, the same as pub/sub with the user registering an endpoint in the format of `/job/{job-name}` to receive the callback.

You can see a sample in our test code here for the endpoint: https://github.com/dapr/dapr/blob/7ea7016c0150e6a7b7bc7c30e34d56f0001c5e90/tests/apps/schedulerapp/app.go#L189

And here is the handler of the method: https://github.com/dapr/dapr/blob/7ea7016c0150e6a7b7bc7c30e34d56f0001c5e90/tests/apps/schedulerapp/app.go#L154

WhitWaldo:
@yaron2 Could you ping me when the docs are live?

One more question - am I right in reading that only the requestor of the job can receive the job trigger notification? Or can any service opt into receiving all the jobs (presumably with some sort of namespace filtering on the server)?

yaron2:
> @yaron2 Could you ping me when the docs are live?
> 
> One more question - am I right in reading that only the requestor of the job can receive the job trigger notification? Or can any service opt into receiving all the jobs (presumably with some sort of namespace filtering on the server)?

I'll ping you, yes 

Re: question, only the requestor of the job will get the callback.

WhitWaldo:
Pushed my latest local commits:
- The previous iteration looked at using just the HTTP endpoints and that's been completed scrubbed in favor of gRPC
- I refactored the DaprClientBuilder functionality out of Dapr.Client and into a `Dapr.Common` shared project as an abstract class. Each of the Dapr client projects going forward would have its own proto-generated client that is built atop this builder (keeping with the idea that we expose the `Dapr*Client` publicly but implement as an internal `Dapr*GrpcClient`. 
- In the same vein, I've moved `DaprException` into this shared project, but didn't change the namespace so it doesn't break existing references for anyone.
- I've updated the TimeSpan extension method to include a `ToTimeSpan` to go with the `FromTimeSpan` method so Golang's interval string can be bidirectionally parsed (and added tests)
- I've added extension support for registering the DaprJobsClient in DI (though I'll tweak this once the HTTP docs are updated so we can register callback endpoints as well)
- `RegisterJob`, `GetJob` and `DeleteJob` are fully implemented via gRPC (though register might need tweaking for HTTP invocations). Register supports each of the scheduling approaches suggested and the various overloads do the necessarily validation of each cross-combination of options.
- Applied each of the suggestions from the previous code review (exceptions are prefixed with "Dapr", extension method uses "AddDaprJobsClient" instead of just "AddDaprJobs", everything includes support for cancellation tokens, all internal bits are abstracted away by .NET types on the SDK itself).
- The `WatchJobs` implementation is a work in progress and is presently stalled as I figure out how this was intended to be realized independent of an HTTP invocation. Again, the biggest hold up here is the following scenario: Developer sets up endpoints to handle `jobA` and `jobB` ahead of time. During execution, the developer would like to set up `jobC` and calls `RegisterJob` accordingly. Presumably, the developer would now like to provide a mechanism by which `jobC` should be handled, but it seems like poor DevEx to simply say "sorry, recompile and redeploy with a static endpoint for newly registered jobs". That works in PubSub because the publisher and the various topics exist independent of whatever is pushing the data, but here, it's all created in Dapr so it feels like there should be some sort of on-the-fly handling mechanism too.
-

philliphoff:
@WhitWaldo Wait, I'm thoroughly confused now. WatchJobs() is in the scheduler protos, right? Isn't that API used only between the sidecar and the new scheduler service? I thought that the app would use the new alpha methods on the dapr runtime protos which doesn't have a streaming "watch" equivalent (that I can see). Perhaps I'm just missing something fundamental.

WhitWaldo:
@philliphoff To be honest, I didn't even see the Scheduler methods on the runtime proto. They look to be the same shape as those on the scheduler proto. I implemented against the scheduler proto, so sure, could just as easily drop the watch method. But there's no dependency from this project on Dapr.Client, including the runtime proto as a result.

Still don't see anywhere to register an HTTP invocation even peeking at the runtime proto, so I'll have to sit tight on what that looks like.

yaron2:
The APIs that SDKs use to invoke Dapr are in this proto: https://github.com/dapr/dapr/blob/master/dapr/proto/runtime/v1/dapr.proto

And for Dapr calling the app via gRPC, it's this: https://github.com/dapr/dapr/blob/master/dapr/proto/runtime/v1/appcallback.proto

The protos don't register HTTP operations - like pub/sub, bindings and other APIs that call into the app, receiving events is decoupled from sending events. An app can, for example, create a job via the protos and receive it with it's choice of protocol: if the app is configured to run in HTTP,  Dapr will deliver the message on an HTTP endpoint as I mentioned above. If the app chose to receive via gRPC, it will use the AppCallback proto. This is the same and consistent with pub/sub, bindings and service invocation.

WhitWaldo:
> The APIs that SDKs use to invoke Dapr are in this proto: https://github.com/dapr/dapr/blob/master/dapr/proto/runtime/v1/dapr.proto
> 
> And for Dapr calling the app via gRPC, it's this: https://github.com/dapr/dapr/blob/master/dapr/proto/runtime/v1/appcallback.proto
> 
> The protos don't register HTTP operations - like pub/sub, bindings and other APIs that call into the app, receiving events is decoupled from sending events. An app can, for example, create a job via the protos and receive it with it's choice of protocol: if the app is configured to run in HTTP, Dapr will deliver the message on an HTTP endpoint as I mentioned above. If the app chose to receive via gRPC, it will use the AppCallback proto. This is the same and consistent with pub/sub, bindings and service invocation.

This all makes a lot more sense now that I realize I was looking at entirely the wrong protos. Latest commit fixes all this to point to the correct runtime protos and fix a bunch of the references. Working on adding the callback support right now.

src/Dapr.Jobs/Protos/dapr/proto/common/v1/common.proto
philliphoff, line 1: One approach we could take is to put the `.proto` files into a separate package that is then referenced by the packages that do gRPC operations, to prevent each package from having a separate copy and managing the gRPC type generation themselves.

src/Dapr.Jobs/DaprJobsClient.cs
```diff
@@ -0,0 +1,97 @@
...
+    /// </summary>
+    /// <param name="jobName">The name of the job being scheduled.</param>
+    /// <param name="cronExpression">The systemd Cron-like expression indicating when the job should be triggered.</param>
+    /// <param name="dueTime">The optional point-in-time from which the job schedule should start.</param>
+    /// <param name="repeats">The optional number of times the job should be triggered.</param>
+    /// <param name="ttl">Represents when the job should expire. If both this and DueTime are set, TTL needs to represent a later point in time.</param>
+    /// <param name="payload">The main payload of the job.</param>
+    /// <param name="cancellationToken">Cancellation token.</param>
+    [Obsolete("The API is currently not stable as it is in the Alpha stage. This attribute will be removed once it is stable.")]
+    public abstract Task ScheduleJobAsync<T>(string jobName, string cronExpression, DateTime? dueTime,
```
philliphoff, line 41: This feels like too many arguments to me; .NET guidelines generally suggest keeping it to 3-4.  In particular, because many are optional, it becomes awkward to deal with the cancellation token.  (I think the standard practice of making the cancellation token the last parameter was a mistake, personally.  .NET should have made it a required first parameter or figured out a way to make it implicit.  I understand why they did it, though, but now we're stuck with it as well as using `Async` as the suffix everywhere.)

philliphoff:
> Working on adding the callback support right now.

@WhitWaldo I think the big question is which model to use:

 - Have users create a route and then decorate it with the job it's intended to be used by (pub-sub style)
 - Have users register a handler for a job and then the SDK generates an appropriate route which calls the handler (actor style)
 - Both?

src/Dapr.Jobs/DaprJobsClient.cs
WhitWaldo: I just added three more overloads to avoid the generic typing when the user doesn't want to submit job data with the request - so this just got worse since you looked at it. 

There are three routes forward from here:
1) Rename the three scheduling approaches (e.g. `ScheduleOneTimeJobAsync`, `ScheduleIntervalJobAsync` or `ScheduleCronJobAsync` (unless you can think of better names) and deal with having all the parameters in each method
2) Put these into an options class (but that just shifts the problem to the options constructor)
3) Build more of a fluent interface so users can opt into the options they want to provide (but should the API change, this might need an overhaul and it just shifts all the optional values into additional chained method calls).

WhitWaldo:
@philliphoff 

> > Working on adding the callback support right now.
> 
> @WhitWaldo I think the big question is which model to use:
> 
>     * Have users create a route and then decorate it with the job it's intended to be used by (pub-sub style)
> 
>     * Have users register a handler for a job and then the SDK generates an appropriate route which calls the handler (actor style)
> 
>     * Both?

I'm building a sample for the gRPC approach where the user just adds a class that implements `AppCallback.AppCallbackBase` and adds an override for `OnJobEventAlpha1`, but that's a very prescribed approach that I feel the SDK could improve on.

I'm inclined to favor the attribute approach and just put a `[Job("jobname")]` on each method with documentation ensuring that the user sets the first argument as whatever type they're deserializing the job payload to (though I'd like to figure out how ASP.NET Core does this in their minimal API request handler delegates so the arguments could be passed in any order).

Thoughts?

src/Dapr.Jobs/TimeSpanExtensions.cs
WhitWaldo: Currently, I've got the SDK exposing `DateTime`s and converting to the ISO 8601 string for the backend

src/Dapr.Jobs/ScheduleJobOptions.cs
WhitWaldo: This type has been removed and we've got the discussion continuing below about this, so marking this part as resolved for now.

philliphoff:
> I'm building a sample for the gRPC approach where the user just adds a class that implements AppCallback.AppCallbackBase and adds an override for OnJobEventAlpha1, but that's a very prescribed approach that I feel the SDK could improve on.

@WhitWaldo While I think we'd eventually want this, I *think* it requires apps to go all-in on the gRPC for everything.  I don't believe an app can, say, use the gRPC callback model for jobs but use the HTTP route approach for pub-sub in the same app.  There's also the issue of only one service can be registered of that  gRPC service type, but that type lumps all of the callbacks together (method invocation, state store, jobs, etc.) which means there needs to be some way to "share" the service with potentially other Dapr packages (e.g. the original Dapr.AspDotNet, any new Dapr.PubSub, etc.).  It's possible (we could use similar ideas to what we did for pluggable components) but probably requires splitting it out into a separate common package.

> I'm inclined to favor the attribute approach and just put a [Job("jobname")] on each method with documentation ensuring that the user sets the first argument as whatever type they're deserializing the job payload to (though I'd like to figure out how ASP.NET Core does this in their minimal API request handler delegates so the arguments could be passed in any order).

@WhitWaldo Thinking a bit more, given that it looks like job handlers are on fixed routes like method invocation (vs. pub-sub, where the subscription includes the route), attribute-based mapping may not work.  If that's the case, then a callback-based approach where the route is auto-generated behind the scenes may be better.

WhitWaldo:
> > I'm building a sample for the gRPC approach where the user just adds a class that implements AppCallback.AppCallbackBase and adds an override for OnJobEventAlpha1, but that's a very prescribed approach that I feel the SDK could improve on.
> 
> @WhitWaldo While I think we'd eventually want this, I _think_ it requires apps to go all-in on the gRPC for everything. I don't believe an app can, say, use the gRPC callback model for jobs but use the HTTP route approach for pub-sub in the same app. There's also the issue of only one service can be registered of that gRPC service type, but that type lumps all of the callbacks together (method invocation, state store, jobs, etc.) which means there needs to be some way to "share" the service with potentially other Dapr packages (e.g. the original Dapr.AspDotNet, any new Dapr.PubSub, etc.). It's possible (we could use similar ideas to what we did for pluggable components) but probably requires splitting it out into a separate common package.

@philliphoff  Ah, I don't know anything about this. Definitely worth digging into as that would be an excellent reason to both not expose/document that today until that can be resolved. I'm not super familiar with all things gRPC and how it works with .NET, so that's worth some homework in the future to figure out a better approach.

> 
> > I'm inclined to favor the attribute approach and just put a [Job("jobname")] on each method with documentation ensuring that the user sets the first argument as whatever type they're deserializing the job payload to (though I'd like to figure out how ASP.NET Core does this in their minimal API request handler delegates so the arguments could be passed in any order).
> 
> @WhitWaldo Thinking a bit more, given that it looks like job handlers are on fixed routes like method invocation (vs. pub-sub, where the subscription includes the route), attribute-based mapping may not work. If that's the case, then a callback-based approach where the route is auto-generated behind the scenes may be better.

I was considering doing a reflection-based registration at startup to find all any methods decorated with `[Job]` and map the names, then invoke accordingly when prompted. I'm also not super-familiar with how Actors presently does this either, so I'll similarly have to dig into that.

Unrelated - I don't know why the DCO check is failing. I double checked all the commits just a moment ago and they all show the DCO message.

philliphoff:
> I was considering doing a reflection-based registration at startup to find all any methods decorated with [Job] and map the names, then invoke accordingly when prompted. I'm also not super-familiar with how Actors presently does this either, so I'll similarly have to dig into that.

@WhitWaldo I'd avoid reflection if at all possible--it's role has pretty much been supplanted by generators that can do the equivalent but at compile-time instead, which is important both for performance as well as being friendly to AOT.

I think the catch is that the job route is either statically mapped for each job name, in which case an attribute seems a bit superfluous, or is dynamically mapped such that it handles all possible jobs and would then have to have multiple attributes, which feels a bit weird.

Because of the expectation of job handlers having static routes, it may just work best to dynamically generate them under the covers.  If and when the jobs feature evolves to allow a true mapping between job name and handler route, then it might be easier to add attributes that enable such maps.

> Unrelated - I don't know why the DCO check is failing. I double checked all the commits just a moment ago and they all show the DCO message.

@WhitWaldo I dislike the whole DCO concept. You may just have to rebase/squash commits, amend with signature, and then force push it back to resolve it.

WhitWaldo:
> > I was considering doing a reflection-based registration at startup to find all any methods decorated with [Job] and map the names, then invoke accordingly when prompted. I'm also not super-familiar with how Actors presently does this either, so I'll similarly have to dig into that.
> 
> @WhitWaldo I'd avoid reflection if at all possible--it's role has pretty much been supplanted by generators that can do the equivalent but at compile-time instead, which is important both for performance as well as being friendly to AOT.
> 
> I think the catch is that the job route is either statically mapped for each job name, in which case an attribute seems a bit superfluous, or is dynamically mapped such that it handles all possible jobs and would then have to have multiple attributes, which feels a bit weird.
> 
> Because of the expectation of job handlers having static routes, it may just work best to dynamically generate them under the covers. If and when the jobs feature evolves to allow a true mapping between job name and handler route, then it might be easier to add attributes that enable such maps.

This makes sense and I'm not opposed to it. I've got a _lot_ of projects that use Metalama for compile-time generation - is there any opposition to using an [open source license](https://www.postsharp.net/metalama/pricing/open-source) for them or do you know if there's a preference to use the dramatically less flexible Roslyn source generators?

> 
> > Unrelated - I don't know why the DCO check is failing. I double checked all the commits just a moment ago and they all show the DCO message.
> 
> @WhitWaldo I dislike the whole DCO concept. You may just have to rebase/squash commits, amend with signature, and then force push it back to resolve it.

I agree - I'd much rather see a single-sign-off process like the one necessary to commit to any of Microsoft's OSS packages and not have to deal with adding the sign-off text (especially since I can't get VS to consistently apply it, so I have to resort to manually typing it with each commit).

I'll wait until we're happy with everything on this thread and then close it, squash it all and re-open the PR ready for a final review. If it's unhappy now with any of the DCO messages despite not seeing what it doesn't care for, I suspect it'll be unhappy about the several upcoming commits as well.

WhitWaldo:
@philliphoff I mentioned this on the other thread, but I thought I'd repeat it here for posterity.

What are you thoughts on dropping the serialization altogether here and just accepting a `byte[]` for the job payload? It puts the responsibility back on the developer to handle serialization however they want and means that we avoid the inevitable bug report of someone experiencing a runtime error because they've specified a different type to deserialize fom than they used whenever they created the job. By instead only accepting a `byte[]`, they can use whatever serialization they desire on their side and gets the SDK out of having to support pluggable serialization.

It also means here that we can drop the second overload for each method and just have a single method for each kind of schedule, each with their different (and mostly optional) arguments.

philliphoff:
> What are you thoughts on dropping the serialization altogether here and just accepting a byte[] for the job payload? It puts the responsibility back on the developer to handle serialization however they want and means that we avoid the inevitable bug report of someone experiencing a runtime error because they've specified a different type to deserialize fom than they used whenever they created the job. By instead only accepting a byte[], they can use whatever serialization they desire on their side and gets the SDK out of having to support pluggable serialization.

We could take the .NET runtime approach and have the client offer a "plain" method that accepts the raw input (probably better to be `ReadOnlyMemory<byte>` rather than `byte[]`) and then offer extension methods that wrap the method with JSON-specific flavors (see `HttpClient.PostAsync()` and `HttpClientJsonExtensions.PostAsJsonAsync<TValue>()`).  Those extension methods could have an additional overload that accepts, for example, a `JsonSerializerOptions` instance to configure things, which then offers very similar behavior to today's Dapr client class but without embedding all of the serialization logic in the "main" methods.  While we still technically implement the serialization, it's more directly configurable by the user.

WhitWaldo:
> > What are you thoughts on dropping the serialization altogether here and just accepting a byte[] for the job payload? It puts the responsibility back on the developer to handle serialization however they want and means that we avoid the inevitable bug report of someone experiencing a runtime error because they've specified a different type to deserialize fom than they used whenever they created the job. By instead only accepting a byte[], they can use whatever serialization they desire on their side and gets the SDK out of having to support pluggable serialization.
> 
> @WhitWaldo We could take the .NET runtime approach and have the client offer a "plain" method that accepts the raw input (probably better to be `ReadOnlyMemory<byte>` rather than `byte[]`) and then offer extension methods that wrap the method with JSON-specific flavors (see `HttpClient.PostAsync()` and `HttpClientJsonExtensions.PostAsJsonAsync<TValue>()`). Those extension methods could have an additional overload that accepts, for example, a `JsonSerializerOptions` instance to configure things, which then offers very similar behavior to today's Dapr client class but without embedding all of the serialization logic in the "main" methods. While we still technically implement the serialization, it's more directly configurable by the user.

Now that's a neat idea - it leaves the door open to have a collection of supported extensions and if there's a third-party option that wants to complement/supplement with more, there's no reason they can't augment the namespace with more from their own package.

Agreed on `ReadOnlyMemory<byte>` I thought about that while typing it out, but couldn't immediately remember what it was about it that didn't work on the Crypto side and didn't want to immediately go look it up.

src/Dapr.Jobs/Protos/dapr/proto/common/v1/common.proto
WhitWaldo: @philliphoff I found some thoughts on SO about doing this [here](https://stackoverflow.com/questions/59422186/importing-proto-files-from-another-project) and gave it a whirl in the latest implementation on this PR and.. it failed spectacularly. While VS sees the linked files and properly maps them, the Dapr projects wouldn't build because they couldn't resolve the generated C# references despite VS finding them.

This would be a great path forward and an excellent purpose for the `Dapr.Common` project, but it's going to need more work.

WhitWaldo:
@yaron2 Another question for you while waiting on the docs (though I'd be surprised if this made it into the first draft): The `Job` [proto](https://github.com/dapr/dapr/blob/7ea7016c0150e6a7b7bc7c30e34d56f0001c5e90/dapr/proto/runtime/v1/dapr.proto#L1182) uses a `google.protobuf.Any` for the data property:
```proto
message Job {
  // The unique name for the job.
  string name = 1;

  // The schedule for the job.
  optional string schedule = 2;

  // Optional: jobs with fixed repeat counts (accounting for Actor Reminders).
  optional uint32 repeats = 3;

  // Optional: sets time at which or time interval before the callback is invoked for the first time.
  optional string due_time = 4;

  // Optional: Time To Live to allow for auto deletes (accounting for Actor Reminders).
  optional string ttl = 5;

  // Job data.
  google.protobuf.Any data = 6;
}
```

Any is documented [here](https://github.com/protocolbuffers/protobuf/blob/06e65a29de31efc65d4ab4e07b19c1636fc45d9c/src/google/protobuf/any.proto#L128). The `Any` definition cites a string for `type_url` and a bytes for `value`:
```proto
message Any {
  string type_url = 1;
  bytes value = 2;
}
```

But in the Go example [you cited](https://github.com/dapr/dapr/blob/7ea7016c0150e6a7b7bc7c30e34d56f0001c5e90/tests/apps/schedulerapp/app.go#L46C1-L57C2), it indicates that the data property is comprised of a string for `type` and a string for `value`:
```go
type jobData struct {
	DataType   string `json:"@type"`
	Expression string `json:"expression"`
}

type job struct {
	Data     jobData `json:"data,omitempty"`
	Schedule string  `json:"schedule,omitempty"`
	Repeats  int     `json:"repeats,omitempty"`
	DueTime  string  `json:"dueTime,omitempty"`
	TTL      string  `json:"ttl,omitempty"`
}
```

If I'm converting a byte array in C# to an Any from the generated proto file and registering that in Dapr when I schedule the job, what are you doing to it such that it's returning a string (so I might reverse it and get my byte[] back)?

I appreciate it!

WhitWaldo:
@philliphoff I started writing a code generator for the handler registration and then spent the morning reading up on what the ASP.NET Core team is doing with regards for optimizing minimal APIs and their delegate handlers. Put simply, they wanted to use source generators there so they can achieve better compatibility with AOT than they can with delegates (hidden behind an opt-in flag: `<EnableRequestDelegateGenerator>true</EnableRequestDelegateGenerator>` in a `PropertyGroup`).

But they're also looking at using the interceptors introduced with C#12 to augment this further and I suppose replace each of the overloads with a specific method for whatever the mapping should be for even better performance (I think you get this if you just opt into AOT regardless of the other flag since it's also considered preview).

Anyway, given that this effectively gives us the benefit of their work to have already implemented the source generators, I intended just opted to handle registration via a minimal API (for now, unless we need others) so that path registration is as easy as:

```cs
app.MapScheduledJob("myJob", (ILogger logger, JobDetails details) =>
{
  // Do something
});
```

The delegate supports dependency injection as well as model binding and the first argument would let them specify the name of the job they're targeting. This extension then just changes the path to "job/{jobName}" and registers it on the `IEndpointRouteBuilder` not dramatically unlike how it's done in `Dapr.Actors.AspNetCore`.

I added a sample project and locally installed the preview bits, but haven't been able to get it to run to trial this out. Going to try rebooting later and see if that helps.

I think it'd be nice to add some helper extension methods to convert types into the `ReadOnlyMemory<byte>` (perhaps just strings and types that can be serialized into JSON) and back, but otherwise, I think this is just about complete once I can get an E2E test working and add some more unit testing outside of the extensions.

philliphoff:
> Anyway, given that this effectively gives us the benefit of their work to have already implemented the source generators, I intended just opted to handle registration via a minimal API (for now, unless we need others) so that path registration is as easy as

@WhitWaldo I think this is a good place to start, given the static nature of job routes, and it minimizes the need for users to understand the internals of Dapr job dispatching.

src/Dapr.Jobs/Extensions/EndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    /// <summary>
+    /// Adds a <see cref="RouteEndpoint"/> to the <see cref="IEndpointConventionBuilder"/> that registers a
+    /// scheduled job trigger invocation.
+    /// </summary>
+    /// <param name="endpoints">The <see cref="IEndpointRouteBuilder"/> to add the route to.</param>
+    /// <param name="jobName">The name of the job that should trigger this method when invoked.</param>
+    /// <param name="handler">The delegate executed when the endpoint is matched.</param>
+    /// <returns></returns>
+    public static IEndpointConventionBuilder MapScheduledJob(this IEndpointRouteBuilder endpoints, string jobName,
```
philliphoff, line 19: Might suggest `MapDaprScheduledJob()`.

WhitWaldo:
Oh good, all the tests are passing again. Glad to see that! I'll see if I can't hammer out some unit tests to go with and get this to run locally so I can trial out that sample project too and get this wrapped up.

WhitWaldo:
> > What are you thoughts on dropping the serialization altogether here and just accepting a byte[] for the job payload? It puts the responsibility back on the developer to handle serialization however they want and means that we avoid the inevitable bug report of someone experiencing a runtime error because they've specified a different type to deserialize fom than they used whenever they created the job. By instead only accepting a byte[], they can use whatever serialization they desire on their side and gets the SDK out of having to support pluggable serialization.
> 
> @WhitWaldo We could take the .NET runtime approach and have the client offer a "plain" method that accepts the raw input (probably better to be `ReadOnlyMemory<byte>` rather than `byte[]`) and then offer extension methods that wrap the method with JSON-specific flavors (see `HttpClient.PostAsync()` and `HttpClientJsonExtensions.PostAsJsonAsync<TValue>()`). Those extension methods could have an additional overload that accepts, for example, a `JsonSerializerOptions` instance to configure things, which then offers very similar behavior to today's Dapr client class but without embedding all of the serialization logic in the "main" methods. While we still technically implement the serialization, it's more directly configurable by the user.

@philliphoff In the latest batch of commits, I added a collection of extension methods that'll support JSON and string serialization (with a `JsonSerializationOptions` passed in as an optional argument) and deserialization along with unit tests for the latter set.

As I discovered after a few failing unit tests, Moq doesn't work with static methods, which makes this project rather difficult to get full coverage on. Is there any interest at some point in seeing if there's a commercial mocking framework that would grant a license for this open-source project (e.g. Telerik's JustMock)? I've got to imagine there are a great many opportunities for richer coverage we just can't test with open-source mocking tooling.

Finally, I added a change to the generic Dapr client builder that took a dependency on `Microsoft.Extensions.Http` so we could source the `HttpClient` from its `IHttpClientFactory.CreateClient` method instead of creating one every time the builder is used and it'd be registered in DI alongside anything else. Unfortunately, this package takes a hard dependency on the 8.0.* logging which broke Dapr.Extensions.Configuration because of its chain of references and its existing reference to 3.1.2 for `Microsoft.Extensions.Configuration`. Updating to 8.0.* on the configuration project introduces a nullable incompatibility minefield, so I've instead I've fixed it for now by reverting the `Microsoft.Extensions.Http` to 3.1.32 (and upgrading the configuration project to the same). All that to say that when we drop .NET 6, it'd be a great time to both review #1004 and keep working through #1316 because with package updates, we're likely going to introduce more nullability work.

WhitWaldo:
Unclear why the build is failing on CI/CD - it's building all the way through locally.

WhitWaldo:
@philliphoff Closing this PR in favor of #1331 now that all the unit tests are passing. Did a squash merge per your suggestion to fix the DCO error.

Issue #1321, opened 2024-07-06
Jobs API needs a .NET client


WhitWaldo:
## Describe the feature
Since there's a new Jobs API showing up in Dapr in the near future, it should receive support in the .NET client.

As I'm actively advocating for the split out of Dapr client functionality into separate assemblies so that developers can explicitly opt-into those pieces of Dapr they're looking to use and so we can avoid some method naming conflicts (especially once I start hammering out the various specialty state APIs), I thought this would be an excellent place to start as it's quite a lightweight API footprint so far.

## Release Note
RELEASE NOTE: **ADD** Support for Dapr Jobs API

PR #1331, opened 2024-07-25
Added .NET client for Dapr Jobs API

WhitWaldo:
# Description

Added a Dapr Jobs client for the new [Jobs API](https://v1-14.docs.dapr.io/reference/api/jobs_api/#schedule-a-job).

This PR is intended to replace the previous one and was performed as a squash merge of it - the third commit (early on) was missing a period in my email address which apparently breaks the DCO process (even though I have an email alias and I can still receive email there). Because I cannot figure out how to fix multi-line comments, I'm providing this updated PR instead.

## Issue reference

Please reference the issue this PR will close: #1321 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Added sample project


WhitWaldo:
@philliphoff Any other changes you think should be made here?

WhitWaldo:
I'm seeing the E2E tests fail on my system following the recent merge with `master`. In hopes those are due to something wrong with those tests, I'll hold off on any other changes to this PR to address them until it's confirmed it's a problem with this PR and not them.

I did add two more unit tests to validate the DaprClientBuilder a bit more where it was lacking existing tests and those are passing.

yaron2:
cc @halspang @philliphoff

src/Dapr.Jobs/DaprJobsServiceException.cs
```diff
@@ -0,0 +1,38 @@
...
+    public DaprJobsServiceException(HttpResponseMessage? response) : base(FormatExceptionForFailedRequest())
+    {
+        Response = response;
+    }
+    
+    /// <summary>
+    /// Gets the <see cref="HttpResponseMessage"/> of the request that failed. Will be <c>null</c> if the
+    /// failure was not related to an HTTP request or preventing the response from being received.
+    /// </summary>
+    public HttpResponseMessage? Response { get; }
```
philliphoff, line 35: I don't know how I feel about exposing the `HttpResponseMessage`. While it *is* an error and therefore something of a special case, it would seem to leak a bit of the abstraction of the API we're building.  Would it be better for the exception to require a message that more specifically describes what about the HTTP response it didn't like?

philliphoff: Same comment for the `DaprMalformedJobException`, too.

src/Dapr.Jobs/Dapr.Jobs.csproj
```diff
@@ -0,0 +1,40 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <TargetFrameworks>net6;net8</TargetFrameworks>
+        <ImplicitUsings>enable</ImplicitUsings>
+        <Nullable>enable</Nullable>
+        <PackageId>Dapr.Jobs</PackageId>
+        <Title>Dapr Jobs Authoring SDK</Title>
+        <Description>Dapr Jobs SDK for scheduling jobs and tasks with Dapr</Description>
+        <VersionPrefix>0.1.0</VersionPrefix>
```
philliphoff, line 10: Would we want to version the "sub" SDKs independently?  While I think it's *technically* the right thing to do, as the number of NuGet packages increase, version management becomes more of a challenge and there's an increase risk of customer confusion (e.g. why do I reference 1.14 of the Dapr.Client library but 0.1.0 of the Dapr.Jobs library).  Traditionally we've followed the version of the Dapr runtime, at least for major/minor versions, and the build automation infrastructure assumes packages are all published as a set.  We could change all that, of course, but I'm not sure if there's significant benefit.

src/Dapr.Common/DaprGenericClientBuilder.cs
```diff
@@ -0,0 +1,226 @@
...
+        };
+
+        this.JsonSerializerOptions = new JsonSerializerOptions(JsonSerializerDefaults.Web);
+        this.DaprApiToken = DaprDefaults.GetDefaultDaprApiToken();
+    }
+
+    /// <summary>
+    /// Property exposed for testing purposes.
+    /// </summary>
+    public string GrpcEndpoint { get; private set; }
```
philliphoff, line 48: With `InternalsVisibleToAttribute`, could these still be `internal`?  Unless users have a specific need, we shouldn't increase the surface area of our public API.

src/Dapr.Jobs/DaprJobsClient.cs
```diff
@@ -0,0 +1,123 @@
...
+    /// </summary>
+    /// <param name="jobName">The name of the job being scheduled.</param>
+    /// <param name="cronExpression">The systemd Cron-like expression indicating when the job should be triggered.</param>
+    /// <param name="startingFrom">The optional point-in-time from which the job schedule should start.</param>
+    /// <param name="repeats">The optional number of times the job should be triggered.</param>
+    /// <param name="ttl">Represents when the job should expire. If both this and DueTime are set, TTL needs to represent a later point in time.</param>
+    /// <param name="payload">The main payload of the job.</param>
+    /// <param name="cancellationToken">Cancellation token.</param>
+    [Obsolete("The API is currently not stable as it is in the Alpha stage. This attribute will be removed once it is stable.")]
+    public abstract Task ScheduleCronJobAsync(string jobName, string cronExpression, DateTime? startingFrom = null,
```
philliphoff, line 49: I don't recall if we discussed it before, but I believe we should prefer (as per the [docs](https://learn.microsoft.com/en-us/dotnet/standard/datetime/choosing-between-datetime#the-datetimeoffset-structure)) `DateTimeOffset` to `DateTime`.

philliphoff, line 49: This would support both the `systemd` and `@`-prefixed period strings, right?  So maybe `ScheduleChronJobAsync` is not the best name?

I kind of like the idea of helping the user out with the various syntax, such as having just a single method:

```csharp
public abstract Task ScheduleJobAsync(string jobName, DaprJobSchedule schedule, ...);

public sealed class DaprJobSchedule
{
  public static DaprJobSchedule FromCronExpression(string expression) => ...;

  public static readonly DaprJobSchedule Monthly = FromCronExpression("@monthly");

  public static readonly DaprJobSchedule Weekly = FromCronExpression("@weekly");

  // ...
}
```

src/Dapr.Jobs/DaprJobsGrpcClient.cs
```diff
@@ -0,0 +1,312 @@
...
+
+    // property exposed for testing purposes
+    internal Autogenerated.Dapr.DaprClient Client => client;
+    
+    internal DaprJobsGrpcClient(
+        GrpcChannel channel,
+        Autogenerated.Dapr.DaprClient innerClient,
+        HttpClient httpClient,
+        Uri httpEndpoint,
+        JsonSerializerOptions _,
```
philliphoff, line 46: Nit: just omit the `JsonSerializerOptions` argument?

src/Dapr.Jobs/DaprJobsGrpcClient.cs
```diff
@@ -0,0 +1,312 @@
...
+            callOptions.Headers.Add(apiTokenHeader.Value.Key, apiTokenHeader.Value.Value);
+
+        return callOptions;
+    }
+
+    /// <summary>
+    /// Returns the value for the User-Agent.
+    /// </summary>
+    /// <returns>A <see cref="ProductInfoHeaderValue"/> containing the value to use for the User-Agent.</returns>
+    private static ProductInfoHeaderValue UserAgent()
```
philliphoff, line 302: Could this be generated just once and reused rather than on every call?

src/Dapr.Jobs/DaprJobsGrpcClient.cs
```diff
@@ -0,0 +1,312 @@
...
+    /// Schedules a one-time job.
+    /// </summary>
+    /// <param name="jobName">The name of the job being scheduled.</param>
+    /// <param name="scheduledTime">The point in time when the job should be run.</param>
+    /// <param name="payload">Stores the main payload of the job which is passed to the trigger function.</param>
+    /// <param name="cancellationToken">Cancellation token.</param>
+    [Obsolete("The API is currently not stable as it is in the Alpha stage. This attribute will be removed once it is stable.")]
+    public override async Task ScheduleOneTimeJobAsync(string jobName, DateTime scheduledTime, ReadOnlyMemory<byte>? payload = null,
+        CancellationToken cancellationToken = default)
+    {
```
philliphoff, line 188: I wonder if some of this duplicate logic could be consolidated to reduce risk of divergence.

src/Dapr.Jobs/Extensions/DaprJobsServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,103 @@
...
+/// </summary>
+public static class DaprJobsServiceCollectionExtensions
+{
+    /// <summary>
+    /// Adds Dapr Jobs client support to the service collection.
+    /// </summary>
+    /// <param name="serviceCollection">The <see cref="IServiceCollection"/>.</param>
+    /// <returns></returns>
+    public static IServiceCollection AddDaprJobsClient(this IServiceCollection serviceCollection)
+    {
```
philliphoff, line 30: This could just call `AddDaprJobsClient(serviceCollection, () => { })`?  Or, perhaps, just make the `configure` argument optional and eliminate this overload.

src/Dapr.Jobs/Extensions/DaprJobsServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    /// <summary>
+    /// Adds Dapr Jobs client support to the service collection.
+    /// </summary>
+    /// <param name="serviceCollection">The <see cref="IServiceCollection"/>.</param>
+    /// <param name="configure">Optionally allows greater configuration of the <see cref="DaprJobsClient"/>.</param>
+    public static IServiceCollection AddDaprJobsClient(this IServiceCollection serviceCollection, Action<DaprJobsClientBuilder>? configure)
+    {
+        ArgumentNullException.ThrowIfNull(serviceCollection, nameof(serviceCollection));
```
philliphoff, line 56: Similar to above, this is just a specialization of the overload below?

src/Dapr.Jobs/Extensions/Helpers/Deserialization/ByteArrayDeserializationExtensions.cs
```diff
@@ -0,0 +1,30 @@
...
+    public static string DeserializeToString(this byte[] byteArray) => Encoding.UTF8.GetString(byteArray);
+
+    /// <summary>
+    /// Attempts to deserialize an array of UTF-8 encoded bytes to the indicated type.
+    /// </summary>
+    /// <typeparam name="TJsonObject">The JSON-compatible type to deserialize to.</typeparam>
+    /// <param name="byteArray">The array of UTF-8 encoded bytes to deserialize.</param>
+    /// <param name="jsonSerializerOptions">Optional options to use for the <see cref="JsonSerializer"/>.</param>
+    /// <returns>The deserialized data.</returns>
+    public static TJsonObject? DeserializeFromJsonBytes<TJsonObject>(this byte[] byteArray, JsonSerializerOptions? jsonSerializerOptions = null)
```
philliphoff, line 25: `ReadOnlySpan<byte>`?

philliphoff: Here and above?

src/Dapr.Jobs/Extensions/Helpers/Serialization/DaprCronJobsSerializationExtensions.cs
```diff
@@ -0,0 +1,64 @@
...
+    /// <param name="cancellationToken">Cancellation token.</param>
+    /// <returns></returns>
+    [Obsolete("The API is currently not stable as it is in the Alpha stage. This attribute will be removed once it is stable.")]
+    public static async Task ScheduleCronJobWithPayloadAsync(this DaprJobsClient client, string jobName,
+        string cronExpression, object payload, JsonSerializerOptions? jsonSerializerOptions = null, DateTime? dueTime = null, int? repeats = null, DateTime? ttl = null,
+         CancellationToken cancellationToken = default)
+    {
+        ArgumentNullException.ThrowIfNull(payload, nameof(payload));
+
+        var serializerOptions = jsonSerializerOptions ?? new JsonSerializerOptions(JsonSerializerDefaults.Web);
```
philliphoff, line 32: Consider making the default serializer options a `static readonly` field to reduce allocations.

src/Dapr.Jobs/Extensions/Helpers/Serialization/DaprCronJobsSerializationExtensions.cs
```diff
@@ -0,0 +1,64 @@
...
+    [Obsolete("The API is currently not stable as it is in the Alpha stage. This attribute will be removed once it is stable.")]
+    public static async Task ScheduleCronJobWithPayloadAsync(this DaprJobsClient client, string jobName,
+        string cronExpression, string payload, DateTime? dueTime = null, int? repeats = null, DateTime? ttl = null,
+         CancellationToken cancellationToken = default)
+    {
+        ArgumentNullException.ThrowIfNull(payload, nameof(payload));
+
+        var payloadBytes = Encoding.UTF8.GetBytes(payload);
+
+        await client.ScheduleCronJobAsync(jobName, cronExpression, dueTime, repeats, ttl, payloadBytes,
```
philliphoff, line 61: While I know that the "scheduler" and the "handler" must ultimately agree on the types being passed, I really with Dapr APIs always included a content-type field whenever it exposed a byte/string field.  That way both have a definitive means to determine the payload type--is this a "string" string or a "JSON" string?--and offer an additional level of flexibility/dynamic-ness to their implementations.

src/Dapr.Jobs/Extensions/Helpers/Serialization/DaprOneTimeJobsSerializationExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+    /// <param name="jsonSerializerOptions">Optional options to use for the <see cref="JsonSerializer"/>.</param>
+    /// <param name="cancellationToken">Cancellation token.</param>
+    [Obsolete("The API is currently not stable as it is in the Alpha stage. This attribute will be removed once it is stable.")]
+    public static async Task ScheduleOneTimeJobWithPayloadAsync(this DaprJobsClient client, string jobName, DateTime scheduledTime,
+        object payload, JsonSerializerOptions? jsonSerializerOptions = null,
+        CancellationToken cancellationToken = default)
+    {
+        ArgumentNullException.ThrowIfNull(payload, nameof(payload));
+
+        var serializerOptions = jsonSerializerOptions ?? new JsonSerializerOptions(JsonSerializerDefaults.Web);
```
philliphoff, line 27: The serialization/deserialization logic across the extension methods could probably be consolidated into common serialize/deserialize helper methods.

src/Dapr.Jobs/Extensions/TimeSpanExtensions.cs
```diff
@@ -0,0 +1,88 @@
...
+    /// Creates a <see cref="TimeSpan"/> given a Golang duration string.
+    /// </summary>
+    /// <param name="interval">The duration string to parse.</param>
+    /// <returns>A timespan value.</returns>
+    public static TimeSpan FromDurationString(this string interval)
+    {
+        interval = interval.Replace("ms", "q");
+
+        //Define regular expressions to capture each segment
+        var hourRegex = new Regex(@"(\d+)h");
```
philliphoff, line 60: Consider replacing `new Regex()` either by using `Regex.Match()` or by a `static readonly Regex regex = new Regex("", Regex.Compiled)` field to avoid having to [re-interpret the expression](https://learn.microsoft.com/en-us/dotnet/standard/base-types/best-practices-regex#handle-object-instantiation-appropriately).

src/Dapr.Jobs/Models/Responses/JobDetails.cs
```diff
@@ -0,0 +1,89 @@
...
+    /// The string-based schedule value returned by the job details payload.
+    /// </summary>
+    [JsonPropertyName("schedule")]
+    public string? Schedule { get; init; } = null;
+
+    /// <summary>
+    /// Allows for jobs with fixed repeat counts.
+    /// </summary>
+    [JsonPropertyName("repeats")]
+    public uint? RepeatCount { get; init; } = null;
```
philliphoff, line 66: Nit: the guidelines says APIs should use `int` unless there's an explicit reason not to (such as you need the extra value space). This is despite it being the more technically correct type.  A long-standing rule, I believe, for the benefit of non-C# .NET languages that do not necessarily have the same signed vs. unsigned support.

src/Dapr.Jobs/Models/Responses/JobDetails.cs
```diff
@@ -0,0 +1,89 @@
...
+    public DateTime? DueTime { get; init; } = null;
+
+    /// <summary>
+    /// A point-in-time value representing with the job should expire.
+    /// </summary>
+    /// <remarks>
+    /// This must be greater than <see cref="DueTime"/> if both are set.
+    /// </remarks>
+    [JsonPropertyName("ttl")]
+    public DateTime? TTL { get; init; } = null;
```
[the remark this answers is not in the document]
philliphoff: Also, `DateTimeOffset`?

src/Dapr.Jobs/Models/Responses/JobDetails.cs
```diff
@@ -0,0 +1,89 @@
...
+    /// <remarks>
+    /// Either this or the <see cref="CronExpression"/> property should be specified.
+    /// </remarks>
+    public TimeSpan? Interval =>
+        Schedule is not null && isIntervalRegex.IsMatch(Schedule) ? Schedule.FromDurationString() : null;
+
+    /// <summary>
+    /// Represents whether the job is scheduled using a Cron expression.
+    /// </summary>
+    public bool IsCronExpression => CronExpression is not null;
```
philliphoff, line 49: I feel like these should be in a separate `DaprJobSchedule` type that we expose instead of the raw `schedule` property; I'm not a fan of how the schedule-related properties spread out over the job details object.

examples/Jobs/JobsSample/Program.cs
```diff
@@ -0,0 +1,41 @@
...
+
+// Configure the HTTP request pipeline.
+
+app.MapDaprScheduledJob("myJob", (ILogger logger, JobDetails details) =>
+{
+    logger.LogInformation("Received trigger invocation for 'myJob'");
+
+    if (details.Payload is not null)
+    {
+        var deserializedPayload = Encoding.UTF8.GetString(details.Payload.Value.ToArray());
```
philliphoff, line 21: We offer extension methods for serializing objects/strings when scheduling jobs; should we have equivalent ones for use in deserializing them from a handler (e.g. `details.PayloadAsJson<TJsonType>()`?

philliphoff: Oh, I see, the extensions were added to `byte[]`.  I feel like we might want to make it more job-details-specific in order to be more discoverable.  Also, we might want to avoid, for example, customers using those extension methods as generic byte-serializing/deserializing methods.

philliphoff (changes requested):
The review was a long time coming, apologies.  By and large, looks pretty good.  I have some questions about some of the APIs and I think there are some places of duplicate implementation that can be consolidated a bit.

src/Dapr.Jobs/DaprJobsClient.cs
WhitWaldo: I need to get in a better habit of doing this myself. Fixed.

src/Dapr.Common/DaprGenericClientBuilder.cs
```diff
@@ -0,0 +1,226 @@
...
+        this.Timeout = timeout;
+        return this;
+    }
+
+    /// <summary>
+    /// Builds out the inner DaprClient that provides the core shape of the
+    /// runtime gRPC client used by the consuming package.
+    /// </summary>
+    /// <exception cref="InvalidOperationException"></exception>
+    protected (GrpcChannel channel, HttpClient httpClient, Uri httpEndpoint) BuildDaprClientDependencies()
```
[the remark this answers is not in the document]
WhitWaldo: I pulled this out primarily to remove duplicate code when spinning up clients in other packages ensuring that they're all using the same configuration data for endpoints, API keys and whatnot. 

However, as was evidenced in the subscriptions implementation, this approach doesn't really lend itself to any use of injected functionality down the line (e.g. logging, compression handlers, serialization handlers, etc) provided by the developer at registration time. Do you have any thoughts on abandoning this approach altogether and instead opting for more of a DI-first, factory-style approach?

I'd be happy to put together a POC in a separate PR if you're more inclined to visit that elsewhere.

src/Dapr.Common/DaprGenericClientBuilder.cs
WhitWaldo: Yep

src/Dapr.Jobs/DaprJobsGrpcClient.cs
WhitWaldo: Done. Also removed `httpEndpoint` as it's not used anywhere either (since this just uses gRPC).

src/Dapr.Jobs/DaprJobsGrpcClient.cs
WhitWaldo: Sure thing

src/Dapr.Jobs/Extensions/DaprJobsServiceCollectionExtensions.cs
WhitWaldo: Done

src/Dapr.Jobs/Extensions/DaprJobsServiceCollectionExtensions.cs
WhitWaldo: This is more keeping with the style of the how the `DaprClient` is currently built via the ASP.NET Core package in that the configuration can reflect only the `DaprJobsClientBuilder` or also include the `IServiceProvider` introduced in #1289 in case anything else needs to be injected as part of the registration (e.g. configuration or external secrets).

src/Dapr.Jobs/Extensions/Helpers/Deserialization/ByteArrayDeserializationExtensions.cs
WhitWaldo: Done

src/Dapr.Jobs/Extensions/Helpers/Serialization/DaprCronJobsSerializationExtensions.cs
WhitWaldo: Done - also replaced each use of DateTime with DateTimeOffset in the extension methods.

src/Dapr.Jobs/Extensions/Helpers/Serialization/DaprOneTimeJobsSerializationExtensions.cs
WhitWaldo: They're close, but I don't know if they can necessarily be consolidated. One takes an object that can be presumably JSON serialized and encoded to UTF-8 bytes and the other takes what's presumably an already serialized string and encodes that to UTF-8 bytes without any further serialization. The idea for the former is that it keeps the SDK aligned with the other packages in that if you pass it serializable data, it'll just work, but the latter lets you handle serialization yourself.

I considered making the latter even just accept a byte[] so it would allow the user to handle serialization, compression and whatever else they wanted to do with the payload themselves and completely opt out of this package doing any of that lifting. Thoughts?

src/Dapr.Jobs/Extensions/TimeSpanExtensions.cs
WhitWaldo: Done

src/Dapr.Jobs/Models/Responses/JobDetails.cs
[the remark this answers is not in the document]
WhitWaldo: I was torn on that one since it _is_ an acronym and not just a name. Updated nonetheless.

I just did a full-project find/refactor on DateTime to DateTimeOffset and have gotten them all.

src/Dapr.Jobs/Models/Responses/JobDetails.cs
WhitWaldo: That's fair - updated. Habit from my own development practice.

WhitWaldo: I was refactoring another part of this when I realized specifically where I'd gotten this from here beyond my own practice. The generated prototype client uses uint requiring [this cast](https://github.com/dapr/dotnet-sdk/blob/13c32ded3f0acba6df73f414ad1e68698fe04596/src/Dapr.Jobs/DaprJobsGrpcClient.cs#L88)

Regardless, I think this is worth exposing in this SDK as int and just having that range checker.

WhitWaldo:
@philliphoff I've thought about where you were coming from in the review and have overhauled the API substantially. There's now a single `ScheduleJobAsync` method that takes a `DaprJobSchedule` along with the other optional arguments. Each of the scheduling extension methods have been removed in favor of two that take an object and serialize it with the provided (albeit optional) `JsonSerializerOptions` and another that takes a string (presumably already serialized) and simply encodes it to UTF-8 bytes.

That leaves the `GetJobsAsync` and `DeleteJobsAsync` otherwise untouched.

A `DaprJobSchedule` supports the prefixed periods, Cron expressions, duration intervals and point-in-time date/time values via a series of factory methods:
- `Yearly` or `Monthly` (among others, except "@every") for prefixed "@yearly" and "@monthly" expressions, respectively)
- `FromDateTime` accepts a `DateTimeOffset` with a specified point-in-time value and converts to an ISO8601 value
- `FromDuration` accepts a `TimeSpan` and converts to the Goland duration value
- `FromExpression` accepts an open-ended string expression (and the comment calls out that it accepts either a Cron-like or '@' prefixed period string - it doesn't do any validation except null/empty checking
- `FromCronExpression` is for those that want a more guided experience building Cron expressions. I added a CronExpressionBuilder class with a fluent API for building out various valid expressions (on T value, every U period, from V to W periods, on X,Y,Z days/months). It overloads `ToString` to output the resulting expression, but it can be passed directly into `FromCronExpression` if the user wants to take that approach over providing their own value to `FromExpression`.

The `DaprJobSchedule` class uses an internal constructor - it's expected the users will use the factory methods to build it (this was mostly to avoid just having umpteen constructor overloads, but if you think that approach is clearer over the factories, I can pivot).

Because a `JobDetails` now returns a `DaprJobSchedule`, the former also includes four boolean properties identifying the expression value as a prefixed period, point-in-time, interval or Cron expression and the value itself can be retrieved from `ExpressionValue`.

I removed the exceptions as they weren't actually used anymore and cleaned up the larger solution to remove the use of the two shared types in favor of accessing them from `Dapr.Common`. I've set up `InternalsVisibleTo` for each project in the solution in that project so that the types aren't exposed as part of the public API, so this PR now tweaks just about all projects in the solution in this regard.

I'm working to add more unit tests for the `CronExpressionBuilder` and make sure I didn't miss any edge situations.

I'm going to go back through the remaining review comments and address each, but I wanted to put the latest changes in one easy-to-find place.

src/Dapr.Jobs/Models/Responses/JobDetails.cs
WhitWaldo: Having thought about it a bit, I agree - I've refactored accordingly. See the latest comment below for details.

src/Dapr.Jobs/Extensions/Helpers/Serialization/DaprOneTimeJobsSerializationExtensions.cs
WhitWaldo: I consolidated in-so-far as there's only a single `ScheduleJobsAsync` method on the client now, but there are still the two serializer extension methods for accepting a string -> UTF-8 bytes and object -> JSON serialization -> UTF-8 bytes (as opposed to the `ReadOnlyMemory<byte>` in on the `ScheduleJobsAsync` method itself which assumes the user took on serialization (and optionally compression) on their own.

src/Dapr.Jobs/Extensions/Helpers/Serialization/DaprCronJobsSerializationExtensions.cs
WhitWaldo: I hear you - while we _could_ do something like that here and wrap the payload with that information, it wouldn't be compatible through use of any of the other APIs/HTTP API. Perhaps a future version of the Jobs API itself could include a metadata dictionary we could store that in.

src/Dapr.Jobs/DaprJobsGrpcClient.cs
WhitWaldo: Done in the latest iteration

src/Dapr.Jobs/DaprJobsClient.cs
WhitWaldo: Ah, the @-prefixes are new in the docs since I wrote this. Yes, I think a class makes more sense in that light and gives us more flexibility should additional approaches be supported in the future. Added a variation of your take of the DaprJobSchedule in the latest iteration of this.

src/Dapr.Common/DaprGenericClientBuilder.cs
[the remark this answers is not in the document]
WhitWaldo: Given the similarity of the downloads between Dapr.Client and Dapr.AspNetCore, I'm curious if you know if there have been any surveys done on just how many people use the DaprClient without relying on dependency injection?

src/Dapr.Jobs/DaprJobsServiceException.cs
WhitWaldo: Ultimately dropped both exceptions because neither one is referenced anywhere. If we need them later on, we can always add them back.

WhitWaldo:
In the last several commits, I've been working to build out a regular expression that can evaluate whether a value in a `DaprJobSchedule` is a Cron expression or not (so it can be a bit more reliable than "it's not any of the other types, so it must be Cron"). I've made incremental progress to capture each of the separate segments, add testing and address unsupported cases, but it passes the current test suite which is pretty comprehensive right now.

I did want to call out why I opted for the approach I did in the latest commit. I created a separate project to run some benchmarking because I wanted to implement this in a performant way that also limited allocations and the results were mildly surprising. I have three different approaches called `SeparateRegex`, `CombinedRegex` and `GeneratorRegex` and they differ as follows:

##### SeparateRegex
This approach reflects how I developed the whole expression - one segment at a time. It works by splitting the expression by a whitespace character. If there are fewer than 6 segments, it's not a Cron expression. Otherwise, evaluate each segment by its respective regular expression and return a boolean value reflecting if they all were a match or not.

##### CombinedRegex
This approach builds off the completed work of the last one in that I create a compiled regular expression of the entire string (as follows) and evaluate the expression as provided:
> ^(\*|(\*\/([0-5]?\d))|(([0-5]?\d)(-([0-5]?\d))?)) (\*|(\*\/([0-5]?\d))|(([0-5]?\d)(-([0-5]?\d))?)) (\*\/(([0-1]?\d)|(2[0-3])))|((([0-1]?\d)|(2[0-3]))(-(([0-1]?\d)|(2[0-3])))?) (\*\/([0-1]?\d))|((([0-1]?\d)(-([0-1]?\d))?))|((,?(JAN|DEC|FEB|MAR|APR|MAY|JUN|JUL|AUG|SEP|OCT|NOV|DEC)+)+)|(-?(JAN|DEC|FEB|MAR|APR|MAY|JUN|JUL|AUG|SEP|OCT|NOV|DEC)) (\*\/(([0-2]?\d)|(3[0-1])))|(((([0-2]?\d)|(3[0-1]))(-(([0-2]?\d)|(3[0-1])))?)) (\*\/(0?[0-6])|(0?[0-6](-0?[0-6])?)|((,?(SUN|MON|TUE|WED|THU|FRI|SAT))+)|((SUN|MON|TUE|WED|THU|FRI|SAT)(-(SUN|MON|TUE|WED|THU|FRI|SAT))?))$

##### GeneratorRegex
This utilizes the new approach introduced in .NET 7 of using a source generator to produce a pared down version of the regular expression engine to only those bits necessary for the provided expression. 

The results of a standard benchmarking run + memory diagnostics are as follows:
image

Surprisingly, Separate performed the best, followed by the Generator approach and then Combined (all within 2 microseconds of one another, so pretty similar). But the allocations were a distinctly different story: Separate performed the worst by far with 5616 bytes versus 216 bytes for Combined and 152 for the Generator, which make a lot of sense and align with expectations.

As the 1.15 release of Dapr will precede the release of .NET 9 and will continue to have full support for .NET 6, it means that while source generators were introduced with .NET 7, use of the `GeneratedRegexAttribute` itself requires C# 11 which is only available in .NET 7 and up so it's not available for all targeted versions (Dapr targets the maximum common language version of C# 10 right now) so in a future release, we can make this tweak and save a few more kb of allocations.

src/Dapr.Common/DaprGenericClientBuilder.cs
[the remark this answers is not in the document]
philliphoff: I don't know of any, though my personal expectation is that more people are using the `AspNetCore` provided DI method than building one manually.

src/Dapr.Common/ArgumentVerifier.cs
```diff
@@ -0,0 +1,53 @@
...
+/// A utility class to perform argument validations. 
+/// </summary>
+internal static class ArgumentVerifier
+{
+    /// <summary>
+    /// Throws ArgumentNullException if argument is null.
+    /// </summary>
+    /// <param name="value">Argument value to check.</param>
+    /// <param name="name">Name of Argument.</param>
+    public static void ThrowIfNull([NotNull] object? value, string name)
```
philliphoff, line 32: Note, `ArgumentNullException.ThrowIfNull()` was added in .NET 6.

src/Dapr.Common/ArgumentVerifier.cs
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    /// <summary>
+    /// Validates string and throws:
+    /// ArgumentNullException if argument is null.
+    /// ArgumentException if argument is empty.
+    /// </summary>
+    /// <param name="value">Argument value to check.</param>
+    /// <param name="name">Name of Argument.</param>
+    public static void ThrowIfNullOrEmpty([NotNull] string? value, string name)
```
philliphoff, line 45: Note: `ArgumentException.ThrowIfNullOrEmpty()` was added in .NET 7 (so we should be able to ditch this these methods as soon as .NET 6 is deprecated.

src/Dapr.Jobs/DaprJobsGrpcClient.cs
```diff
@@ -0,0 +1,216 @@
...
+            job.Ttl = ((DateTimeOffset)ttl).ToString("O");
+        }
+
+        var envelope = new Autogenerated.ScheduleJobRequest { Job = job };
+
+        var callOptions = CreateCallOptions(headers: null, cancellationToken);
+
+        try
+        {
+            await client.ScheduleJobAlpha1Async(envelope, callOptions);
```
philliphoff, line 108: Pass the cancellation token?  And then, if/when a `RpcException` is thrown of type "cancelled" and our token was the requesting one, throw the equivalent `OperationCanceledException` (as we don't want to report that the Dapr endpoint indicated a failure if/when the caller was the one to do the cancellation).

src/Dapr.Jobs/Extensions/ByteArrayDeserializationExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+
+using System.Text;
+using System.Text.Json;
+
+namespace Dapr.Jobs.Extensions;
+
+/// <summary>
+/// Provides utility extensions for deserializing an array of UTF-8 encoded bytes.
+/// </summary>
+public static class ByteArrayDeserializationExtensions
```
philliphoff, line 22: `internal`?

src/Dapr.Jobs/Extensions/StringExtensions.cs
```diff
@@ -0,0 +1,25 @@
...
+
+internal static class StringExtensions
+{
+    /// <summary>
+    /// Extension method that validates a string against a list of possible matches.
+    /// </summary>
+    /// <param name="value">The string value to evaluate.</param>
+    /// <param name="possibleValues">The possible values to look for a match within.</param>
+    /// <returns></returns>
+    public static bool EndsWithAny(this string value, IReadOnlyList<string> possibleValues) => possibleValues.Any(value.EndsWith);
```
philliphoff, line 24: Nit: may want to allow/require an `StringComparison` to be passed.

src/Dapr.Jobs/Models/DaprJobSchedule.cs
```diff
@@ -0,0 +1,143 @@
...
+    public static DaprJobSchedule FromDuration(TimeSpan duration)
+    {
+        ArgumentVerifier.ThrowIfNull(duration, nameof(duration));
+        return new DaprJobSchedule(duration.ToDurationString());
+    }
+
+    /// <summary>
+    /// Specifies a schedule in which the job is triggered to run once a year.
+    /// </summary>
+    public static DaprJobSchedule Yearly => new DaprJobSchedule("@yearly");
```
philliphoff, line 94: Should these be `public static readonly` and assigned, rather than created on each call?

src/Dapr.Jobs/Models/Responses/JobDetails.cs
```diff
@@ -0,0 +1,45 @@
...
+// limitations under the License.
+// ------------------------------------------------------------------------
+
+namespace Dapr.Jobs.Models.Responses;
+
+/// <summary>
+/// Represents the details of a retrieved job.
+/// </summary>
+/// <param name="Schedule">Represents the schedule that triggers the job.</param>
+public sealed record JobDetails(DaprJobSchedule Schedule)
```
philliphoff, line 20: `DaprJobDetails`?

src/Dapr.Common/ArgumentVerifier.cs
WhitWaldo: Updated in the Jobs SDK, though there's plenty of room to remedy this across the rest of the solution in another PR.

src/Dapr.Jobs/Models/Responses/JobDetails.cs
WhitWaldo: Done

src/Dapr.Jobs/Extensions/StringExtensions.cs
WhitWaldo: Fair enough

src/Dapr.Jobs/Extensions/ByteArrayDeserializationExtensions.cs
WhitWaldo: Did you one better - removed altogether since its only usage anymore appeared to be a unit test (removed them as well).

src/Dapr.Jobs/Models/DaprJobSchedule.cs
WhitWaldo: I didn't see a point in having a backing field for each just for the readonly, so I gave them all getters and assigned the values once.

WhitWaldo:
@philliphoff I changed how the inbound job triggers are handled in the latest commit. Where it handled before more like how topics are subscribed to in the pub/sub context, after responding to an inquiry in Discord, it occurred to me that should someone schedule some jobs during runtime, this approach would require redeploying the app.

Instead, I've changed it to operate more like the dynamic subscription in that it maps to all POST requests to "/job/{jobName}". It attempts to deserialize the inbound request body to a `DaprJobDetails` and then passes both the job name and job details (or their null values) through to either of the async delegates (one provides an `IServiceProvider` in case someone needs to access some injected services as part of processing the data).

I've also updated the unit tests and example project to validate the new approach.

While updating the unit tests, I realized that I didn't have a really solid approach to dealing with the serialization and deserialization of the `DaprJobDetails` and `DaprJobSchedule`, so I've added some `JsonConverters` and an internal type so that the developer is still presented with a straightforward way of interacting with the SDK while ensuring that those complex types. This makes the payload itself smaller and simpler when it's serialized, so win-win.

src/Dapr.Jobs/DaprJobHandler.cs
```diff
@@ -0,0 +1,21 @@
...
+namespace Dapr.Jobs;
+
+/// <summary>
+/// The delegate representing how a triggered job is handled.
+/// </summary>
+/// <param name="serviceProvider">The service provider that facilitates injecting resources to the delegate.</param>
+/// <param name="jobName">The name of the triggered job.</param>
+/// <param name="jobDetails">The details of the triggered job.</param>
+/// <returns></returns>
+public delegate Task InjectableDaprJobHandler(IServiceProvider serviceProvider, string? jobName, DaprJobDetails? jobDetails );
```
philliphoff, line 12: A thought: there may be other things we'll want to pass to such handlers (e.g. a `CancellationToken`).  It might be best to consolidate the individual arguments into a `DaprJobHandlerContext` object and just pass that one thing.  In that way, additional things can be added in a way that won't break existing code and won't require additional overloads of the `MapDaprJob()` method.

It would be cool to be able to reuse some of ASP.NET's minimal API parameter binding logic so that users can just write their handlers in the same way as they normally would (while just injecting the additional job details), but I don't think that's exposed from the infrastructure.

WhitWaldo: @philliphoff 

I've been trying to think of some way to make this more flexible and while reading through a little bit of how ASP.NET Core does it in their minimal APIs, I've come up with a similar approach and while it's not as fancy as theirs, it also doesn't need to do route binding. Ideally, we can add an analyzer for this or something at some point, but the developer can pass a delegate of their own into the extension and the first two parameters must be a `string?` for the job name and a `DaprJobDetails?` for the job details. Anything after that is dynamically injected for them like the minimal APIs do and the cancellationToken is passed in as the last parameter.

This approach gets rid of the two delegate handlers I defined and means the developer doesn't have to deal with an `IServiceProvider`, but gets their `CancellationToken` and we don't have to deal with maintaining a context object.

I want to write several unit tests to validate this out and make sure it works, but I updated the [example](https://github.com/dapr/dotnet-sdk/blob/10cc382c60c4f13a4edb898c4d5e628bd46a97d0/examples/Jobs/JobsSample/Program.cs#L16) to show off the new approach.

Issue #1305, opened 2024-06-13
Actor example method invocation doesn't respect method signature


m3nax:
## Expected Behavior

The `SaveData` method called by cli without remoting (as written in readme) correctly invokes the actor

## Actual Behavior

As written in the readme of Actor example to call SaveData method from bash we need to use the following command:
``` BASH
curl -X POST http://127.0.0.1:3500/v1.0/actors/DemoActor/abc/method/SaveData -d '{ "PropertyA": "ValueA", "PropertyB": "ValueB" }'
```
but `SaveData` method have 2 parameters `data` and `ttl`
``` CSHARP
Task SaveData(MyData data, TimeSpan ttl);
```

## Steps to Reproduce the Problem

First try, following the readme, failing for missing ttl parameter
```
# Run example app
dapr run --dapr-http-port 3500 --app-id demo_actor --app-port 5010 dotnet run

# Execute
curl -X POST http://127.0.0.1:3500/v1.0/actors/DemoActor/abc/method/SaveData -d '{ "PropertyA": "ValueA", "PropertyB": "ValueB" }'
```

Second try, with fixed payload, failing with exception: 
` System.ArgumentException: Method IDemoActor.SaveData has more than one parameter and can't be invoked through http `
```
# Run example app
dapr run --dapr-http-port 3500 --app-id demo_actor --app-port 5010 dotnet run

# Execute
curl -X POST http://127.0.0.1:3500/v1.0/actors/DemoActor/abc/method/SaveData -d '{ "data": {"PropertyA": "ValueA", "PropertyB": "ValueB" }, "ttl": "00:10:00" }' 
```

## Release Note

RELEASE NOTE:

philliphoff:
@m3nax The original intent was that non-remoted (plain HTTP) actor methods accept only a single argument (ignoring `CancellationToken`); it was the remoting infrastructure that allowed for multiple arguments.  In this case, it would seem to be an oversight to have updated the sample and  `README.md` to match that limitation when the TTL feature was added.  (Probably a new type, `MyDataWithTTL`, should have been created with those two values as properties and passed to `SetData()` instead.)

[...]

PR #1334, opened 2024-08-02
Incremental source generator for actors

m3nax:
# Description

Converted the Actor source generator into an IncrementalSourceGenerator to ease the future definition of new generation pipeline for other objects (E.g: https://github.com/dapr/dotnet-sdk/issues/1305#issuecomment-2234152155)

It also add a null check for actorProxy parameter passed in generated Actor ctor, definition of AnalyzerReleases.Shipped.md and AnalyzerReleases.Unshipped.md diagnostics, remove [deprecated nugets](https://github.com/dotnet/roslyn-sdk/blob/main/src/Microsoft.CodeAnalysis.Testing/README.md#obsolete-packages).

## Issue reference

_

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


all.sln
```diff
@@ -348,3 +347,5 @@ Global
 		SolutionGuid = {65220BF2-EAE1-4CB2-AA58-EBE80768CB40}
 	EndGlobalSection
 EndGlobal
+lobalSection
```
KrylixZA, line 350: Something looks wrong here. It should at least say `GlobalSection`, but are you sure you need this specific section of changes at all?

src/Dapr.Actors.Generators/ActorClientGenerator.cs
```diff
@@ -20,284 +26,262 @@ namespace Dapr.Actors.Generators;
...
     }
 
-    #region ISourceGenerator Members
-
-    /// <inheritdoc />
-    public void Execute(GeneratorExecutionContext context)
+    /// <summary>
+    /// Returns the descriptor for the actor client to generate.
+    /// </summary>
+    /// <param name="context"></param>
```
KrylixZA, line 60: Meaningful documentation around parameters, expected returns and exceptions thrown goes a long way to making the library easier to use. Please could you fill in the docs here?

m3nax: Tomorrow i will improve summaries.

src/Dapr.Actors.Generators/Extensions/IEnumerableExtensions.cs
```diff
@@ -0,0 +1,34 @@
+﻿namespace Dapr.Actors.Generators.Extensions
+{
+    internal static class IEnumerableExtensions
+    {
+        /// <summary>
+        /// Returns the index of the first item in the sequence that satisfies the predicate. If no item satisfies the predicate, -1 is returned.
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
```
KrylixZA, line 8: Please remember to document the params here too.

KrylixZA: Maybe just double check all XML docs for this PR.

all.sln
m3nax: Fixed

m3nax:
@KrylixZA really thank for taking time to review my pr.

src/Dapr.Actors.Generators/ActorClientGenerator.cs
```diff
@@ -20,284 +26,265 @@ namespace Dapr.Actors.Generators;
...
-
-    /// <inheritdoc />
-    public void Execute(GeneratorExecutionContext context)
+    /// <summary>
+    /// Returns the descriptor for the actor client to generate.
+    /// </summary>
+    /// <param name="context">Current generator syntax context passed from generator pipeline.</param>
+    /// <param name="cancellationToken">Cancellation token used to interrupt the generation.</param>
+    /// <returns>Returns the descriptor of actor client to generate.</returns>
+    static ActorClientDescriptor CreateActorClientDescriptor(
```
KrylixZA, line 63: I presume the omission of an access modifier to this method is deliberate? Could we get away with this being private, or does it need to be internal (C# default)?

Same question for the `GenerateActorClientCode` method below.

src/Dapr.Actors.Generators/Templates.cs
```diff
@@ -0,0 +1,87 @@
...
+    {
+        /// <summary>
+        /// Returns the <see cref="SourceText"/> for the ActorMethodAttribute.
+        /// </summary>
+        /// <param name="destinationNamespace">Namespace where to generate attribute.</param>
+        /// <returns>The <see cref="SourceText"/> representing the ActorMethodAttribute.</returns>
+        /// <exception cref="ArgumentNullException">Throws when destinationNamespace is null.</exception>
+        public static SourceText ActorMethodAttributeSourceText(string destinationNamespace)
+        {
+            if (destinationNamespace == null)
```
KrylixZA, line 21: I am not going to dictate here, but I do prefer to use `is null` rather than `== null` checks in my code. This is because the `==` can be overwritten to behave differently. It's highly unlikely to happen, but it can. You can read up about it [on this StackOverflow thread](https://stackoverflow.com/questions/40676426/what-is-the-difference-between-x-is-null-and-x-null) if you'd like. 

If that's not your style, I'm happy for you to resolve this comment with a simple comment saying something like "I'll stick to `== null` in my code".

KrylixZA (approved):
My knowledge of source generators is limited. I joined the review more to learn than anything else. Thanks for the education! 😉

That said, LGTM. We'll still need one of the Dapr maintainers to approve before it merges though as I am not on that list yet.

Thanks for your contributing!

src/Dapr.Actors.Generators/ActorClientGenerator.cs
m3nax: Fixed

src/Dapr.Actors.Generators/Templates.cs
m3nax: Handled, thank

m3nax:
@WhitWaldo I have resolved the conflicts with the Master, do you have time to look at this branch?

src/Dapr.Actors.Generators/Extensions/IEnumerableExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+        internal static int IndexOf<T>(this IEnumerable<T> source, Func<T, bool> predicate)
+        {
+            if (predicate is null)
+            {
+                throw new ArgumentNullException(nameof(predicate));
+            }
+
+            int index = 0;
+
+            foreach (var item in source)
```
WhitWaldo, line 21: Instead of a foreach with a separate variable tracking the index, why not just use a for and repurpose the incrementing value?

src/Dapr.Actors.Generators/Helpers/SyntaxFactoryHelpers.cs
```diff
@@ -0,0 +1,159 @@
+﻿using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace Dapr.Actors.Generators.Helpers
+{
+    /// <summary>
+    /// Syntax factory helpers for generating syntax.
+    /// </summary>
+    public static partial class SyntaxFactoryHelpers
```
WhitWaldo, line 10: Do these need to be public or can they just be internal?

WhitWaldo (commented):
Two nits to address - otherwise looks good.

src/Dapr.Actors.Generators/Helpers/SyntaxFactoryHelpers.cs
m3nax: You are right

src/Dapr.Actors.Generators/Extensions/IEnumerableExtensions.cs
m3nax: To avoid enumerating the source parameter twice (one for the Count() and one to iterate source elements), I do not use the `Count()` function that I need to write the `for(int i=0; i < source.Count(); i++)`

WhitWaldo (approved):
Looks good to me!

Issue #1336, opened 2024-08-14
Prioritize IConfiguration for environment variable retrieval during DI registrations


WhitWaldo:
In today's .NET Dapr Client SDK, environment variables like `DAPR_API_KEY`, `DAPR_HTTP_ENDPOINT` and `DAPR_GRPC_ENDPOINT` are sourced directly as named, which can lead to unexpected registration conflicts to modern .NET developers. I suggest that the DI registration extensions for the DaprClient itself, but also the Workflow and Jobs clients be updated to prioritize retrieving these values directly from an injected `IConfiguration` for consistency and to align with ASP.NET Core's centralized configuration management approach [detailed here](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/configuration/).

For example, when registering each of these environment variables with a prefix, the current approach would recognize the prefixed variables leading to the client ignoring the specified and registered keys in favor or the defaults:

```cs
var builder = WebApplication.CreateBuilder(args);
builder.Configuration.AddEnvironmentVariables("test_"); //Retrieves all the envvars that start with "test_" and removes the prefix when sourced from IConfiguration
builder.Services.AddDaprClient(); //Would ignore the IConfiguration and default to empty strings as it wouldn't find the expected envvars

var app = builder.Build();
app.Run();
```

Currently, the DaprClient DI registration bypasses `IConfiguration` leading to direct envvar retrieval or default empty strings as a fallback defying expectations. 

This change should have a minimal impact as it wouldn't require signature changes and would simply augment the implementation already in use to source first from IConfiguration, where available, and then to an environment variable and finally an empty string.

I'll provide a PR within the next week with a proposal to illustrate what I'm suggesting and would welcome any feedback.

PR #1338, opened 2024-08-14
Prioritize retrieval of environment variables from IConfiguration instead of directly

WhitWaldo:
# Description
I've made a couple of changes here:
1) Refactored DaprDefaults to pull the environment variable names out to public constants so other classes can reference them (and eliminate possible typos). Also refactored the default ports for the same reason.
2) Updated the DaprServiceCollectionExtensions to prefer pulling the Dapr HTTP endpoint/port, gRPC endpoint/port and API token from IConfiguration, failing over to pull from their respective environment variables directly and then defaulting to their default localhost ports if not otherwise available.

I noticed while putting this together that the environment variables documentation doesn't include either `DAPR_HTTP_ENDPOINT` nor `DAPR_GRPC_ENDPOINT` and have created a [separate PR](https://github.com/dapr/docs/pull/4313) to include them. 

### Possible breaking change
But as neither `DAPR_HTTP_ENDPOINT` nor `DAPR_GRPC_ENDPOINT` are otherwise indicated in the docs to clarify this, I assume it's a bug that today's DaprDefaults will ignore the ports specified in `DAPR_HTTP_PORT` or `DAPR_GRPC_PORT` if the endpoint is specified (opting to use the endpoint specified with its port instead) and will only use them when the endpoint isn't specified (and defaults instead to "127.0.0.1").

I don't think that's necessarily right - that the endpoint should generally specify only the schema and hostname and that if the port is provided, it should be used (instead of potentially inferring the port from the schema, e.g. https => 443 or http => 80). As such, this implementation changes this behavior and will set the endpoint (schema, host and port) and then set the port if also specified.

## Issue reference

Please reference the issue this PR will close: #1336 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests

This isn't yet ready to merge as I want to apply the same to the other Dapr clients as well (Jobs, Workflow, etc.) and largely get away from using DaprDefaults where possible.

Also, this presently builds atop the [Jobs PR](https://github.com/dapr/dotnet-sdk/pull/1331), so it appears a little heavier than it necessarily is.


WhitWaldo:
Closing - going to replace with a PR not based on the Jobs PR.

Issue #1342, opened 2024-08-28
How to run the Actor tests in the DevContainer?


paule96:
## Ask your question here

I tried lately to contribute to the dapr-sdk for dotnet. I was happy to see that there is a DevContainer configured for this repository.
Sadly it seems to be not anymore maintained.
I first added the missing dotnet SDKs (7 & 8), which let me successfully build and develop the code.
After that I tried to run the tests. Some of the tests require dapr to be initialized. So I did run `dapr init`.

After that the tests are still failing the following errors:

Version 1:

```bash
---- System.Exception : Command: "run --app-id serializerapp --dapr-http-port 45969 --dapr-grpc-port 41649 --metrics-port 39537 --components-path ./../../../../../test/Dapr.E2E.Test/components --config ./../../../../../test/Dapr.E2E.Test/configuration/featureconfig.yaml --log-level debug --app-port 44271 -- dotnet run --project ./../../../../../test/Dapr.E2E.Test.App/Dapr.E2E.Test.App.csproj --framework net6 -- 44271 --urls http://localhost:44271 --json-serialization" timed out while waiting for output: "System.String[]"
```

Version 2:

```bash
Flag --components-path has been deprecated, This flag is deprecated and will be removed in the future releases. Use "resources-path" flag instead
ℹ️Starting Dapr with id serializerapp. HTTP Port: 34887. gRPC Port: 38055
Flag --components-path has been deprecated, use --resources-path
Flag --dapr-http-max-request-size has been deprecated, use '--max-body-size 4Mi'
Flag --dapr-http-read-buffer-size has been deprecated, use '--read-buffer-size 4Ki'
time="2024-08-28T07:28:44.155062526Z" level=info msg="Starting Dapr Runtime -- version 1.14.1 -- commit 81cb1f5c4709535b03799e7e1826fbab0315b90b" app_id=serializerapp instance=8491b9a3b59a scope=dapr.runtime type=log ver=1.14.1
time="2024-08-28T07:28:44.155088148Z" level=info msg="Log level set to: debug" app_id=serializerapp instance=8491b9a3b59a scope=dapr.runtime type=log ver=1.14.1
time="2024-08-28T07:28:44.155301613Z" level=warning msg="mTLS is disabled. Skipping certificate request and tls validation" app_id=serializerapp instance=8491b9a3b59a scope=dapr.runtime.security type=log ver=1.14.1
[...]

Issue #476, opened 2020-11-13
Actors uses different serializers for different purposes, not all of them are pluggable


rynowak:
## Expected Behavior

Actors uses a single serializer by default, and allows flexiblity for different choices for compatibility or user freedom.

The problem with not providing a consistent default is that different serializers have non-overlapping features sets, and different configuration settings. The choice of serializer usually ends up being tightly coupled with the users' data model through attributes and other extensibility.

For instance DataContract serializer requires types and properties to "opt-in" to serialization - System.Text.Json assumes that types and public properties should be serialized - and you should "opt-out" for anything you want to exclude. It takes **non-obvious work** to migrate between these models, and many users are not familiar with the nuances.

## Actual Behavior

Actors uses DataContract serializer (XML) for "remoting" method invocation (the choice of DCS is not configurable).

Actors uses System.Text.Json (JSON) for "non-remoting" method invocation (neither the choice of S.T.J nor the options passed to it are configurable).

Actors uses System.Text.Json (JSON) for state storage (the serializer and options have a replaceable abstraction).

Sidenote: Actors currently uses BinaryFormatter for exceptions in "remoting" method invocation, but that is planned for removal.

## Steps to Reproduce the Problem

Build an Actor system that uses "remoting". Try to share data models between the state store and the actor method invocation. You'll find that you need to annotate data models for *both* DataContract serializer and System.Text.Json.

## Release Note

We need to surface a design proposal and assess the impact of a change here. This is not a good users experience, but no simple change will fix it.

rynowak:
We're not going to make any changes here. This has some rough edges (using the same types with state and actor proxy invocation) but users that are using Actors today are overall happy with the experience. We'll revisit this in the future based on feedback.

tomkerkhove:
We are actually having issues with this. Would be good to standardize on one serializer.

techgeek03:
This is a also an issue for us. @rynowak I started conversation on discord in the dotnet-sdk channel. I want to make a proposal for the change. Can you please comment on that and let me know if you think I should create new proposal.

onionhammer:
Having a ton of issues with this..

virtualcca:
Is there an update to this issue?

Now we encounter that if some newer types (such as dateonly) are returned in the interface method of the actor, DCS(DataContract serializer)related errors  will appear

Under Net7, STJ can support dateonly these types
Now you can only temporarily use DateOnly to be converted to DateTime as an Actor's interface to external calls

[...]

PR #1317, opened 2024-07-03
Implementing nullability annotations on Dapr.Client SDK

WhitWaldo:
# Description

This implements the nullability annotations across the Dapr.Client SDK both on the type level and with a few changes to the logic as necessary to facilitate these changes.

I couldn't leave the challenge posted [here](https://github.com/dapr/dotnet-sdk/pull/1259#pullrequestreview-1987064520) unanswered.

I again raise the issue over at #1261 to propose dropping support for both .NET 6 and .NET 7 as both are out of support officially and further, adopting a standard policy of targeting the [latest C# language version](https://github.com/dapr/dotnet-sdk/issues/1318). Here, this is relevant because while the csproj doesn't lack support for the `<nullability>enable</nullability>` tag, we don't adopt a specific C# language version, so because we target .NET 6, we implicitly target C# 10 and C# 10 requires explicit `#nullability enable` notations at the tops of each file implementing the same (e.g. it ignores the csproj annotation). 

## Issue reference

We strive to have all PR being opened based on an issue, where the problem or feature have been discussed prior to implementation.

Please reference the issue this PR will close: #1316

Though this itself won't close #1316, but will at least let us check off the box for Dapr.Client as that's a larger over-arching issue tracking the end goal.

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [X] Code compiles correctly
* [X] Created/updated tests
* [N/A] Extended the documentation

Issue #1046, opened 2023-02-27
Establish .NET Runtime Support Policy


halspang:
## Background
The .NET SDK needs to establish a firm guideline on which versions of .NET it will build on and support in testing. 

This topic comes up somewhat frequently, a quick scan gives at least two issues that have been created for different purposes and this conversation comes up in other mediums as well.

#1036 
#925 

As such, a firm and known stance for what we support is important.

## Proposal
I propose that this SDK adopts Microsoft's [support policy for .NET Core](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core). 

We will support the versions of .NET that are currently receiving patches/support. When a version of .NET is no longer listed as supported, the first release of this SDK after the date of deprecation will remove the framework from our build/testing. 

This will keep us working with modern frameworks that are receiving updates for features, security, and bug fixes. Given the small nature of maintainers/approvers of this SDK, this also serves to lower the burden there and keep the SDK more focused.

For the main binaries, we will still target 1 release to build to. This will be the oldest LTS release (currently `net6`). We will then support all LTS and STS releases in our testing framework. Once the LTS release we're using as our build runtime is out of support, we'll move to the next LTS. For example, when `net6` retires, we'll move to building with `net8`, skipping over `net7`.

This does mean that we won't be looking at supporting things like `netstandard2.0/2.1`. Again, this is mostly a maintenance issue, but it also allows us to focus on new functionalities and features.

halspang:
It appears we've had this conversation before with @rynowak, another maintainer. #429

halspang:
Given the lack of discussion here and what was said in this issue, we'll adopt the Microsoft support policy for .NET: https://dotnet.microsoft.com/en-us/platform/support/policy. 

After 1.11 is released, we'll remove any remaining references to anything < .NET 6. Going forward our releases will match that cadence so the first release after a version has gone EOL it will be removed from the SDK.

philliphoff:
With 1.13, we moved to explicitly targeting *both* .NET 6 and .NET 8 (which were *both* under LTS at the time) as this allows the SDK to take advantage of advancements/optimizations of newer runtimes rather than a "lowest common denominator" approach.

With 1.14, explicit targeting of .NET 7 is considered "deprecated" (only Workflow explicitly targeted it anyway) and that target/testing will be removed for 1.15.

[...]

PR #1343, opened 2024-08-28
Bug/476 multiple methods per interface with JSON serialization doesn´t work

paule96:
# Description

Now the Actor manager provides the `MethodName` for the serialization `ActorMessageSerializersManager`.
This makes it now possible for the JSON serialization to support Actor interfaces with more than one method defined in it.

As a side effect, this PR solves #1342 too. 
Changes made in the dev container:

- use docker-in-docker instead of docker-from-docker
   - now the devcontainer is isolated from the host and is not affected by the host configuration. But it also means accessing the dapr instance of the devcontainer is only accessible from inside the container.
- install all dotnet sdks that the tests requires. (dotnet 6 is still the default, maybe we should change that too)
- add the `C# DevKit` as an additional extension. This can be a breaking change for people that don't have a visual studio license and want to run the dev container on their local host. Users that run the dev container on GitHub get a license automatically injected.
- Unblocks the `localinit.sh` script before running it

Changes made to the E2E Tests

- it's now easier to debug failing start of the E2E test App.

## Issue reference

It's part of the issue #476 starting with this [comment](https://github.com/dapr/dotnet-sdk/issues/476#issuecomment-2306990754).

It fixes also #1342 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
   * partly, I added at the interesting locations in the code comments.
   * This change shouldn't affect the current documented actor model, it just fixes a broken feature

cc @onionhammer as original contributor :)


paule96:
@onionhammer it would be nice if you could try out these changes. For me, it seems to not be possible yet. I will look the next days into #1342 to solve the debugging / contribution story for others.

paule96:
@onionhammer This is now ready to review :) That was a fun research. I tried to don't break existing code paths as much as I could.

onionhammer (approved):
This looks good to me, although I have no authority here :)

WhitWaldo:
> install all dotnet sdks that the tests requires. (dotnet 6 is still the default, maybe we should change that too)
.NET 7 fell out of support May 14 of this year and .NET 8 will do the same on November 10, 2024 with .NET 9 being introduced in the same time frame. That should be right around the 1.15 release, so I think there will be a good case to be made to drop .NET 6 and .NET 7 explicit support, target .NET 8 instead and include .NET 9 in the test suite for 1.16.

paule96:
@WhitWaldo  if I change the default version to dotnet 9.0 the new security vulnerability scanning of the dotnet CLI turns on implicitly and produces, of course, a lot of errors. 

I guess the upgrading of the default SDK needs to be delayed. 

maybe a separate issue is needed to clean up the solution of, the old tests, old implementation, and what else is connected to it. I guess the documentation needs to be updated too.
In that issue, the devcontainer shouldn't be forgotten. 

So for now I would try to just update the default to dotnet 8.0 (support until November 10, 2026, it's LTS), and postpone the change to dotnet 9.0 to a later point in time.

WhitWaldo:
That's right - the [plan](https://github.com/dapr/dotnet-sdk/issues/1046) is to put off changing the targeted releases until after the release of Dapr 1.15. There, we'll target .NET 6 and .NET 8 still, so no changes needed in the short term. But we'll put something prominent announcing EOL in the release notes so it's clear that in 1.16 we'll drop support for .NET 6 in favor of .NET 8 (and language version C# 12). An open question on that thread is whether we should go on targeting .NET 9 as well (though I'm also inclined to lean yes since there are frequently so many performance improvements in each).

You're right, that'll require some significant doc updates in addition to the other projects, but I've got a [draft PR](https://github.com/dapr/dotnet-sdk/pull/1317) that'll likely handle the latter as I'm gradually adding nullability annotations throughout the .NET SDK and there are several other PRs in place that'll shift up some of the project layouts as well that'll help.

Things to tackle after the next release!

paule96:
@WhitWaldo do you know anybody I could ping to get a review? (maybe yourself)

WhitWaldo:
@paule96 I'm not a maintainer, so my check isn't worth a whole lot. I suspect this will get more attention as we edge closer to the 1.15 release and there's more of a general code freeze. If it hasn't been looked at closer to that timeline, I'll give it a look myself (for whatever it might be worth).

paule96:
@dapr/maintainers-dotnet-sdk can anybody review this pr :)

yaron2:
cc @halspang @philliphoff

paule96:
@halspang & @philliphoff friendly reminder

WhitWaldo:
I'm not already super familiar with how Dapr handles serialization under the hood, so I'm learning as I read through your changes here. The approach already taken registers the interface and references it throughout using an int, why it does this, I'm not entirely sure, but I'd assume it's to avoid name conflicts and shorten the amount of data passed around. If that's at all accurate, I'm concerned that this approach, which binds the serialization to the name of the method runs a risk of being insufficiently specific (especially in light of #1350 and the opportunity for polymorphic bindings reflecting the same method names).

Now, perhaps that's not a problem, as different interfaces should register different IDs and you're keying the serializers by both this interface and method. Perhaps you've considered this, but given your richer familiarity with the underlying problem space, does it make more sense to have a similar ID registry for the method names as we appear to have for the actor interfaces rather than the approach you've taken here?

Second, could you add some unit tests that prove that this approach works sufficiently well for method overloads?

paule96:
@WhitWaldo sorry for the late reply. Was kinda busy these days.
Thank you for pointing out the other pullrequest. 

My first comment to overloads is:

It's not supported. Their is an explicit exception for that in the file: https://github.com/dapr/dotnet-sdk/blob/ee8be673374f7ff0c65b9ff4f506de874f6a2ad3/src/Dapr.Actors/Description/InterfaceDescription.cs#L135-L151

I will try to see what I can do to enable support for it. But I guess it will still work with names instead of Ids. The problematic part is, that we need to type match here. And because the transport types are json, it maybe becomes a challenge. (not sure yet)



And I can understand that actors get an "id". Because their polymorphy is possible. By the language design of methods I'm not aware of any function that will provide polymophic behavior to methods. 

The only thing that I could see, if to trick around with inheritance, to like use the `new` modifier in a class for methods. But that would effectively not work in dapr actors space, because the actors are proxied with the interface, and the interface needs to point to an exact method and not to a `new` flagged method.

paule96:
@WhitWaldo I now have created the basic changes for method overloading. I will not push it to this branch or PR. The problem is, that by the design of the actors part of dapr, it doesn't seem to be supported.

For now I would put it on hold, but would like to know where we can find the framework independent documentation for dapr. Where you can see how dapr works outside of the SDK you use. 
The reason is write now on my local changes, I changed the method name from:

`yourcoolnamespace.evenbetterclassname.SuperFancyMethod`

to:

`yourcoolnamespace.evenbetterclassname.SuperFancyMethod()`

or if it has parameters:

`yourcoolnamespace.evenbetterclassname.SuperFancyMethod(FullNameOfParameterType, FullNameOfParameterType)`

And I'm quite sure that this will be a breaking change, in special if you try to call an actor from example a python app.
But I couldn't find the documentation where this standards are documented

WhitWaldo:
> @WhitWaldo I now have created the basic changes for method overloading. I will not push it to this branch or PR. The problem is, that by the design of the actors part of dapr, it doesn't seem to be supported.
> 
> For now I would put it on hold, but would like to know where we can find the framework independent documentation for dapr. Where you can see how dapr works outside of the SDK you use. The reason is write now on my local changes, I changed the method name from:
> 
> `yourcoolnamespace.evenbetterclassname.SuperFancyMethod`
> 
> to:
> 
> `yourcoolnamespace.evenbetterclassname.SuperFancyMethod()`
> 
> or if it has parameters:
> 
> `yourcoolnamespace.evenbetterclassname.SuperFancyMethod(FullNameOfParameterType, FullNameOfParameterType)`
> 
> And I'm quite sure that this will be a breaking change, in special if you try to call an actor from example a python app. But I couldn't find the documentation where this standards are documented

The actor runtime itself is in [dapr/dapr](https://github.com/dapr/dapr/tree/master/pkg/actors), but I don't know that there's any specific documentation available that details any of it. As I understand it, it works similarly under the hood to [Orleans](https://learn.microsoft.com/en-us/dotnet/orleans/overview) (well, sort of) and [Service Fabric Actors](https://learn.microsoft.com/en-us/azure/service-fabric/service-fabric-reliable-actors-introduction) (closer)

paule96:
@WhitWaldo so can we skip the topic for now on this pullrequest and maybe open a new issue for method overloading? Because that was not scope of this PR

WhitWaldo:
@paule96 I agree method overloading isn't in scope, but I'm concerned about the opportunity for developers attempt to do so and then run into an odd runtime experience (as happens where with more than one method per interface).

I'm going to review this again this week and see if we can't get it merged as-is.

paule96:
@WhitWaldo is their a chance that you can take a lock to this :)

paule96:
@WhitWaldo  done :)

WhitWaldo (approved):
Looks good to me!

WhitWaldo:
I neglected to say it earlier - thank you very much for your contribution here @paule96 !

marcduiker:
@holopin-bot @paule96 Big thanks from me as well! Here's a digital badge as a small token of appreciation!

Issue #1324, opened 2024-07-09
Add Streaming Subscription Support


JoshVanL:
Dapr v1.14 adds support for [streaming subscriptions](https://github.com/dapr/proposals/blob/main/0013-RS-pubsub-subscription-streaming.md) which allows for dynamically subscribing to pubsub topics over a bi-directional gRPC stream.  This API is exposed via the [`SubscribeTopicEventsAlpha1`](https://github.com/dapr/dapr/blob/master/dapr/proto/runtime/v1/dapr.proto#L64) gRPC API.

On first call, the client sends an [`InitialRequest`](https://github.com/dapr/dapr/blob/7ea7016c0150e6a7b7bc7c30e34d56f0001c5e90/dapr/proto/runtime/v1/dapr.proto#L434) containing the pubsub, topic name, etc. From then on, topic event messages are sent from daprd to the client over the stream. The client reports the processing status of this message back to daprd. A message will not be considered processed until the client has responded with this status. There can be multiple in-flight topic messages. Daprd unsubscribes from the topic when the stream is closed.

[Here](https://github.com/dapr/go-sdk/pull/578/) is a reference implementation for the go-sdk.

philliphoff:
Some initial thoughts on a possible implementation:

 - To align with the "component specific clients" approach, creation of a new `Dapr.PublishSubscribe` (`Dapr.Messaging`?) package
 - Creation of a new `Dapr.DaprPublishSubscribeClient` type
 - Client exposes a `Subscribe()` method that accepts a handler called for each topic event

```csharp
namespace Dapr.PublishSubscribe;

public sealed record DaprSubscriptionOptions
{
  public IReadOnlyDictionary<string, string> Metadata { get; init; } = Dictionary<string, string>();

  public string? DeadLetterTopic { get; init; }
}

public sealed record TopicRequest
{
    public required string Id { get; init; }
    public required string Source { get; init; }
    public required string Type { get; init; }
    public required string SpecVersion { get; init; }
    public required string DataContentType { get; init; }
    public required string Topic { get; init; }
    public required string PubSubName { get; init; }
    public string? Path { get; init; }
    public object? Extensions { get; init; } // TODO: Determine what this should look like.
}

public delegate Task<TopicResponse> TopicHandler(TopicRequest request);

public abstract class DaprPublishSubscribeClient
{
    public DaprPublishSubscribeClient CreateClient();

[...]

PR #1346, opened 2024-09-04
Add .NET client for dynamic pubsub subscriptions

WhitWaldo:
# Description

Implements streaming subscription support as a discrete Dapr.Messaging package.

This was based on the work done by @philliphoff [here](https://github.com/dapr/dotnet-sdk/tree/philliphoff-streaming-subs). The requirements call for a mechanism to return an unbounded stream of message from the subscription and a follow-up call to the sidecar for each message to convey an action that should be taken on it (e.g. drop, retry or mark as completed).

While I think his use of a delegate to force the developer to process each message and immediately notate the action that should be taken on the message, I instead wanted to expose a method returning an IAsyncEnumerable as I think there's some interesting opportunities for combining this with some Rx goodness. 

~~My approach then also requires that in addition to setting up the initial subscription, the developer call `AcknowledgeMessageAsync` with the message ID and the action to take, but also bakes in a policy so that a default configurable action is taken after a timeout window in which the developer fails to otherwise indicate success or failure.~~

~~Because a separate connection will be created with the sidecar for every combination of pubsub component and topic, the actual implementation is an internal class called `PublishSubscribeReceiver` that maintains a single streaming connection to the sidecar for each instance in the `ConnectionManager`. In order to facilitate some future backpressure support and otherwise decouple the receipt of messages from the sidecar from the subscription, I write inbound messages to a `Channel<TopicMessage>` and read from it from the subscription method implementing `IAsyncEnumerable<TopicMessage>`. As a side effect of each message read out, it also registers the message identifier with a `TaskCompletionSource` and a `CancellationTokenSource` bound to the provided cancellation token so that if the developer fails to acknowledge the message within the configured timeout window, it the default message handler action will be taken automatically.~~

*Update:* After hammering out a concept of the above, I ultimately agreed with @philliphoff that the approach isn't as smooth as his original approach, so I've modified it accordingly and detailed it more in [this comment](https://github.com/dapr/dotnet-sdk/pull/1346#discussion_r1760465212).

Finally, I opted to implement the extensions property on the `TopicMessage` with a `Dictionary<string, Value>` where the `Value` is the Protobuf Value struct (meaning that a developer is going to have to query the type of the struct and retrieve the appropriate property with the matching value). While I considered just strongly converting all values to a string, there are some Values types this can represent that would make that cumbersome (e.g. List or other structs). Ultimately, if the developer is opting to use this, they can figure out how to retrieve the Value they intend to.

# Additional notes
This should ideally be merged after #1331 as it takes a dependency on a package introduced (Dapr.Common which introduces a generic client builder class and contains shared exceptions) in that PR. I created a stub for it here, but it'd be nice to avoid the conflict later on.

This implementation also creates a separate Dapr.Protos package that each of the existing libraries can take advantage of once this is merged in that ensures that there's only one place that Protos have to be updated at instead of each library having its own copy.

I think there's an opportunity to improve on how we do logging across all these distinct projects using a static logging type. As the internal GrpcClient isn't DI-injected, it eliminates several of the simpler approaches to just use an `ILogger<>` or `ILoggerFactory`. I removed the use of the configuration options as a carrier of the logging implementation as it just didn't feel like a great fit there, but I think it's worth exploring a more concrete and centrally-defined approach to more uniform logging going forward for the .NET client. But perhaps in another issue.

## Issue reference

Please reference the issue this PR will close: #1324 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Extended the documentation - [https://github.com/dapr/docs/pull/4358](https://github.com/dapr/docs/pull/4358)


src/Dapr.Messaging/Dapr.Messaging.csproj
```diff
@@ -0,0 +1,21 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <Description>This package contains the reference assemblies for developing messaging services using Dapr.</Description>
+        <ImplicitUsings>enable</ImplicitUsings>
+        <Nullable>enable</Nullable>
+        <PackageId>Dapr.Messaging</PackageId>
+        <Title>Dapr Messaging SDK</Title>
+        <Description>Dapr Messaging SDK for building applications that utilize messaging components.</Description>
+        <VersionSuffix>alpha</VersionSuffix>
```
philliphoff, line 10: Probably reasonable to add an "alpha" suffix but otherwise keep the assembly version the same as other Dapr libraries.  (I had some feedback on the jobs PR related to this.)

src/Dapr.Messaging/PublishSubscribe/ConnectionManager.cs
```diff
@@ -0,0 +1,69 @@
...
+// ------------------------------------------------------------------------
+
+using Grpc.Core;
+using C = Dapr.AppCallback.Autogen.Grpc.v1;
+using P = Dapr.Client.Autogen.Grpc.v1;
+
+namespace Dapr.Messaging.PublishSubscribe;
+
+/// <summary>
+/// Maintains access to
```
philliphoff, line 21: Nit: incomplete docs.

src/Dapr.Messaging/PublishSubscribe/DaprPublishSubscribeClient.cs
```diff
@@ -0,0 +1,64 @@
...
+{
+    /// <summary>
+    /// Dynamically subscribes to a Publish/Subscribe component and topic.
+    /// </summary>
+    /// <param name="pubsubName">The name of the Publish/Subscribe component.</param>
+    /// <param name="topicName">The name of the topic to subscribe to.</param>
+    /// <param name="options">Configuration options.</param>
+    /// <param name="cancellationToken">Cancellation token.</param>
+    /// <returns>An <see cref="IAsyncEnumerable{TopicMessage}"/> containing the various messages returned by the subscription.</returns>
+    public abstract IAsyncEnumerable<TopicMessage> SubscribeAsync(string pubsubName, string topicName, DaprSubscriptionOptions options, CancellationToken cancellationToken);
```
philliphoff, line 29: Are you subscribed to the messages after completion of `SubscribeAsync()` or only after calling `GetAsyncEnumerator()` on the resulting `IAsyncEnumerator`?  If the former, are messages buffered until enumeration actually starts?  If the latter, can multiple calls to `GetAsyncEnumerator()` be made?

src/Dapr.Messaging/PublishSubscribe/DaprPublishSubscribeGrpcClient.cs
```diff
@@ -0,0 +1,94 @@
...
+    /// </summary>
+    /// <param name="pubsubName">The name of the Publish/Subscribe component.</param>
+    /// <param name="topicName">The name of the topic to subscribe to.</param>
+    /// <param name="options">Configuration options.</param>
+    /// <param name="cancellationToken">Cancellation token.</param>
+    /// <returns>An <see cref="IAsyncEnumerable{TopicMessage}"/> containing the various messages returned by the subscription.</returns>
+    public override IAsyncEnumerable<TopicMessage> SubscribeAsync(string pubsubName, string topicName, DaprSubscriptionOptions options,
+        CancellationToken cancellationToken)
+    {
+        var receiver = new PublishSubscribeReceiver(pubsubName, topicName, options, daprClient);
```
philliphoff, line 54: Do we need to guard against concurrent/duplicate subscriptions to the same pubsub/topics?

src/Dapr.Messaging/PublishSubscribe/DaprPublishSubscribeGrpcClient.cs
```diff
@@ -0,0 +1,94 @@
...
+        await receiver.AcknowledgeMessageAsync(messageId, messageAction, cancellationToken);
+    }
+
+    /// <summary>
+    /// Unsubscribes a streaming subscription for the specified Publish/Subscribe component and topic.
+    /// </summary>
+    /// <param name="pubsubName">The name of the Publish/Subscribe component.</param>
+    /// <param name="topicName">The name of the topic to subscribe to.</param>
+    /// <param name="cancellationToken">Cancellation token.</param>
+    public override async Task UnsubscribeAsync(string pubsubName, string topicName, CancellationToken cancellationToken)
```
philliphoff, line 85: I feel like the thing that's returned from `SubscribeAsync()` should be used to unsubscribe.  Having a separate method on the client feels a bit disconnected.

src/Dapr.Messaging/PublishSubscribe/Extensions/PublishSubscribeServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,104 @@
...
+        return services;
+    }
+
+    /// <summary>
+    /// Adds Dapr Publish/Subscribe support to the service collection.
+    /// </summary>
+    /// <param name="services">The <see cref="IServiceCollection"/>.</param>
+    /// <param name="configure">Optionally allows greater configuration of the <see cref="DaprPublishSubscribeClient"/> using injected services.</param>
+    /// <returns></returns>
+    public static IServiceCollection AddDaprPubSubClient(this IServiceCollection services, Action<IServiceProvider, DaprPublishSubscribeClientBuilder>? configure)
```
philliphoff, line 83: Could the first two overloads ultimately just call into this method?

src/Dapr.Messaging/PublishSubscribe/PublishSubscribeReceiver.cs
```diff
@@ -0,0 +1,261 @@
...
+            await tcs.Task;
+        }
+    }
+
+    /// <summary>
+    /// Retrieves the subscription stream data from the Dapr sidecar.
+    /// </summary>
+    /// <param name="channelWriter">The channel writer instance.</param>
+    /// <param name="cancellationToken">Cancellation token.</param>
+    private async Task FetchDataFromSidecar(ChannelWriter<TopicMessage> channelWriter, CancellationToken cancellationToken)
```
philliphoff, line 195: Nit: `FetchDataFromSidecarAsync()`

src/Dapr.Messaging/PublishSubscribe/PublishSubscribeReceiver.cs
```diff
@@ -0,0 +1,261 @@
...
+    /// <summary>
+    /// Specifies the action that should be taken on the message after processing it.
+    /// </summary>
+    /// <param name="messageId">The identifier of the message to acknowledge.</param>
+    /// <param name="messageAction">The action to take on the message.</param>
+    /// <param name="cancellationToken">Cancellation token.</param>
+    public async Task AcknowledgeMessageAsync(string messageId, TopicMessageAction messageAction, CancellationToken cancellationToken)
+    {
+        var stream = await connectionManager.GetStreamAsync(cancellationToken);
+        await stream.RequestStream.WriteAsync(new P.SubscribeTopicEventsRequestAlpha1
```
philliphoff, line 94: Note that gRPC streams must be written to sequentially and the underlying gRPC libraries do not manage such threading.  Writers need to ensure they have exclusive rights to the stream.

src/Dapr.Messaging/PublishSubscribe/PublishSubscribeReceiver.cs
```diff
@@ -0,0 +1,261 @@
...
+    private readonly DaprSubscriptionOptions options;
+    /// <summary>
+    /// A channel used to decouple the messages received from the sidecar to their consumption.
+    /// </summary>
+    private readonly Channel<TopicMessage> channel = Channel.CreateUnbounded<TopicMessage>();
+    /// <summary>
+    /// A collection of <see cref="TaskCompletionSource"/> used to signal acknowledgement of received messages so a status
+    /// can be sent back to the sidecar indicating what behavior should happen to each.
+    /// </summary>
+    private readonly Dictionary<string, TaskCompletionSource<bool>> acknowledgementTasks = new();
```
philliphoff, line 53: I wonder if a `ConcurrentDictionary` could avoid the acknowledgement locks.

src/Dapr.Messaging/PublishSubscribe/DaprPublishSubscribeClient.cs
philliphoff: (I think I have my answers having looked at the implementation, but we may want to be explicit about the semantics here so that the semantics cannot vary at the whims of the current implementation.)

philliphoff, line 29: While I'm all for an `IAsyncEnumerable<T>` based method, I wonder if some users may find it a bit intimidating.  I still think a `IDisposable SubscribeAsync(string pubsubName, string topicName, DaprSubscriptionOptions, DaprTopicHandler handler)` style method would be useful and still align with Rx concepts.

In this case, `DaprTopicHandler` would be defined as something like `Func<TopicMessage, Func<Task>, Task>`, an async delegate that accepts a `TopicMessage` and an async "ack" method.  It would probably be pretty easy to have an extension method that just layers on top of the `IAsyncEnumerable` method.

src/Dapr.Messaging/PublishSubscribe/PublishSubscribeReceiver.cs
WhitWaldo: Solid idea. Replaced here and in the `WaitForAcknowledgementAsync` method.

src/Dapr.Messaging/PublishSubscribe/Extensions/PublishSubscribeServiceCollectionExtensions.cs
WhitWaldo: Definitely. Done.

src/Dapr.Messaging/PublishSubscribe/PublishSubscribeReceiver.cs
WhitWaldo: That's definitely a pro to taking your initial approach. I'm working on rewriting this to adopt that style with some tweaks and will have an update soon.

src/Dapr.Messaging/PublishSubscribe/DaprPublishSubscribeGrpcClient.cs
WhitWaldo: Good call - I think so. If this later supports some sort of partitioning functionality to allow multiple subscribers to access the same resource, we can revisit it, but I've added a guard against this so existing receivers will be re-used.

WhitWaldo: The approach I've taken here is simple - allow only one subscriber per topic name and pubsub component name and throw if a duplicate subscription is made. 

We can revisit this to allow multiple subscriptions per combination down the line, but the biggest difficulty in enabling that is lack of consumer group/partition support on the Dapr runtime side. Put simply, having more than one consumer of the same topic will open the door to consumers potentially getting out of sync on deliveries and potentially returning different acknowledgements for the same message identifiers. I propose we save that problem for another day and opt for the easier approach for now. 

Should developers want to support multiple subscribers, they need only persist the messages to an internal queue/channel from the delegate and pull from it how they want (as it wouldn't need acknowledgement support at that downstream level) to easily work around this SDK restriction.

src/Dapr.Messaging/PublishSubscribe/DaprPublishSubscribeClient.cs
WhitWaldo: I've switched this up to more closely resemble the implementation you started with. I agree that it's just cleaner and far more simply accommodates the sequential acknowledgement (didn't realize that initially) without having all these other methods the developer is responsible for calling here and there. The handler just sets up callers up for success without having to think much about what's happening behind the scenes.

As of the latest commit, the user calls Register on the `DaprPublishSubscribeGrpcClient` and that returns the `PublishSubscribeReceiver` corresponding to the pubsub component name and topic name (or throws if that combination is already registered). The developer then calls `SubscribeAsync` on that and it kicks off an unbounded operation to read the messages from the sidecar to a Channel on the type (for backpressure purposes). In the same method, the messages are read out of the Channel one by one and a `CancellationTokenSource` linked with the timeout value specified in the configuration options. Each message is evaluated by the handler delegate which returns an action to take and either this result is passed back to the sidecar or the cancellation token will have timed out and the default action returned.

This way, the messages are buffered if they're returned faster than the handler is evaluating them at any given moment. As a mention in another comment in here somewhere, this leaves the door open for this implementation to eventually support consumer groups/partitions at some later point to allow richer back-pressure mitigation.

src/Dapr.Messaging/PublishSubscribe/PublishSubscribeReceiver.cs
WhitWaldo: Update in place - described [here](https://github.com/dapr/dotnet-sdk/pull/1346#discussion_r1760465212)

src/Dapr.Messaging/PublishSubscribe/DaprPublishSubscribeClient.cs
WhitWaldo: I think the approach you started with makes more sense in light of the sequential acknowledgement requirement. I describe the new approach [here](https://github.com/dapr/dotnet-sdk/pull/1346#discussion_r1760465212).

src/Dapr.Messaging/PublishSubscribe/DaprPublishSubscribeGrpcClient.cs
```diff
@@ -24,71 +24,37 @@ internal sealed class DaprPublishSubscribeGrpcClient : DaprPublishSubscribeClien
...
-    public override async Task UnsubscribeAsync(string pubsubName, string topicName, CancellationToken cancellationToken)
+    /// <returns></returns>
+    public override IAsyncDisposable Register(string pubSubName, string topicName, DaprSubscriptionOptions options, TopicMessageHandler messageHandler, CancellationToken cancellationToken)
     {
-        if (!clients.TryGetValue((pubsubName, topicName), out var receiver))
-        {
-            throw new Exception($"Unable to find receiver instance for specified publish/subscribe component name '{pubsubName}' and topic '{topicName}'.");
-        }
+        var key = (pubSubName, topicName);
+        if (clients.ContainsKey(key))
```
philliphoff, line 53: There's a race condition here if two threads call `Register()` at the same time.  It could be resolved with a simple `lock` or, say, a `ConcurrentDictionary` and `Lazy<T>` scheme.

src/Dapr.Messaging/PublishSubscribe/PublishSubscribeReceiver.cs
```diff
@@ -47,159 +42,124 @@ internal sealed class PublishSubscribeReceiver : IAsyncDisposable
...
-                Id = messageId,
-                Status = new C.TopicEventResponse
+                while (channel.Reader.TryRead(out var message))
                 {
-                    Status = messageAction switch
+                    using var cts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+                    cts.CancelAfter(options.MessageHandlingPolicy.TimeoutDuration);
+
+                    //Evaluate the message and return an acknowledgement result
+                    var messageAction = await messageHandler(message, cts.Token);
```
philliphoff, line 92: So with this change, acknowledgments are sent back to the sidecar in the same order in which their corresponding message was received, correct?  I thinks that's fine--but I don't think we *have* to do that, necessarily.  (Just to be clear, while we need to write to the stream sequentially, I don't think the sidecar requires they be written in any particular order.  One option would be to have a separate `Channel` for acks that is written to as soon as the message handling is complete (or times out).  Then a separate `Task` can sit and wait on the channel, writing to the gRPC stream as they come in.  That would allow messages to be ack'd as soon as possible while also ensuring acks are written to the stream one-by-one.

There may be a larger question of whether it makes sense for an application to potentially have multiple messages being handled at the same time, or whether it would prefer messages to be handled one-by-one.

src/Dapr.Messaging/PublishSubscribe/DaprPublishSubscribeGrpcClient.cs
```diff
@@ -24,71 +24,37 @@ internal sealed class DaprPublishSubscribeGrpcClient : DaprPublishSubscribeClien
...
+            throw new Exception(
+                $"A subscription has already been created for Dapr pub/sub component '{pubSubName}' and topic '{topicName}'");
 
-        await receiver.DisposeAsync();
+        clients[key] = new PublishSubscribeReceiver(pubSubName, topicName, options, connectionManager, messageHandler);
```
philliphoff, line 57: Will callers need to subsequently call `SubscribeAsync()` on the returned receiver?  (The method signature only returns `IAsyncDisposable` so I'm not sure how they'd do that.)

WhitWaldo: Whoops - missed that one. Fixed so it returns `PublishSubscribeReceiver`. They'll have to call `SubscribeAsync` on that, which is is why I renamed this method `Register` rather than have back-to-back `Subscribe` methods.

src/Dapr.Messaging/PublishSubscribe/DaprPublishSubscribeGrpcClient.cs
WhitWaldo: Done

src/Dapr.Messaging/PublishSubscribe/PublishSubscribeReceiver.cs
WhitWaldo: I confirmed with @yaron2 that the message acknowledgement order doesn't matter. I was looking through the Go implementation and it doesn't seem to have any restriction on there being multiple subscribers to the same combination - it just kicks off yet another subscription. I don't see any reason we can't do that ourselves then and, as you suggested, just have them all write to a Channel and have a single task read off it to handles the acks. I'll have an update with this soon.

src/Dapr.Messaging/PublishSubscribe/DaprPublishSubscribeGrpcClient.cs
WhitWaldo: Revisited this later on - no such restriction any longer.

src/Dapr.Messaging/PublishSubscribe/ConnectionManager.cs
```diff
@@ -0,0 +1,148 @@
...
+using C = Dapr.AppCallback.Autogen.Grpc.v1;
+using P = Dapr.Client.Autogen.Grpc.v1;
+
+namespace Dapr.Messaging.PublishSubscribe;
+
+/// <summary>
+/// Maintains the streaming connection to the Dapr sidecar so it can be repurposed without
+/// multiple callers opening separate connections.
+/// </summary>
+internal sealed class ConnectionManager : IAsyncDisposable
```
philliphoff, line 25: This seems to get reused for each `PublishSubscribeReceiver`, right?  I don't think that should be so, if I understand the semantics of the `SubscribeTopicEventsAlpha1()` API. When called, the method returns a bi-directional stream.  The Dapr runtime expects that the very first "request" sent from the caller to be the topic to subscribe to, and then the remainder of the "requests" will be message acknowledgements (until either side closes the stream).

To subscribe to multiple topics in parallel, or to subscribe after a previous subscription has completed, the client must create a new stream (i.e. another call to `SubscribeTopicEventsAlpha1()`) for each subscription, right?

WhitWaldo: In light of @JoshVanL 's confirmation of @philliphoff 's read of the spec, I've gone ahead and made an update to remove the `ConnectionManager` altogether. Instead, the following now happens:

Each call to `DaprPublishSubscribeClient.Register` (via `DaprPublishSubscribeGrpcClient`) creates a new `PublishSubscribeReceiver` for the indicated pubsub component name, topic name, options, and handling delegate. Following a call to `SubscribeAsync` with a cancellation token, This validates that the subscription hasn't already been initialized (to prevent the initial request from being submitted twice), then does the following:
1) Sends the initial request via `SubscribeTopicEventsRequestInitialAlpha1` and sets up an asynchronous reader of inbound messages to pass into the topic message `Channel` writer.
2) Sets up an asynchronous reader of the acknowledgements `Channel` reader holding each of the topic acknowledgements (each containing a message ID and `TopicEventResponseStatus` reflecting the action to take on each message).
3) Sets up an asynchronous reader of the topic message `Channel` reader holding each of the messages to pass each into the delegate `TopicMessageHandler` as provided to be evaluated.

As each message is evaluated, either an action is returned or it times out and a default configured action is paired with the message ID and persisted back to the acknowledgements `Channel`. As these "acknowledgements" are received, they're sent back over the persisted connection to the sidecar.

I refactored this a little bit so that if the operation is cancelled or disposed, it attempts first to complete and drain the inbound message `Channel` then complete and drain the acknowledgement `Channel`.

No changes needed to the example project already written.

philliphoff: > How should the underlying connection be shared if it cannot receive more than one initial request?

I assume @JoshVanL is referring to the underlying gRPC connection (i.e. in .NET, `GrpcChannel`)?

src/Dapr.Messaging/PublishSubscribe/PublishSubscribeReceiver.cs
```diff
@@ -0,0 +1,293 @@
...
+    }
+
+    /// <summary>
+    /// Dynamically subscribes to messages on a PubSub topic provided by the Dapr sidecar.
+    /// </summary>
+    /// <param name="cancellationToken">Cancellation token.</param>
+    /// <returns>An <see cref="IAsyncEnumerable{TopicMessage}"/> containing messages provided by the sidecar.</returns>
+    public async Task SubscribeAsync(CancellationToken cancellationToken = default)
+    {
+        //Prevents the receiver from performing the subscribe operation more than once (as the multiple initialization messages would cancel the stream).
```
philliphoff, line 96: If the intent is that `SubscribeAsync()` is called once and only once, and there's nothing of any consequence that  the user can do between `Register()` and `SubscribeAsync()` (e.g. setup listeners, as that's done in the `Register()` call itself), then perhaps `SubscribeAsync()` should just be a private, implicit call by `Register[Async]()`?  In that case, `Register()` can just become `SubscribeAsync()` which returns an `IAsyncDisposable`.  Messages automatically start flowing when the method completes and stops when returned the disposable is called.

src/Dapr.Messaging/PublishSubscribe/PublishSubscribeReceiver.cs
```diff
@@ -0,0 +1,293 @@
...
+        if (isDisposed)
+            return;
+        isDisposed = true;
+
+        //Stop processing new events
+        topicMessagesChannel.Writer.Complete();
+        
+        try
+        {
+            //Process any remaining messages on the channel
```
philliphoff, line 266: We could argue whether to wait for already pending handlers, but I'm fairly certain we wouldn't want to execute any *new* ones.  That is, we can rely on the Dapr runtime to do the right thing if the subscription stream is closed without receiving acks for some delivered messages (i.e. those messages should get redelivered when the subscription is registered again).

If we wanted to get fancy, we could add a `DrainAsync()` to be optionally called before `DisposeAsync()` which stopped accepting new work but allowed known messages to start/complete handling (within some set timeout).  Then `DisposeAsync()` could be made more of a "halt immediately" operation.  That said, I'm fine with only having `DisposeAsync()` and it being a more graceful "wind things down, don't start new work but wait a reasonable amount of time for existing work to finish" method.

In any case, we need to ensure that we only wait for *some* period of time before bailing altogether, both for pending handlers as well as pushing the acks across the wire.

We probably also want to add some optional logging so that callers can have some visibility into the shutdown process (e.g. why is it taking 30s for the application to exit?).  In that vein, we could allow the user to pass an optional `ILoggerFactory`.

WhitWaldo: On the ILoggerFactory note, I was planning to put some time (after this and the Jobs API PR merge) to see if we can't better support a DI-first approach across all the new (and old) SDKs as it would really facilitate dealing with log and metric collection, but also provide a means for users to submit serialization and/or compression handlers as optional registrations so the SDK can handle each while not having to pass them around all over the place. Then on this class, we could just put an optional `ILoggerFactory?` and failover to a `NullLoggerFactory` if the user doesn't provide their own.

Especially if there's any sort of Dapr usage survey held again, I'd really like to get a feel for how many people are relying on DaprClientBuilder.Build to get a DaprClient instance and how many just use the DI package (given that the NuGet package counts for both Dapr.Client and Dapr.AspNetCore are pretty even). I'd really like to get to a place where this can be used universally in a DI-first manner.

To this immediate point though, that makes sense to leave additional messages unprocessed and only handle acknowledgements. I was originally thinking that there probably wouldn't be a ton of messages in there, but if the delegate is a bit heavy and the service has been running a while, I can see there being some sizable backlog over time. But especially if the message has been acknowledged, I can see the user being confused if those messages showed back up just because they incorrectly disposed of the subscriber on their end and it terminated right there.

src/Dapr.Messaging/PublishSubscribe/PublishSubscribeReceiver.cs
```diff
@@ -0,0 +1,293 @@
...
+        var acknowledgementProcessorTask = ProcessAcknowledgementChannelMessagesAsync(cancellationToken);
+        var topicMessageProcessorTask = ProcessTopicChannelMessagesAsync(cancellationToken);
+
+        try
+        {
+            await Task.WhenAll(fetchMessagesTask, acknowledgementProcessorTask, topicMessageProcessorTask);
+        }
+        catch (OperationCanceledException)
+        {
+            await DisposeAsync();
```
philliphoff, line 116: Nit: the caller should dispose of the instance regardless of whether it completed naturally, was cancelled, or otherwise failed, so probably no reason to do this here.

src/Dapr.Messaging/PublishSubscribe/PublishSubscribeReceiver.cs
```diff
@@ -0,0 +1,293 @@
...
+                initialRequest.Metadata.Add(key, value);
+            }
+        }
+        
+        //Send this request to the Dapr sidecar
+        await stream.RequestStream.WriteAsync(
+            new P.SubscribeTopicEventsRequestAlpha1 { InitialRequest = initialRequest }, cancellationToken);
+
+        //Each time a message is received from the stream, push it into the topic messages channel
+        await foreach (var response in stream.ResponseStream.ReadAllAsync(cancellationToken))
```
philliphoff, line 236: I wonder if there will be a need to bound this (and perhaps make configurable)?  Forcing messages to stay "in the channel" would (should?) put backpressure on the Dapr runtime to keep it from potentially overwhelming the application or, in the case of multiple application replicas, prevent one replica from starving the other by accepting all the messages (but not yet processing them).  The Dapr runtime should manage this somewhat through round-robin-ing delivery amongst the connected streams, though.

src/Dapr.Messaging/PublishSubscribe/PublishSubscribeReceiver.cs
WhitWaldo: Immediately marking both writers as completed during DisposeAsync and added a timeout to the options that's honored during the acknowledgement message drain.

src/Dapr.Messaging/PublishSubscribe/PublishSubscribeReceiver.cs
WhitWaldo: Done + updated example to reflect the change

src/Dapr.Messaging/PublishSubscribe/PublishSubscribeReceiver.cs
WhitWaldo: Good point

src/Dapr.Messaging/PublishSubscribe/PublishSubscribeReceiver.cs
WhitWaldo: Trivial addition since we've got Channels under the hood. Added an optional configuration value to `DaprSubscriptionOptions` and if not null and > 0, it'll create a bounded channel instead of an unbounded channel and wait to push additional messages until there's room in the subscription loop.

src/Dapr.Messaging/PublishSubscribe/PublishSubscribeReceiver.cs
```diff
@@ -0,0 +1,293 @@
...
+        }
+    }
+
+    /// <summary>
+    /// Acknowledges the indicated message back to the Dapr sidecar with an indicated behavior to take on the message.
+    /// </summary>
+    /// <param name="messageId">The identifier of the message the behavior is in reference to.</param>
+    /// <param name="behavior">The behavior to take on the message as indicated by either the message handler or timeout message handling configuration.</param>
+    /// <returns></returns>
+    private async Task AcknowledgeMessageAsync(string messageId, TopicResponseAction behavior)
```
philliphoff, line 156: Nit: this should also take a `CancellationToken` argument?  (But probably the original one passed to `ProcessTopicChannelMessagesAsync()`, not the linked one as we shouldn't consider time writing the acknowledgement as part of the time given to the user's handler.

WhitWaldo: Might as well - added

src/Dapr.Messaging/PublishSubscribe/PublishSubscribeReceiver.cs
WhitWaldo: We talked about the DI bits a good deal in Discord and I've got a POC with a rough version of what I had in mind that I linked you, so going to mark this one as resolved for now.

src/Dapr.Messaging/PublishSubscribe/TopicMessage.cs
```diff
@@ -0,0 +1,46 @@
...
+
+    /// <summary>
+    /// The matching path from the topic subscription/routes (if specified) for this event.
+    /// </summary>
+    public string? Path { get; init; }
+
+    /// <summary>
+    /// A map of additional custom properties sent to the app. These are considered to be CloudEvent extensions.
+    /// </summary>
+    public Dictionary<string, Value> Extensions { get; init; } = new();
```
philliphoff, line 45: Nit: should be `IReadOnlyDictionary`?

src/Dapr.Messaging/PublishSubscribe/PublishSubscribeReceiver.cs
```diff
@@ -0,0 +1,305 @@
...
+
+    /// <summary>
+    /// Dynamically subscribes to messages on a PubSub topic provided by the Dapr sidecar.
+    /// </summary>
+    /// <param name="cancellationToken">Cancellation token.</param>
+    /// <returns>An <see cref="IAsyncEnumerable{TopicMessage}"/> containing messages provided by the sidecar.</returns>
+    internal async Task SubscribeAsync(CancellationToken cancellationToken = default)
+    {
+        //Prevents the receiver from performing the subscribe operation more than once (as the multiple initialization messages would cancel the stream).
+        if (hasInitialized)
```
philliphoff, line 108: This doesn't completely eliminate the risk of a race (e.g. two threads in a thread pool).  That said, given this is now an internal method that is intended to (and we guarantee is) called only once, we might just dispense with it (and just document the method as "not thread safe").

src/Dapr.Messaging/PublishSubscribe/PublishSubscribeReceiver.cs
```diff
@@ -0,0 +1,305 @@
...
+using Grpc.Core;
+using P = Dapr.Client.Autogen.Grpc.v1;
+
+namespace Dapr.Messaging.PublishSubscribe;
+
+/// <summary>
+/// A thread-safe implementation of a receiver for messages from a specified Dapr publish/subscribe component and
+/// topic.
+/// </summary>
+public sealed class PublishSubscribeReceiver : IAsyncDisposable
```
philliphoff, line 25: Nit: this could be `internal` now?

PR #1347, opened 2024-09-13
Added unit test to Secrets API test suite

WhitWaldo:
# Description

Just added a simple unit test to the Secrets API test suite to validate that a secret that contains forward slashes in it would have the secret name properly realized through and through (e.g. that it wouldn't limit to any single part of the path).

## Issue reference

Please reference the issue this PR will close: Not a filed issue so much as [a concern](https://discord.com/channels/778680217417809931/901141875585609869/1283946035387039805) raised on Discord.

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


test/Dapr.Client.Test/SecretApiTest.cs
```diff
@@ -93,6 +95,32 @@ public async Task GetSecretAsync_ReturnSingleSecret()
...
+            {
+                return await DaprClient.GetSecretAsync("testStore", "us-west-1/org/xpto/secretabc");
+            });
+
+            request.Dismiss();
+
+            //Get Request and validate
+            var envelope = await request.GetRequestEnvelopeAsync<Autogenerated.GetSecretRequest>();
+            envelope.StoreName.Should().Be("testStore");
+            envelope.Key.Should().Be("us-west-1/org/xpto/secretabc");
```
KrylixZA, line 113: Would it be worth defining a method-level constant for the value of `"us-west-1/org/xpto/secretabc"` since it's repeated 5 times in this test?

test/Dapr.Client.Test/SecretApiTest.cs
```diff
@@ -11,6 +11,8 @@
 // limitations under the License.
 // ------------------------------------------------------------------------
 
+using FluentAssertions.Equivalency;
```
KrylixZA, line 14: I see this using is outside of the namespace, where the `using System;` is within. How much do we care about consistency?

test/Dapr.Client.Test/SecretApiTest.cs
WhitWaldo: Sure, it could be. But this is just a unit test and not part of the library, so I habitually tend to favor ease of readability over performance or optimization.

test/Dapr.Client.Test/SecretApiTest.cs
WhitWaldo: This isn't in use by anything, so I've just removed it in the latest commit. Undoubtedly an over-eager addition by the IDE.

WhitWaldo:
@philliphoff or @halspang  When you have a moment, would you please give this a quick once-over so we can merge it? 

No impact on the client - just adding a unit test to validate coverage for a concern raised on Discord.

philliphoff (approved):

Issue #694, opened 2021-06-06
Actor descriptions do not take interface inheritance into account


eNeRGy164:
Based on comment in #648

@eNeRGy164 - can you please separate this commit from this PR and open an issue to discuss. Any behavior changes need to be really well understood. 

I suspect that this is a breaking change we would not accept until the next major release as it *tightens* validation we're performing. We need to talk through all of the possible cases that would be impacted.

_Originally posted by @rynowak in https://github.com/dapr/dotnet-sdk/pull/684#discussion_r646188069_

# Issue
If an interface inherits from another interface (that does inherit from `IActor`. The inherited methods are not detected.

This is because the `Type.GetMethods` method does not return inherited methods from interfaces (in contrast to `class`es.

This resuts in 2 effects.
1. For inherited methods no method descriptions are generated
2. The detection for overloads will not generate an exception when the two overloads are declared on different interfaces.

I created an implementation that does detect method from inheritance, but this would probably be a breaking change.

rynowak:
@halspang - can you double check whether this is done or not?

halspang:
@rynowak - It does not appear to be done. The linked PR is just changing some messages and adding tests. One of which appears to specifically check/enforce this behavior?

```csharp
public void InterfaceDescription_CreateGeneratesMethodDescriptions_WhenTypeHasTaskMethods_ButDoesNotSeeInheritedMethods()
{
    // Arrange
    Type type = typeof(IChildActor);

    // Act
    TestDescription description = new(type);

    // Assert
    using var _ = new AssertionScope();
    description.Methods.Should().NotContainNulls();
    description.Methods.Should().AllBeOfType<MethodDescription>();
    description.Methods.Should().BeEquivalentTo(
        new { Name = "GetInt" }
    );
}
```

PR #1350, opened 2024-09-24
Facility to provide interface while registering actors

### LINKED ISSUE #1341: Allow to define which interfaces are important for the dapr actor host
## Describe the proposal

Currently, if you do something like:

```csharp
builder.Services.AddActors(options =>
{
    options.Actors.RegisterActor<MyActor>(nameof(MyActor));
});
```

The complete interface hierarchy needs to be compatible with actors. So if you have something like:

```csharp
public interface IInternalMyActor : IMyActor
{
    void SomeInternalMethod()
}

public interface IMyActor: IActor
{
    Task SomeMethod()
}

public class MyActor: Actor, IInternalMyActor{
   // TODO: implementation
}
```

If you try to start this actor host you will get the following error message:

```
: 'Method 'SomeInternalMethod' of actor interface 'MySample.IInternalMyActor ' does not return Task or Task<>. The actor interface methods must be async and must return either Task or Task<>. (Parameter 'actorInterfaceType')'
```

Error happens at `InterfaceDescription:224`.

It would be nice if we would be able to provide at the actor registration another generic parameter which interface should be used to access the actor. So like:

```csharp
builder.Services.AddActors(options =>
{
    options.Actors.RegisterActor<IMyActor, MyActor>(nameof(MyActor));
});
```

That would make it possible to have a quite complicated structure of interface on an actor. Also it would make it alot easier to use `default interface methods`, because you can have an more advanced `IActor` interface in your project that just map members from the class `Actor`, so you can use it in your `default interface methods`.

siri-varma:
# Description

Added facility to provide Interface types while registering Actors.

## Issue reference

Please reference the issue this PR will close: #1341

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


siri-varma:
@philliphoff @cgillum Would you folks be able to help with approving the workflows please ?

WhitWaldo:
@siri-varma All in all, I think this looks fine, but I think it'd be helpful to have documentation calling this out as it might be unclear to someone why they're experiencing an exception when implementing an interface that doesn't itself ultimately inherit from an IActor parent.

Otherwise, I think all the code looks fine. If you could write up a few words about it in [this section](https://github.com/dapr/dotnet-sdk/blob/b8e276728935c66b0a335b5aa2ca4102c560dd3d/daprdocs/content/en/dotnet-sdk-docs/dotnet-actors/dotnet-actors-usage.md#dependency-injection), I'd be happy to accept this.

siri-varma:
@WhitWaldo I updated the readme.md. Let me know if it looks good. Thanks

daprdocs/content/en/dotnet-sdk-docs/dotnet-actors/dotnet-actors-usage.md
```diff
@@ -136,8 +136,11 @@ public void ConfigureServices(IServiceCollection services)
     // Register actor runtime with DI
     services.AddActors(options =>
     {
-        // Register actor types and configure actor settings
+        // Register actor types and configure actor settings.
```
WhitWaldo, line 139: Just a rule, I don't consider periods necessary on comments - more often than not, comments reflect partial sentence fragments and not full sentences due appropriate capitalization and punctuation.

siri-varma: Removed the period.

WhitWaldo:
This will also close #694

daprdocs/content/en/dotnet-sdk-docs/dotnet-actors/dotnet-actors-usage.md
```diff
@@ -150,6 +153,12 @@ public void ConfigureServices(IServiceCollection services)
     services.AddSingleton<BankService>();
 }
 ```
+> [!IMPORTANT]  
+> When registering actors, note the return type requirements for the methods inside the interfaces:
+> * Pattern 1: `options.Actors.RegisterActor<MyActor>()`
+>     * In this case, all interfaces implemented by `MyActor` must have methods that return only `Task` or `Task<T>`. This applies to every method in all interfaces `MyActor` implements.
+> * Pattern 2: `options.Actors.RegisterActor<IMyActor, MyActor>()`
+>     * Here, only the `IMyActor` interface is required to have methods that return `Task` or `Task<T>`. Other interfaces `MyActor` are not subject to this restriction.
```
m3nax, line 161: If we force the use of `IActor` interface we can add a Diagnostics to block the compilation when the derived class contains methods that doesn't return `Task<T>`. Similar to what is done in the actors' source generator for the `CancellationToken` ([example](https://github.com/dapr/dotnet-sdk/blob/a50e6de3063cbc85649b56aa357d705888b96489/src/Dapr.Actors.Generators/Diagnostics/CancellationTokensMustBeTheLastArgument.cs))

I usually don't like to find out particularities from documentation, if possible I prefer my editor to guide me.

What do you think about it?

WhitWaldo:
First of all, thank you for putting this together.

However, `Dapr.Actors` is in maintenance and new functionality is being added to [`Dapr.Actors.Next`](https://github.com/dapr/dotnet-sdk/releases#release-v1.19.0-preview.2). This concept is already embodied in the new Actors SDK (both for manual and automatic registrations).

I certainly would welcome any contributions in the future!

Issue #1349, opened 2024-09-05
DaprClient.GetMetadataAsync() throwing Exception


sanjeev-saxena-us:
Dapr.Client .NET package, this used to work in 1.10 but appears to be broken in 1.12 onwards; is now throwing a "Object reference not set to an instance". Have tried with 1.14 and same issue.

alicejgibbons:
Able to reproduce with the following findings:
- only happens when `app-port` (app channel) is set and hasn't been found/established by daprd
- Running without app channel or when daprd has discovered the app already allows `DaprClient.GetMetadataAsync()` to work fine.

Eg. Running a sample dotnet app like the following works fine, regardless of where the `GetMetadataAsync()` command is
`dapr run --app-id DaprClient --dapr-http-port 5780 --dapr-grpc-port 5701 -- dotnet run`

But running
`dapr run --app-id DaprClient --app-port 5292 --dapr-http-port 5780 --dapr-grpc-port 5701` fails with the following error when calling `GetMetadataAsync()` ***before*** the application has been discovered on specified port. 

```
Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
   at Dapr.Client.DaprClientGrpc.GetMetadataAsync(CancellationToken cancellationToken)
   at Program.<Main>$(String[] args) in /Users/alicegibbons/Documents/Customers/Alegeus/DaprClientTest/TodoApi/Program.cs:line 15
   at Program.<Main>(String[] args)
```

JoshVanL:
Based on my limited DotNet experience, I believe the issue here is that the [DotNet client method](https://github.com/dapr/dotnet-sdk/blob/3d1fa01d0f52298c880f8d92b08690aa317be270/src/Dapr.Client/DaprClientGrpc.cs#L2394) is not checking for a nil pointer of the gRPC message for `ActorRuntime`. It is currently the case that `ActorRuntime` in the Metadata message will be nil until the app port is available and read by daprd.

```
[...]

Issue #1296, opened 2024-05-30
GetMetadataAsync throws NullReferenceException


sclarke81:
## Expected Behavior

Calling GetMetadataAsync should return the metadata.

## Actual Behavior

Calling GetMetadataAsync throws a NullReferenceException because `response.ActorRuntime` is null.

## Steps to Reproduce the Problem

I'm using Dapr self-hosted without Docker containers (`dapr init --slim`). I'm also using the [Dapr Sidekick](https://github.com/man-group/dapr-sidekick-dotnet) project to run the sidecars. I haven't started the placement service because I'm not using actors.

## Release Note

RELEASE NOTE:

jpiquot:
I have the same issue on Azure Container Apps (1.12.5) but works in local development machine (1.13.2) running with Aspire.

WhitWaldo:
Should be fixed with #1353

PR #1353, opened 2024-10-06
Added fix to handle null return values

WhitWaldo:
# Description

Added some simple checks to handle null response values.

## Issue reference

Please reference the issue this PR will close: #1349 and #1296

## Checklist

* [X] Code compiles correctly
* [N/A] Created/updated tests - going to wait on this until #1317 is in place


src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -2390,10 +2390,17 @@ public override async Task<DaprMetadata> GetMetadataAsync(CancellationToken canc
...
             {
                 var response = await client.GetMetadataAsync(new Autogenerated.GetMetadataRequest(), options);
-                return new DaprMetadata(response.Id,
-                                        response.ActorRuntime.ActiveActors.Select(c => new DaprActorMetadata(c.Type, c.Count)).ToList(),
-                                        response.ExtendedMetadata.ToDictionary(c => c.Key, c => c.Value),
-                                        response.RegisteredComponents.Select(c => new DaprComponentsMetadata(c.Name, c.Type, c.Version, c.Capabilities.ToArray())).ToList());
+                if (response is null)
+                    return null;
+
+                return new DaprMetadata(response.Id ?? "",
```
KrylixZA, line 2396: I don't quite know the standard for this repo but I know a lot of developers prefer to use `string.Empty` over `""`. Is that a valid recommendation for the Dapr .NET SDK?

KrylixZA (approved):
LGTM, just one comment.

src/Dapr.Client/DaprClientGrpc.cs
WhitWaldo: To my knowledge, there isn't a convention in the SDK either way with regards to that. Pretty much the only one I'm familiar with is private properties named using camel case without an underscore prefix.

I suspect the convention used throughout the rest of Dapr.Client will be "" for empty strings. That might be a worthwhile convention to set as we gradually migrate to individual clients for each feature, but I don't know if it's worth enforcing at this juncture (especially as it'd mean there's some inconsistency for a while).

KrylixZA: Sounds good to me.

philliphoff: As a third alternative, I've always preferred `String.Empty`. :-)

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -2390,10 +2390,17 @@ public override async Task<DaprMetadata> GetMetadataAsync(CancellationToken canc
             try
             {
                 var response = await client.GetMetadataAsync(new Autogenerated.GetMetadataRequest(), options);
-                return new DaprMetadata(response.Id,
-                                        response.ActorRuntime.ActiveActors.Select(c => new DaprActorMetadata(c.Type, c.Count)).ToList(),
-                                        response.ExtendedMetadata.ToDictionary(c => c.Key, c => c.Value),
-                                        response.RegisteredComponents.Select(c => new DaprComponentsMetadata(c.Name, c.Type, c.Version, c.Capabilities.ToArray())).ToList());
+                if (response is null)
```
philliphoff, line 2393: Probably not necessary?  A successful gRPC method will *always* return a valid response object?  (It's properties may be `null`, but the object itself cannot be?)

philliphoff (dismissed):
One comment, but otherwise LGTM.

src/Dapr.Client/DaprClientGrpc.cs
WhitWaldo: I _think_ you're right. This was more just an excess of caution in case I wasn't right about that. I can remove it if you think it unnecessary.

philliphoff (approved):

bhanuprakash-1:
How do we know in which SDK version(or when) this fix will be available? I'm using the 1.14.0 version (latest available as of today based on this: https://www.nuget.org/packages/Dapr.Client).

WhitWaldo:
@bhanuprakash-1 This will be available as part of the 1.15 client release available in January.

Issue #1311, opened 2024-06-20
Non-Dapr endpoints invocation supported?


ZenwalkerD:
## Ask your question here
Dear All,

In the Dapr documentation i noticed that a service can invoke a Non-Dapr endpoints like  https://reqres.in/api/users?page=2 (external APIs).

But when i checked the Dotnet Dapr documentation; i could not find any code related to that.

I tried below code and it does not work..

```
 var client = DaprClient.CreateInvokeHttpClient("https://reqres.in");
 var response = client.GetAsync(client.BaseAddress + "api/users?page=1").Result;
 var result = response.Content.ReadAsStringAsync().Result;
```

Please let me know if the support is planned or where can i find a documentation?

WhitWaldo:
Using Dapr for service invocation against Dapr endpoints gives you service discoverability. If you don't need the discoverability and know the port of what you're trying to access, it gives you a mechanism to attempt invoking a REST endpoint using an HTTP-based invocation. 

In your example, you're going a step further and accessing a completely external API - might I ask why you're bothering with the DaprClient and not just using an HttpClient to invoke it like you would any other REST endpoint?

philliphoff:
> I tried below code and it does not work..

@ZenwalkerD Can you be more specific about what didn't work?  Did the call fail, and in what way?  Could you provide a repro code/project that demonstrates the issue?

ZenwalkerD:
> In your example, you're going a step further and accessing a completely external API - might I ask why you're bothering with the DaprClient and not just using an HttpClient to invoke it like you would any other REST endpoint?

Because of Daprs feature of resiliency. Dapr provides retries and other policies. So i was hoping to make use of all those features in invoking external network calls. Otherwise i have to rely on some libraries like Polly.net.

ZenwalkerD:
> > I tried below code and it does not work..
> 
> @ZenwalkerD Can you be more specific about what didn't work? Did the call fail, and in what way? Could you provide a repro code/project that demonstrates the issue?
This is what i am getting when i deploy my app on Azure ACA:

`{"errorCode":"ERR_DIRECT_INVOKE","message":"failed to invoke, id: https, err: rpc error: code = Unimplemented desc = "}`

Code is very simple. Plain ASP NET CORE WEB API project with a GET controller having above provided code. Assuming that all dependencies are added for Dapr and others.

philliphoff:
[...]

Issue #1352, opened 2024-09-27
Unable to pass Authorization header


JulianJJ:
## Expected Behavior

When I pass Authorization header I should receive in the target API just like dapr api key

## Actual Behavior

Authorization header is not received in the target API

WhitWaldo:
Could you share some more information about what you're calling and how? Bonus points if you could share a simple repro project!

WhitWaldo:
@JulianJJ I put together my own sample project and was unable to reproduce what you're seeing. Again, your own repro would be helpful, but in lieu of that, could you answer a couple of questions for me:
- Could you confirm that you're using `InvokeMethodAsync` and passing it an `HttpRequestMessage` decorated with the Authorization header?
- Does the receiving API receive the request at all?
- Do you see any headers on the target API? Is it only missing the Authorization header?

WhitWaldo:
@JulianJJ Thank you for the additional details. I'll circle back should I need any other information.

WhitWaldo:
@JulianJJ Were you able to put a repro project together for this bug?

JulianJJ:
@WhitWaldo Not yet.. Will let you know soon

WhitWaldo:
@JulianJJ Circling back here - is this still something you're experiencing an issue with?

susin03:
As additional info, I'm running the apps with dotnet Aspire with this code:

builder.AddProject<Projects.ProjectA>("projectA")
    .WithDaprSidecar(appId: "projectA");

builder.AddProject<Projects.ProjectB>("projectB")
    .WithDaprSidecar(appId: "projectB");

susin03:
Hey! So, the following happened: I debugged the whole request path, and was able to observe that the headers were still present during the last interaction in my AspNet app.
When the request is made via Dapr (Windows, running in Docker) the Authorization Header is removed.

To be sure of that, I just deployed my app in my Kuberntes environment and the same AspNet code worked as intended.

I don't know how to proceed with the issue now. Thanks in advance.

JulianJJ:
Yes that is correct and I think its not yet fixed. This is very problematic.

WhitWaldo:
> Hey! So, the following happened: I debugged the whole request path, and was able to observe that the headers were still present during the last interaction in my AspNet app. When the request is made via Dapr (Windows, running in Docker) the Authorization Header is removed.
> 
> To be sure of that, I just deployed my app in my Kuberntes environment and the same AspNet code worked as intended.
[...]

PR #1354, opened 2024-10-06
Removed invalid service invocation check when using HttpRequestMessage

WhitWaldo:
# Description

While building out a repro project to investigate #1352, I was unable to validate that (awaiting further information), but did find a bug in the client when a service invocation involved an `HttpRequestMessage`. It would incorrectly validate the host/port of the destination request with the local HTTP client endpoint for a match before proceeding - this would likely never work unless it were calling an endpoint on its own service.

## Issue reference

Please reference the issue this PR will close: #1311 

## Checklist

* [X] Code compiles correctly
* [N/A] Created/updated tests - No existing tests to update

### DIFF
--- a/src/Dapr.Client/DaprClientGrpc.cs
+++ b/src/Dapr.Client/DaprClientGrpc.cs
@@ -423,12 +423,7 @@ namespace Dapr.Client
         public override async Task<HttpResponseMessage> InvokeMethodWithResponseAsync(HttpRequestMessage request, CancellationToken cancellationToken = default)
         {
             ArgumentVerifier.ThrowIfNull(request, nameof(request));
-
-            if (!this.httpEndpoint.IsBaseOf(request.RequestUri))
-            {
-                throw new InvalidOperationException("The provided request URI is not a Dapr service invocation URI.");
-            }
-
+            
             // Note: we intentionally DO NOT validate the status code here.
             // This method allows you to 'invoke' without exceptions on non-2xx.
             try


WhitWaldo:
@philliphoff Could you give this a quick look when you have a moment? Fix for an as-yet unreported bug discovered while working on another.

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -423,12 +423,7 @@ public override HttpRequestMessage CreateInvokeMethodRequest<TRequest>(HttpMetho
         public override async Task<HttpResponseMessage> InvokeMethodWithResponseAsync(HttpRequestMessage request, CancellationToken cancellationToken = default)
         {
             ArgumentVerifier.ThrowIfNull(request, nameof(request));
-
-            if (!this.httpEndpoint.IsBaseOf(request.RequestUri))
```
philliphoff, line 427: I *think* this is intentional.  The `HttpRequestMessage` (from the `DaprClient` docs) "must be a conforming Dapr service invocation request".  This implies that the base `RequestUri` would be the Dapr sidecar host/port (as created by the `CreateInvokeMethodRequest()` method).

WhitWaldo: For the life of me, I cannot remember my scenario for why I thought this was out of date. I'm good closing this until it comes to mind again.

Issue #1115, opened 2023-06-22
Make AddDaprSecretStore compliant with DI


ChristPetitjean:
## Describe the feature
The actual implementation need a DaprClient instance to work in API. But, because the mapping is performed somewhere before the app is built, we cannot use `app.Services.GetRequiredService<DaprClient>`. So this make this extension method using a new instance that is unmockable and so unusable during testing phase.

For example:
Program.cs:
```csharp
      builder.Configuration.AddDaprSecretStore("secretstore", new DaprClientBuilder().Build());
```

TestFile.cs
```csharp
            var daprClientMock = new Mock<DaprClient>();
            daprClientMock
                .Setup(s => s.GetSecretAsync(
                           It.IsAny<string>(), 
                           It.IsAny<string>(), 
                           It.IsAny<IReadOnlyDictionary<string, string>>(), 
                           It.IsAny<CancellationToken>()))
                .ReturnsAsync(new Dictionary<string, string>
                {
                    { "ConnexionStrings:mongodb", this._mongoDbRunner.ConnectionString }
                });
            using var client = Application
                .WithWebHostBuilder(
                    builder =>
                    {
                        builder.ConfigureTestServices(
                            services =>
                            {
                                services.AddSingleton(_ => daprClientMock.Object);
                            });
                    })
                .CreateClient();
```

This code will not work because of the extension method that is creating its own instance instead of using the DI one.

It will be great to use DI for this method.
Workaround is to not use it at all and only use secretAsync method from the client.

PR #1364, opened 2024-10-15
Support DaprClient injection into AddDaprSecretStore registration

WhitWaldo:
# Description

Added an overload that accepts an `IServiceProvider` so that the `DaprClient` that's DI registered can be injected into the store configuration registration so it needn't be built from the static method.

## Issue reference

Please reference the issue this PR will close: #1115 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


WhitWaldo:
I'm unable to find any documentation that illustrates usage for any of the extension registrations in this project to augment to reflect this, so that might be a bit of a larger task to add somewhere.

WhitWaldo:
After discussing further with @philliphoff I think this PR is unnecessary and can be resolved in favor of richer documentation and existing overloads.

PR #1365, opened 2024-10-15
Removes floating classes and introduces Dapr.Common project

WhitWaldo:
# Description

There are a number of classes that have been floating around in "src/Shared" and referenced by the various packages. Another PR #1363 seeks to use a type available in a NuGet package in an update to one of these types necessitating they be in a project now.

It also pulls `DaprException` from the `Dapr.Client` project and makes it available more commonly for still other PRs (#1331)

Doing this here so as to make the impact of the PR in #1331 significantly smaller and opens the door to future smaller PRs (e.g. removing the dependency on `ArgumentVerifier` throughout in favor of the null argument exception API that's built-in).

This should not introduce a breaking change as I updated the namespace for each type to use `Dapr` as they originally were assigned to. It's just a referential change in each .csproj from a floating file to a referenced project.

## Issue reference

Please reference the issue this PR will close: None

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


philliphoff (approved):
LGTM

philliphoff (approved):

PR #1379, opened 2024-10-24
Small refactoring on dapr client and invocation handler. Add improvements on test structure.

RafaelJCamara:
# Description

- Small refactor to improve code readability (both on dapr client and the invocation handler)
- Improve test structure by removing the test definition from the test class itself to a test data class, therefore doing a small separation of concerns between test data definition and test data usage.

## Issue reference

No issues to be referenced.

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


test/Dapr.Client.Test/TestData/InvocationHandlerTestsData.cs
```diff
@@ -0,0 +1,56 @@
...
+
+namespace Dapr.Client.Test.TestData;
+
+public class TryRewriteUriWithNoAppIdRewritesUriToDaprInvokeTestData : TheoryData<string, string, string>
+{
+    public TryRewriteUriWithNoAppIdRewritesUriToDaprInvokeTestData()
+    {
+        Add(null, "http://bank", "https://some.host:3499/v1.0/invoke/bank/method/");
+        Add("bank", "http://bank", "https://some.host:3499/v1.0/invoke/bank/method/");
+        Add("Bank", "http://bank", "https://some.host:3499/v1.0/invoke/Bank/method/");
```
RafaelJCamara, line 11: If someone could please clarify me the intention behind testing this with a lowercase and uppercase B, that would be amazing, because most of this test data is duplicated and just changes on this casing.

WhitWaldo: It's proving that the operation works in a case-insensitive manner.

RafaelJCamara: Just as I suspected. Thanks a lot!

test/Dapr.Client.Test/InvocationHandlerTests.cs
```diff
@@ -79,47 +83,10 @@ public void TryRewriteUri_FailsForRelativeUris()
         }
 
         [Theory]
-        [InlineData(null, "http://bank", "https://some.host:3499/v1.0/invoke/bank/method/")]
```
WhitWaldo, line 82: I disagree with removing all these tests - again, they're each verifying a unique and separate variation of the inputs that is worth knowing is covered.

test/Dapr.Client.Test/TestData/InvocationHandlerTestsData.cs
```diff
@@ -0,0 +1,56 @@
+﻿using Xunit;
```
WhitWaldo, line 1: Please put these back with the inline data attributes. When building unit tests, I don't expect to have to look in a separate file to add additional variations, I expect to see them inline using the convenience attributes provided by the framework. This might make sense in an actual library, but not in a unit test class.

src/Dapr.Client/InvocationHandler.cs
```diff
@@ -98,63 +96,68 @@ internal string? DaprApiToken
         /// <inheritdoc />
         protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
         {
-            var original = request.RequestUri;
-            if (!this.TryRewriteUri(request.RequestUri, out var rewritten))
+            var originalUri = request.RequestUri;
```
WhitWaldo, line 99: I'm disinclined to accept this change which only changes the inline variable. If you're rewriting the method to do something and you need to change the variable so it's more meaningful, by all means, but to change it for the sake of changing it isn't necessary.

src/Dapr.Client/InvocationHandler.cs
```diff
@@ -98,63 +96,68 @@ internal string? DaprApiToken
...
             {
-                request.RequestUri = original;
+                request.RequestUri = originalUri;
                 request.Headers.Remove("dapr-api-token");
             }
         }
 
         // Internal for testing
-        internal bool TryRewriteUri(Uri? uri, [NotNullWhen(true)] out Uri? rewritten)
+        internal bool TryRewriteUri(Uri? uri, [NotNullWhen(true)] out Uri? rewrittenUri)
```
WhitWaldo, line 127: Same thing - I'm disinclined to accept this change that renames the argument just because you can. Again, if you're changing the method itself and need to change the name so it's useful in your rewrite, that's one thing, but changing it just for the sake of doing so here, especially for something that's strictly internal for testing purposes, doesn't make sense.

test/Dapr.Client.Test/DaprClientTest.cs
```diff
@@ -29,11 +29,7 @@ public void CreateInvokeHttpClient_WithAppId()
         [Fact]
         public void CreateInvokeHttpClient_InvalidAppId()
         {
-            var ex = Assert.Throws<ArgumentException>(() => 
-            { 
-                // The appId needs to be something that can be used as hostname in a URI.
```
WhitWaldo, line 34: I think it's useful to retain this comment in the test as it provides more context to the test name and what it's covering

WhitWaldo (changes requested):
Thank you for your pull request. I appreciate the effort you’ve put into it. However, I noticed several changes that seem unnecessary, particularly the renaming of arguments without a clear reason. As mentioned in my comments, we prefer to avoid such changes unless they are essential, as our XML documentation already provides clarity.

If a new feature requires renaming for better functionality or readability, that’s understandable. However, changes should not be made solely for the sake of change. Please revert these specific changes, and in future pull requests, ensure that any renaming is justified by a significant need.

test/Dapr.Client.Test/InvocationHandlerTests.cs
RafaelJCamara: I'm not removing them, I just placed them in the class data _**TryRewriteUriWithNoAppIdRewritesUriToDaprInvokeTestData**_ and all of them are being called here.

The reason why I've done this was to make this testing class a bit more cleaner.

If you don't agree with moving the test data to that class, I can revert this change.

RafaelJCamara: Which I've now seen you want to revert back to the attributes.

RafaelJCamara:
@WhitWaldo I was looking at the codebase to understand how somethings worked and I thought it was a good idea to suggest some improvements.

I see that most of my suggested changes are proposed for revert. Do you think it still makes sense to have the rest of the changes up or shall we have put this PR to sleep?

WhitWaldo:
> @WhitWaldo I was looking at the codebase to understand how somethings worked and I thought it was a good idea to suggest some improvements.
> 
> I see that most of my suggested changes are proposed for revert. Do you think it still makes sense to have the rest of the changes up or shall we have put this PR to sleep?


Thank you for your follow-up. I appreciate your understanding regarding the previous points. Several of the remaining changes in this PR appear to be formatting-based (e.g., tab inserts) or minor syntax updates (like converting a get body to a lambda expression). Generally, I would prefer not to accept these types of changes unless they are part of a larger feature addition.

I do see value in the following change in this PR:
```cs
var ex = Assert.Throws<ArgumentException>(() => 
{ 
    _ = DaprClient.CreateInvokeHttpClient(daprEndpoint: "ftp://localhost:3500");
});
```
to
```cs
var ex = Assert.Throws<ArgumentException>(() => DaprClient.CreateInvokeHttpClient(daprEndpoint: "ftp://localhost:3500"));
````
...and the others like it. If you could revert the other changes and keep this one, I'd be happy to accept this PR.

For future reference however, I'd prefer that PRs include more substantial feature additions or changes rather than consisting solely of small formatting and refactoring changes.

RafaelJCamara:
I'll close this one and perhaps open another one with such improvements.

Thanks a lot!

Issue #1385, opened 2024-10-27
Workflow DI Registration Bug


WhitWaldo:
## Expected Behavior

Demo should work as expected.

## Actual Behavior

As reported by a user in the Discord channel, the WorkflowConsoleApp quickstart is not working properly and showing the following error:
> Unable to resolve service for type `Microsoft.DurableTask.Client.DurableTaskClient` while attempting to activate `Dapr.Workflow.DaprWorkflowClient`.

Regression thought to be due to #1363 

Should also update the integration test as it isn't demonstrating any issue with this.

## Steps to Reproduce the Problem

Start quickstart app from latest bits in master.

## Release Note
RELEASE NOTE: **FIX** Bug during workflow client registration

PR #1386, opened 2024-10-27
Fix for DI registration not completing as expected

WhitWaldo:
# Description

#1363 changed how clients registered themselves under the hood with DI to make for a more uniform process. With Workflow, this meant switching to a factory that attempts to load values from an injected IConfiguration.

## Fix breakdown
- Updated to make `IConfiguration` injection optional as the caller might not be using ASP.NET Core and thus may not have actually registered this
- Updated to make `IHttpClientFactory` optional because while it's registered as part of using `AddWorkflowClient`, the caller might not be using that route (though they should)
- When the IConfiguration change was originally made, the idea was to use a factory that the `IConfiguration` and `IHttpClientFactory` are injected to as those would be resolved and used as part of the registration of the durable task functionality. Unfortunately, two key steps were overlooked: The factory itself was never registered and the method performing the necessary additional registrations was never invoked because it lived in a DI registration that would never resolve to invoke it. Corrected this by registering the factory as intended and by creating a scoped provider during registration that resolves and invokes the method ensuring that the types are registered on the `IServiceCollection` as intended.

## Issue reference

Please reference the issue this PR will close: #1385 

## Checklist

* [X] Code compiles correctly
* [ ] Created/updated tests - WIP


src/Dapr.Workflow/DaprWorkflowClientBuilderFactory.cs
```diff
@@ -28,27 +28,25 @@ namespace Dapr.Workflow;
...
         _httpClientFactory = httpClientFactory;
-        _services = services;
     }
     
     /// <summary>
     /// Responsible for building the client itself.
     /// </summary>
     /// <returns></returns>
-    public void CreateClientBuilder(Action<WorkflowRuntimeOptions> configure)
+    public void CreateClientBuilder(IServiceCollection services, Action<WorkflowRuntimeOptions> configure)
```
philliphoff, line 47: It's not obvious to me why `DaprWorkflowClientBuilderFactory` itself needs to be registered with the container as opposed to just making this `CreateClientBuilder()` method an `IServiceCollection` extension method.  Is it because `AddDurabletaskClient()` needs a registered `IHttpClientFactory` and `IConfiguration`?  If so, you *should* be able to get them using the `builder.Services`.

philliphoff (changes requested):
I wonder if we might simplify the registration here a bit.

src/Dapr.Workflow/DaprWorkflowClientBuilderFactory.cs
WhitWaldo: You're right - that was an artifact of having put together the fix that I neglected to remove. I've removed that registration and am resolving the services myself to create a new instance within the service provider scope.

philliphoff (commented):
Needs fix for build break, but otherwise looks good.  Wish we didn't have to create a services scope to pull dependencies from the container, but I don't see a way to defer it any further.

philliphoff (approved):

PR #1401, opened 2024-11-05
Added source generator for actor type registration

WhitWaldo:
# Description

Using the .NET SDK today, one must register each actor type with dependency injection like so:

```cs
services.AddActors(options => {
  options.Actors.RegisterActor<MyActorA>();
  options.Actors.RegisterActor<MyActorB>();
  options.Actors.RegisterActor<MyActorC>();
  options.Actors.RegisterActor<MyActorD>();
});
```

As one uses an increasing number of actor types, this can get cumbersome to deal with maintaining this list of types. That's where this source generator comes into play. If all the actors are in the same project, the developer need only do the following:

```cs
services.RegisterAllActors();
```

And if the developer is interested in also including actor types from any assemblies referenced by the project as well, they need only indicate as much, as follows:

```cs
services.RegisterAllActors(true);
```

## Issue reference

Please reference the issue this PR will close: N/A

This was the result of a conversation between @philliphoff and myself recently.

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


WhitWaldo:
Superseded by #1441

Issue #1412, opened 2024-11-22
NullReferenceException when trying to sub with streaming subscriptions


Aimless321:
## Expected Behavior

Receive messages in the registered handler when using streaming subscriptions.

## Actual Behavior

Running version 1.14 of the Dapr runtime and using the master branch of the .NET SDK, i receive a NullRefrenceException when trying to subscribe to a topic using the new Messaging Client.

The call to `messagingClient.SubscribeAsync` never returns, and gives me an error which i traced back to a NullReferenceException on `response.EventMessage` in `PublishSubscribeReceiver.cs` line 252.
```
== APP ==       System.NullReferenceException: Object reference not set to an instance of an object.
== APP ==          at Dapr.Messaging.PublishSubscribe.PublishSubscribeReceiver.FetchDataFromSidecarAsync(AsyncDuplexStreamingCall`2 stream, ChannelWriter`1 channelWriter, CancellationToken cancellationToken)
== APP ==          at Dapr.Messaging.PublishSubscribe.PublishSubscribeReceiver.FetchDataFromSidecarAsync(AsyncDuplexStreamingCall`2 stream, ChannelWriter`1 channelWriter, CancellationToken cancellationToken)
== APP ==          at Dapr.Messaging.PublishSubscribe.PublishSubscribeReceiver.SubscribeAsync(CancellationToken cancellationToken)
== APP ==          at Dapr.Messaging.PublishSubscribe.DaprPublishSubscribeGrpcClient.SubscribeAsync(String pubSubName, String topicName, DaprSubscriptionOptions options, TopicMessageHandler messageHandler, CancellationToken cancellationToken)
== APP ==          at Worker.Services.WorkerService.ExecuteAsync(CancellationToken stoppingToken) in C:\Users\WesleyB\Projects\StrykerOrchestrator\Worker\Services\WorkerService.cs:line 37
== APP ==          at Microsoft.Extensions.Hosting.Internal.Host.TryExecuteBackgroundServiceAsync(BackgroundService backgroundService)
```

In WireShark i can see the GRPC call to the sidecar, and i also receive logging from the sidecar:
```
time="2024-11-22T13:54:48.3234217+01:00" level=info msg="Subscribing to pubsub 'task-queue' topic 'myTopic'" app_id=worker scope=dapr.runtime.pubsub.streamer type=log ver=1.14.4
[...]

PR #1415, opened 2024-11-28
Fixes + unit tests for streaming PubSub implementation

WhitWaldo:
# Description

Fixed the linked issue which called out the following problems:
- Sometimes inbound messages from the Dapr runtime are null which cause a NullReferenceError during delegate processing. This has been updated to add a null check and skip such messages.
- The implementation used an `await Task.WhenAny` which caused the whole of the operation to be blocking. This was unintentional and has been corrected.

While fixing these, I realized that the project was devoid of several useful unit tests that could ensure that the rest of the class was operating as it should, so I took the opportunity to add a few internal fields and methods (available only to the test project) that would assist in some of this validation.

Finally, I modified the `AddDaprPubSubClient` extension method to pass an `IConfiguration?` into the `DaprPublishSubscribeClientBuilder` so it can optionally source the various configuration-based values from there during registration.

## Issue reference

Please reference the issue this PR will close: #1412 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [N/A] Extended the documentation - done in another issue


src/Dapr.Messaging/PublishSubscribe/PublishSubscribeReceiver.cs
```diff
@@ -115,20 +120,38 @@ internal async Task SubscribeAsync(CancellationToken cancellationToken = default
...
+    {
+        await acknowledgementsChannel.Writer.WriteAsync(acknowledgement);
+    }
+
+    private static void HandleTaskCompletion(Task task, object? state)
+    {
+        if (task.Exception != null)
         {
-            // Will be cleaned up during DisposeAsync
+            throw task.Exception;
```
philliphoff, line 154: I haven't kept up with the exception behavior of the latest .NET runtime; will this get handled?

philliphoff (dismissed):

src/Dapr.Messaging/PublishSubscribe/PublishSubscribeReceiver.cs
WhitWaldo: It _should_ wrap any exception in an `AggregateException` that can be caught by any outer caller since there aren't any try/catch blocks that'll transparently handle it internally.

philliphoff (dismissed):

philliphoff (approved):

PR #1417, opened 2024-12-03
Jobs registration update

WhitWaldo:
# Description

I updated the Jobs client builder factory to achieve some uniformity with the other client builders.

This approach means we don't have to do either of the following:
- Pass a whole `Action<DaprClientBuilder>?` into every `AddXClient()` method 
- Register a single `DaprClient` that's used by all Dapr clients for a singular means of configuring Dapr clients 

Rather, this approach means each application can drop completely different values in their local `IConfiguration` instances and they'll be picked up when each Dapr client is resolved.

## Issue reference

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


WhitWaldo:
This PR is no longer necessary as most of it was included in #1382

Issue #1416, opened 2024-12-03
Actor source generator generates invalid code for generic interfaces


WhitWaldo:
Actor source generator (dotnet): Dapr.Actors.Generators.ActorClientGenerator
1.14.x

## Expected Behavior
Given the following actor interface:
```cs
[GenerateActorClient]
    public interface IGenericActor<in TFoo> : IActor
    {
        [ActorMethod]
        Task DoSomethingAsync(TFoo foo);
    }
```
should emit a class like:
```cs
    public sealed class GenericActorClient<TFoo> : ActorServiceTest.IGenericActor<TFoo>
    {
        // ... body
    }
```

## Actual Behavior

The source generator emits:
```cs
public sealed class GenericActorClient : ActorServiceTest.IGenericActor<TFoo>
    {
        // ... the body is okay
    }
```

which is invalid.

## Steps to Reproduce the Problem

Code above is enough to repro. Example the output of the generator.

## Release Note
RELEASE NOTE: **FIX** Bug in actor source generator

PR #1419, opened 2024-12-04
FIX: Actor source generator generates invalid code for generic interfaces

m3nax:
REF: https://github.com/dapr/dotnet-sdk/issues/1416


m3nax:
Failed test isn't related with the pr https://github.com/dapr/dotnet-sdk/actions/runs/12167263263/job/33935403480?pr=1419#step:18:41

WhitWaldo:
@m3nax Wouldn't we expect that the generated code would include most of the using statements of the original source? Your tests all pass, but I'm concerned that the output wouldn't actually work in the real world for lack of a reference to `Dapr.Actors`, et all.

WhitWaldo (approved):
Looks good to me! Thank you very much for your contribution!

PR #1434, opened 2024-12-19
Adds workflow replay-safe logger

WhitWaldo:
# Description

By popular demand in Discord and after realizing that a lot of the plumbing was already available in the [Durable Task SDK](https://github.com/microsoft/durabletask-dotnet/tree/main) itself, I've implemented the version of the ReplaySafe logger used in that version (as opposed to the version in the Azure Functions Durable Task (https://github.com/Azure/azure-functions-durable-extension/tree/dev/src/WebJobs.Extensions.DurableTask). I made a few tweaks to the visibility on the context itself so conform to how Dapr implements the SDK, but nothing major.

As a result, there's a method available on the `WorkflowContext` for `CreateReplaySafeLogger` that will provision an `ILogger` using the category name or the provided type of the context (provided either as an argument or a generic type) from the `ILoggerFactory` provisioned within the Durable Task SDK itself and its various registrations.

I updated one of the Workflow examples to include the intended usage.

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


src/Dapr.Workflow/DaprWorkflowContext.cs
```diff
@@ -94,6 +96,25 @@ public override Guid NewGuid()
...
         }
+        
+        /// <summary>
+        /// Returns an instance of <see cref="ILogger"/> that is replay-safe, meaning that the logger only
+        /// writes logs when the orchestrator is not replaying previous history.
+        /// </summary>
+        /// <param name="categoryName">The logger's category name.</param>
+        /// <returns>An instance of <see cref="ILogger"/> that is replay-safe.</returns>
+        public override ILogger CreateReplaySafeLogger(string categoryName) =>
+            new ReplaySafeLogger(this, this.innerContext.CreateReplaySafeLogger(categoryName));
```
philliphoff, line 107: Is there a reason not to just passthrough the `ILogger` created by the inner context?  Since we return an `ILogger` as well, we could always wrap it in the future if/when necessary.

WhitWaldo: I don't see that the inner context (`TaskOrchestrationContext`) creates an `ILogger`. I modeled this after [how it creates](https://github.com/microsoft/durabletask-dotnet/blob/c8cb34c0bbb21dcab24dec103febb1493726f7f2/src/Abstractions/TaskOrchestrationContext.cs#L388) its own `ReplaySafeLogger` (private to `TaskOrchestrationContext` or I'd just expose it) from its `ILoggerFactory`.

Further, in case I am missing it (very possible), by letting the user provide their own workflow type or category name, it would let them decorate the category with a value that makes sense for their situation as opposed to having an `ILogger<TaskOrchestrationContext>` which could make resolution of activities to their actual workflows difficult when running back through traces.

WhitWaldo: I typed up a whole response and then realized what it is you said. You're right - no point having a separate ReplaySafeLogger and creation method when I can use the virtual methods right off the TaskOrchestrationContext.

philliphoff (approved):

Issue #1426, opened 2024-12-15
Add analyzers that help prevent developer gotchas


WhitWaldo:
## Describe the feature
This is intended more as a parent issue to track progress for individual analyzers, but the goal is that we reduce some of the opportunities for developers to inadvertently misconfigure things that we might catch at build time and reduce runtime exceptions. Here's a short and evolving list of opportunities here:

- **Actor cross-serialization errors** Ensure that when developers retain DataContract serialization or opt into JSON-based serialization, they're not inadvertently cross-serializing in a way that breaks the serializers and deserializers for either approach. 
- **Identify workflows and workflow activities that are referenced, but not registered in DI**
- **Identify actors that are referenced, but not registered in DI**
- **Ensure that during startup, those resources that require a `app.UseX()` have it specified in Program.cs**

## Release Note
RELEASE NOTE: ADD Analyzers to help prevent runtime errors

jev-e:
I'd be interested in looking at this; although partially motivated by self improvement as my knowledge of Roslyn analysers isn't super strong, but an area I'd like to improve.

siri-varma:
Planning to add an analyzer to make sure `MapDaprScheduledJobHandler` is in place to receive invocations for the jobs api. Let me know what you folks think ?

WhitWaldo:
> Planning to add an analyzer to make sure `MapDaprScheduledJobHandler` is in place to receive invocations for the jobs api. Let me know what you folks think ?

I didn't realize analyzers can look up installed packages. If that's the case, sounds like a great idea.

Could take it a step farther, though a future release might render it obsolete. Because only the app that schedules a job can receive job invocations, you could do a scan for all the job names that are registered in the app and validate there's a corresponding mapped handler for each.

ngruson:
> > Can we add an analyzer that checks the input and output argument types when calling activities from the workflow? And if the activity class name matches? See the [DurableTaskAnalyzers](https://github.com/Azure/azure-functions-durable-extension/tree/dev/src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers).
> 
> I added the former to the list. Can you get into more detail on the latter ask and how it differs from "Identify workflows and workflow activities that are referenced, but not registered in DI" which is already in the table?
> 
> I'll go through the linked analyzers and get them added as well. Good find!

I also thought of this working on a workflow over the weekend.
Another addition could be: when you call an activity, does the input object type match the input type of the activity?
[...]

PR #1440, opened 2025-01-09
Add Roslyn analyzer for workflows

ngruson:
# Description

This PR adds a Roslyn analyzer that checks if workflows and activities are registered properly.
For workflows, the analyzer checks if a workflow is referenced in `DaprWorkflowClient.ScheduleNewWorkflowAsync` calls (other logic may be added to determine if a workflow is referenced?).
For workflow activities, the analyzer checks if activities are referenced in workflows in `WorkflowContext.CallActivityAsync` calls.

For both scenarios, code fix providers have been added to add `RegisterWorkflow` or `RegisterActivity` calls to `AddDaprWorkflow` code.

We have to come up with a general approach for Roslyn analyzers. Maybe we can move these remarks to the issue.
- Class library per building block (i.e. `Dapr.Workflow.Analyzers`).
- Define a range for diagnostic rule IDs. For now, I have used DAPR1001 and DAPR1002 for the diagnostics described above. For each building block, we might define a different range for the first digit following DAPR.
- The analyzers should be included in the NuGet package for the building block (`Dapr.Workflow`).

Other open questions:
- If  a workflow and its activities are defined in a class library, it should not require workflow and activity registrations to be in the class library. This is usually done in the hosted application.

## Issue reference

This PR belongs to issue #1426.

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


WhitWaldo:
Pointing to merge to `release-1.16` instead of `master`.

WhitWaldo:
Thank you very much for your contribution @ngruson !

I particularly like the classes you used in this and your other PR for validating the analyzers and codefixes and have refactored them out to a separate project to use with all the analyzer tests. Thanks for putting it all together.

PR #1441, opened 2025-01-11
Add Roslyn analyzer for actor registration

ngruson:
# Description

This PR adds a Roslyn analyzer that checks if actors are registered properly.
For every actor in the project, the analyzer checks if there's a corresponding `RegisterActor` call in the application setup.

If there's no `AddActors` call yet, it will add it as well. The analyzer will look for `WebApplication.CreateBuilder` and adds the actor registration directly below it.

- Introduced `Dapr.Actors` package reference.
- Added `Microsoft.CodeAnalysis.CSharp.Analyzer.Testing` package reference (v1.1.2).
- Created `Dapr.Actors.Analyzers` and `Dapr.Actors.Analyzers.Test` projects.
- Implemented `ActorRegistrationAnalyzer` to warn about unregistered actors.
- Developed `ActorRegistrationCodeFixProvider` to suggest actor registration.
- Added unit tests for both the analyzer and code fix provider.
- Enhanced test project with necessary references and utility methods.

## Issue reference

This PR belongs to issue https://github.com/dapr/dotnet-sdk/issues/1426.

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


src/Dapr.Actors.Analyzers/ActorAnalyzer.cs
```diff
@@ -0,0 +1,206 @@
...
+/// <summary>
+/// Analyzes actor registration in Dapr applications.
+/// </summary>
+[DiagnosticAnalyzer(LanguageNames.CSharp)]
+public class ActorAnalyzer : DiagnosticAnalyzer
+{
+    private static readonly DiagnosticDescriptor DiagnosticDescriptorActorRegistration = new(
+        "DAPR0001",
+        "Actor class not registered",
+        "The actor class '{0}' is not registered",
```
WhitWaldo, line 18: We'll likely need to revisit this in the future as there's consideration for facilitating cross-service actor invocation, so it might very well be the case that an actor is referenced that doesn't exist in the same service.

WhitWaldo:
@ngruson I've refactored your solution a bit to fit in with the current bits on 1.16 and to put the analyzers in separate files.

Otherwise, this looks good to go. Thank you very much for your contribution!

PR #1443, opened 2025-01-14
Bugfix: Crypto ReadOnlyMemory<byte> decryption times out

WhitWaldo:
# Description
While updating the Crypto example in preparation for creating a Quickstart, I realized that the `DecryptAsync` overload that accepted a `ReadOnlyMemory<byte>` timed out whenever it was used. I suspected the issue is related to the `MemoryMarshal`, which was being used to avoid an allocation. But reviewing the implementation, I realized that we were still getting that allocation later on when we created a new `MemoryStream` from the `MemoryMarshal` output. 

I've instead opted to create a `MemoryStream`, write the input to it and reset the position to 0 to use in the `EncryptAsync` and `DecryptAsync` operations. 

After making this change, both examples have been tested to work as expected (both stream and byte-based).

## Checklist

* [X] Code compiles correctly
* [N/A] Created/updated tests


src/Dapr.Client/Extensions/ReadOnlyMemoryExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+using System.IO;
+using System.Runtime.InteropServices;
+
+namespace Dapr.Client;
+
+internal static class ReadOnlyMemoryExtensions
+{
+    public static MemoryStream CreateMemoryStream(this ReadOnlyMemory<byte> memory, bool isReadOnly)
+    {
+        if (memory.IsEmpty)
```
philliphoff, line 24: So it was this missing check that allowed `TryGetArray()` to hang?  Strange.

WhitWaldo: Yeah, I'm not really sure why that would impact this unless there's some scenario where it the array is empty and it doesn't know how to create the MemoryStream (which doesn't seem right).

WhitWaldo: In fact, we have an awfully timely example - here's a run on a new E2E test I'm building out that uses all the examples as test sources: https://github.com/dapr/dotnet-sdk/actions/runs/12779350972/job/35623810397?pr=1444

It fails with a timeout when taking the overload this fix is intended to address, but again, not entirely sure what the disconnect is between the two and why this works, but that doesn't.

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -1670,24 +1670,18 @@ public override async Task<ReadOnlyMemory<byte>> EncryptAsync(string vaultResour
         ReadOnlyMemory<byte> plaintextBytes, string keyName, EncryptionOptions encryptionOptions,
         CancellationToken cancellationToken = default)
     {
-        if (MemoryMarshal.TryGetArray(plaintextBytes, out var plaintextSegment) && plaintextSegment.Array != null)
-        {
-            var encryptionResult = await EncryptAsync(vaultResourceName, new MemoryStream(plaintextSegment.Array),
-                keyName, encryptionOptions,
-                cancellationToken);
+        var memoryStream = plaintextBytes.CreateMemoryStream(true);
```
philliphoff, line 1673: `using var` ?

philliphoff (approved):
Just the `using` comment, but otherwise looks good.

src/Dapr.Client/DaprClientGrpc.cs
WhitWaldo: Done

philliphoff (approved):

PR #1477, opened 2025-02-28
Add a Job Analyzer to check if scheduled jobs have corresponding invocation urls

siri-varma:
# Description

Added a Roslyn Analyzer that checks if every scheduled job has a Job invocation url.

The analyzer checks for patterns like

* /job/123
* /job/{jobName}

## Issue reference

Please reference the issue this PR will close: #1426 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests

Thank you @ngruson for the initial PRs. It served as a great reference


all.sln
```diff
@@ -155,6 +155,10 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "JobsSample", "examples\Jobs
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Dapr.Workflow.Test", "test\Dapr.Workflow.Test\Dapr.Workflow.Test.csproj", "{E90114C6-86FC-43B8-AE5C-D9273CF21FE4}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Dapr.Jobs.Analyzer", "src\Dapr.Jobs.Analyzer\Dapr.Jobs.Analyzer.csproj", "{28B87C37-4B52-400F-B84D-64F134931BDC}"
```
WhitWaldo, line 158: Could you please rename the projects as `Dapr.Jobs.Analyzers` and `Dapr.Jobs.Analyzers.Tests` so they're consistent with other analyzer projects?

src/Dapr.Jobs.Analyzer/DaprJobsAnalyzerAnalyzer.cs
```diff
@@ -0,0 +1,118 @@
...
+    /// <summary>
+    /// DaprJobsAnalyzerAnalyzer.
+    /// </summary>
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class DaprJobsAnalyzerAnalyzer : DiagnosticAnalyzer
+    {
+        private static readonly DiagnosticDescriptor DaprJobHandlerRule = new DiagnosticDescriptor(
+            id: "DAPRJOBS0001",
+            title: "Ensure Post Mapper handler is present for all the Scheduled Jobs",
+            messageFormat: "There should be a mapping post endpoint for each scheduled job to make sure app receives notifications for all the scheduled jobs",
```
WhitWaldo, line 19: Could we reword this to be something like "Job invocations require the `MapJobInvocationHandler` be set and configured for each anticipated job on `IEndpointRouteBuilder`" with something similar for the title?

siri-varma:
* In the Dapr Jobs client, we have MapDaprScheduledJobHandler. However, the analyzer still throws an error because it is unable to detect the MapPost call within the method. I think this is what you were mentioning about @WhitWaldo. If we need to address this scenario, I can add a separate analyzer that checks for the presence of MapDaprScheduledJobHandler.

src/Dapr.Jobs.Analyzer/DaprJobsAnalyzerAnalyzer.cs
```diff
@@ -0,0 +1,118 @@
...
+                {
+                    string jobName = literal.Token.ValueText;
+
+                    // Now, we will check for a corresponding endpoint route.
+                    CheckForEndpointRoute(context, jobName);
+                }
+            }
+        }
+
+        private static void CheckForEndpointRoute(SyntaxNodeAnalysisContext context, string jobName)
```
WhitWaldo, line 65: While this looks like it can determine if an invocation mapping is registered with `IEndpointRouteBuilder` or not, it's not actually necessarily if the application never registers a job. Especially as jobs are only invoked on the application that registers them, if it's never registering them, no mapping is necessary as it's never invoked.

Given that, might we look around the assembly and see if there's any use of `ScheduleJobAsync` on a `DaprJobClient` type first?

siri-varma:
Will sign the commits shortly

test/Dapr.Jobs.Analyzer.Test/DaprJobAnalyzerTest.cs
```diff
@@ -0,0 +1,420 @@
...
+    using Microsoft.Extensions.DependencyInjection;
+    using Microsoft.Extensions.Hosting;
+
+    public class DaprJobsAnalyzerAnalyzerTests
+    {
+
+#if NET6_0
+        private static readonly ReferenceAssemblies referenceAssemblies = ReferenceAssemblies.Net.Net60;
+#elif NET7_0
+        private static readonly ReferenceAssemblies referenceAssemblies = ReferenceAssemblies.Net.Net70;
```
WhitWaldo, line 18: .NET 7 isn't a supported target for Dapr.Jobs projects.

You can also drop .NET 6 as it will no longer be supported starting in 1.16

src/Dapr.Jobs.Analyzer/DaprJobsAnalyzerAnalyzer.cs
```diff
@@ -0,0 +1,118 @@
...
+
+namespace Dapr.Jobs.Analyzer
+{
+    /// <summary>
+    /// DaprJobsAnalyzerAnalyzer.
+    /// </summary>
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class DaprJobsAnalyzerAnalyzer : DiagnosticAnalyzer
+    {
+        private static readonly DiagnosticDescriptor DaprJobHandlerRule = new DiagnosticDescriptor(
```
WhitWaldo, line 16: Should this be made `internal static` so it's accessible by the test without duplicating the definition?

WhitWaldo (changes requested):
Couple of notes

test/Dapr.Jobs.Analyzer.Test/DaprJobAnalyzerTest.cs
```diff
@@ -0,0 +1,420 @@
...
+                    }
+                }";
+
+            var expectedDiagnostic = new DiagnosticResult(DaprJobHandlerRule);
+
+            await VerifyAnalyzerAsync(testCode, expectedDiagnostic);
+        }
+
+        [Fact]
+        public async Task AnalyzeJobSchedulerHandler_ShouldRaiseDiagnostic_WhenJobHasHasEndpointButDoesNotMapToJobName()
```
WhitWaldo, line 71: Seems like this sort of specificity wouldn't end well if the developer passes in a variable to register and subscribe without a constant string.

test/Dapr.Jobs.Analyzer.Test/DaprJobAnalyzerTest.cs
siri-varma: I was receiving errors when I tried to remove .NET 6. That is why I had to add it. Let me do some investigation

src/Dapr.Jobs.Analyzer/DaprJobsAnalyzerAnalyzer.cs
WhitWaldo: Ah, based on your tests, it looks like you're already doing this in line 52 - could you add a comment indicating that's what magic is happening there?

WhitWaldo (changes requested):
Few more thoughts

test/Dapr.Jobs.Analyzer.Test/DaprJobAnalyzerTest.cs
WhitWaldo: I haven't fully removed the .NET 6 support throughout so perhaps that's why it's bugging you on it.

src/Dapr.Jobs.Analyzer/DaprJobsAnalyzerAnalyzer.cs
siri-varma: https://github.com/dapr/dotnet-sdk/pull/1477/files#diff-95a4bc519fc6b5c49452acf4c17caa549c2f54e1e114de4431bf24ef32d51c3bR52

Thats right. If you see the line 52 line there is an if check. We first determine if `ScheduleJobAsync ` is present as you mentioned and then only call the `CheckForEndpointRoute`

test/Dapr.Jobs.Analyzer.Test/DaprJobAnalyzerTest.cs
siri-varma: In addition, string interpolation and concatenation are other cases where this approach may not work. If we want to avoid this complexity, we could simplify the analyzer to check only for the usage of MapDaprScheduledJobHandler. However, this approach has a caveat—we would still generate warnings if developers use IEndpointRouteBuilder directly.

A better approach might be to check for both IEndpointRouteBuilder and MapDaprScheduledJobHandler. This ensures that a mapping exists for the job while also acknowledging that this approach will have some false positive warnings.

I will try to think if there are other approaches available.

siri-varma: The number of places from where the MapPost endpoint can be populated is not exhaustive. It can come from a Config File, other nugets, and string concatenation and interpolations. Going down the current route will lead to large number of False positive which as we were discussing is not a good experience.

I am going to resort back to just checking if `MapDaprScheduledJobHandler` is being called when "ScheduleJobAsync" is present.

siri-varma: Pushed new changes

src/Dapr.Jobs.Analyzer/DaprJobsAnalyzerAnalyzer.cs
siri-varma: done

test/Dapr.Jobs.Analyzer.Test/DaprJobAnalyzerTest.cs
siri-varma: Once the .NET 6 changes are in, I will remove the .NET if

WhitWaldo:
Changing the base to merge into `release-1.16` instead of `master`.

src/Dapr.Jobs.Analyzer/AnalyzerReleases.Shipped.md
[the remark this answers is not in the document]
WhitWaldo: I'll write up some docs somewhere for the 1.16 release documenting each of them.

src/Dapr.Jobs.Analyzer/MapDaprScheduledJobHandlerAnalyzer.cs
```diff
@@ -0,0 +1,115 @@
...
+namespace Dapr.Jobs.Analyzer
+{
+    /// <summary>
+    /// DaprJobsAnalyzer.
+    /// </summary>
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public sealed class MapDaprScheduledJobHandlerAnalyzer : DiagnosticAnalyzer
+    {
+        private static readonly DiagnosticDescriptor DaprJobHandlerRule = new DiagnosticDescriptor(
+            id: "DAPRJOBS0001",
```
WhitWaldo, line 16: Update to DAPR3001 per the above comment

src/Dapr.Jobs.Analyzer/MapDaprScheduledJobHandlerAnalyzer.cs
```diff
@@ -0,0 +1,115 @@
...
+                }
+            }
+        }
+
+        private static void CheckForEndpointRoute(SyntaxNodeAnalysisContext context, string jobName)
+        {
+            var root = context.SemanticModel.SyntaxTree.GetRoot();
+
+            // Search for MapPost with the corresponding route
+            int mapDaprScheduledJobHandlersCount = root.DescendantNodes()
```
WhitWaldo, line 68: This can be 'var' instead of 'int'

test/Dapr.Jobs.Analyzer.Test/MapDaprScheduledJobHandlerTest.cs
```diff
@@ -0,0 +1,264 @@
...
+                        using var scope = app.Services.CreateScope();
+
+                        var daprJobsClient = scope.ServiceProvider.GetRequiredService<DaprJobsClient>();
+
+                        daprJobsClient.ScheduleJobAsync(""myJob"", DaprJobSchedule.FromDuration(TimeSpan.FromSeconds(2)),
+                            Encoding.UTF8.GetBytes(""This is a test""), repeats: 10).GetAwaiter().GetResult();
+                    }
+                }";
+
+            var expectedDiagnostic = new DiagnosticResult("DAPRJOBS0001", DiagnosticSeverity.Warning);
```
WhitWaldo, line 53: ID update here as well - DAPR3001

test/Dapr.Jobs.Analyzer.Test/MapDaprScheduledJobHandlerTest.cs
```diff
@@ -0,0 +1,264 @@
...
+                        var builder = WebApplication.CreateBuilder();
+                        builder.Services.AddDaprJobsClient();
+                        var app = builder.Build();
+                        using var scope = app.Services.CreateScope();
+
+                        var daprJobsClient = scope.ServiceProvider.GetRequiredService<DaprJobsClient>();
+                    }
+                }";
+
+            var expectedDiagnostic = new DiagnosticResult("DAPRJOBS0001", DiagnosticSeverity.Warning);
```
WhitWaldo, line 84: And here - DAPR3001

test/Dapr.Jobs.Analyzer.Test/MapDaprScheduledJobHandlerTest.cs
```diff
@@ -0,0 +1,264 @@
...
+                        var daprJobsClient = scope.ServiceProvider.GetRequiredService<DaprJobsClient>();
+
+                        daprJobsClient.ScheduleJobAsync(""myJob"", DaprJobSchedule.FromDuration(TimeSpan.FromSeconds(2)),
+                            Encoding.UTF8.GetBytes(""This is a test""), repeats: 10).GetAwaiter().GetResult();
+                        daprJobsClient.ScheduleJobAsync(""myJob2"", DaprJobSchedule.FromDuration(TimeSpan.FromSeconds(2)),
+                            Encoding.UTF8.GetBytes(""This is a test""), repeats: 10).GetAwaiter().GetResult();
+                    }
+                }";
+
+            var expectedDiagnostic = new DiagnosticResult("DAPRJOBS0001", DiagnosticSeverity.Warning);
```
WhitWaldo, line 120: And.. here - DAPR3001

WhitWaldo (changes requested):
Change to the analyzer ID, otherwise looks good.

test/Dapr.Jobs.Analyzer.Test/MapDaprScheduledJobHandlerTest.cs
siri-varma: Done

test/Dapr.Jobs.Analyzer.Test/MapDaprScheduledJobHandlerTest.cs
siri-varma: Done

test/Dapr.Jobs.Analyzer.Test/MapDaprScheduledJobHandlerTest.cs
siri-varma: Done

src/Dapr.Jobs.Analyzer/MapDaprScheduledJobHandlerAnalyzer.cs
siri-varma: Done

src/Dapr.Jobs.Analyzer/AnalyzerReleases.Shipped.md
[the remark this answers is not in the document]
siri-varma: Done

test/Dapr.Jobs.Analyzer.Test/DaprJobAnalyzerTest.cs
siri-varma: Removed both .NET 6 and .NET 7

siri-varma:
Will resolve the conflicts on this

siri-varma:
Done. Fixed the conflicts

all.sln
```diff
@@ -155,6 +155,10 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "JobsSample", "examples\Jobs
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Dapr.Workflow.Test", "test\Dapr.Workflow.Test\Dapr.Workflow.Test.csproj", "{E90114C6-86FC-43B8-AE5C-D9273CF21FE4}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Dapr.Jobs.Analyzers", "src\Dapr.Jobs.Analyzer\Dapr.Jobs.Analyzers.csproj", "{28B87C37-4B52-400F-B84D-64F134931BDC}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Dapr.Jobs.Analyzer.Tests", "test\Dapr.Jobs.Analyzer.Test\Dapr.Jobs.Analyzer.Tests.csproj", "{CADEAE45-8981-4723-B641-9C28251C7D3B}"
+EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Messaging", "Messaging", "{8DB002D2-19E9-4342-A86B-025A367DF3D1}"
```
WhitWaldo, line 162: This project doesn't look right - everything typically has a "Dapr." prefix. Is this a merge gone wrong?

siri-varma: https://github.com/dapr/dotnet-sdk/blob/aa70c5bb5d4e5dfd1661efb41bfc7bd710c48eca/all.sln#L158

I see "Messaging" in release-1.16 🤔

WhitWaldo:
Thank you very much for your contribution @siri-varma !

Issue #1478, opened 2025-03-03
Need the ability to create Job specific handlers in the Jobs SDK


siri-varma:
## Describe the proposal

The https://github.com/dapr/dotnet-sdk/blob/cb065f2089da3f8725ea672a3172cb5bcb1f6a59/src/Dapr.Jobs/Extensions/EndpointRouteBuilderExtensions.cs#L33 allows configuring only one handler across all Jobs.

I would like the ability to provide different handlers by passing in the Job name.

## Current 
---------------------

```
    public static IEndpointRouteBuilder MapDaprScheduledJobHandler(this IEndpointRouteBuilder endpoints,
        Delegate action, TimeSpan? timeout = null) {
       
            MapPost("/job/{jobName}", ....)
    }
```

## Planned Change
----------------------

```
    public static IEndpointRouteBuilder MapDaprScheduledJobHandler(this IEndpointRouteBuilder endpoints,
        Delegate action, TimeSpan? timeout = null, string name= "{jobName}")

       MapPost("/job/{name}", ....)
     Add user provided job name to the route
```

siri-varma:
closing this pr

PR #1479, opened 2025-03-03
Add the ability to create job specific handler

siri-varma:
# Description

Add job name parameter to the EndpointBuilder Extension

## Issue reference

Please reference the issue this PR will close: #1478

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


WhitWaldo (commented):
The current implementation is more like an actor reminder invocation - switch on the job name provided and do something with it.

Why not create a separate method if the developer would like to instead register for a specific job name?

siri-varma:
Hey Whit ! I don't think this PR adds much value. I will close it.

Issue #1488, opened 2025-03-17
Migrate Dapr.Cryptography out to separate client


WhitWaldo:
Part of the work described in #1427 

With 1.16, we'll create a new client and run both the version in `Dapr.Client` and the separate package in parallel. Mark the affected methods in `Dapr.Client` with an `[Obsolete("Use the Dapr.Configuration NuGet package instead as this method will be removed from Dapr.Client in v1.17")]`.

PR #1490, opened 2025-03-17
Add Dapr.Cryptography + bugfix

WhitWaldo:
# Description

This PR implements three things:
1) It splits out the cryptography functionality out to a separate `Dapr.Cryptography` package to be released with 1.16. All existing methods on `Dapr.Client` have an updated `[Obsolete]` attribute with a message telling developers that this method will be removed with the release of 1.17.
2) An issue was reported [as a PR](https://github.com/dapr/dotnet-sdk/pull/1425) which identified a race condition wherein either the read or write buffer could run out of content and the other stream would stall as they were filling and draining in tandem. This PR offers a fix to run either operation independently while still waiting for both to complete before completing the work - this PR applies to both the implementation in `Dapr.Client` and the one in `Dapr.Cryptography`.
3) I identified a few small changes I could make for performance improvements and have applied these to both projects as well. The only difference is that `Dapr.Cryptography` wraps one of the methods in a try/catch block so as to better handle exceptions thrown by a `CancellationToken`, but this introduces another type that I didn't want to add to `Dapr.Client`.

## Issue reference

Please reference the issue this PR will close: #1488 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


src/Dapr.Cryptography/Encryption/DaprEncryptionGrpcClient.cs
```diff
@@ -0,0 +1,355 @@
...
+                ComponentName = vaultResourceName,
+                DataEncryptionCipher = encryptionOptions.EncryptionCipher.GetValueFromEnumMember(),
+                KeyName = keyName,
+                KeyWrapAlgorithm = encryptionOptions.KeyWrapAlgorithm.GetValueFromEnumMember(),
+                OmitDecryptionKeyName = shouldOmitDecryptionKeyName
+            };
+
+            if (!shouldOmitDecryptionKeyName)
+            {
+                ArgumentException.ThrowIfNullOrEmpty(encryptionOptions.DecryptionKeyName,
```
siri-varma, line 102: Quick question, I think shouldOmitDecryptionKeyName  will only be false in case string is not empty and has valid characters. In that case is this check required ?

src/Dapr.Cryptography/Encryption/EncryptionOptions.cs
```diff
@@ -0,0 +1,65 @@
...
+    /// </summary>
+    /// <remarks>
+    /// This defaults to 4KB and generally should not exceed 64KB.
+    /// </remarks>
+    public int StreamingBlockSizeInBytes
+    {
+        get => streamingBlockSizeInBytes;
+        set
+        {
+            if (value <= 0)
```
siri-varma, line 47: Should we also raise an illegal argument if the size is greater than 64 ?

siri-varma (approved):
Looks Good

src/Dapr.Cryptography/Encryption/EncryptionOptions.cs
WhitWaldo: The size could be configured to be larger than 64 as a configurable Dapr option and we wouldn't have any insight into it. If it's not configured and they pick too large a value, Dapr will throw an error accordingly.

WhitWaldo:
> Looks Good

I appreciate it, but it's incomplete. Building out an example and documentation as well.

src/Dapr.Cryptography/Encryption/DaprEncryptionProcessor.cs
```diff
@@ -0,0 +1,95 @@
...
+            var buffer = ArrayPool<byte>.Shared.Rent(blockSize);
+            try
+            {
+                int bytesRead;
+                ulong sequenceNumber = 0;
+
+                while ((bytesRead = await bufferedStream.ReadAsync(buffer.AsMemory(0, blockSize), cancellationToken)) !=
+                       0)
+                {
+                    await duplexStream.RequestStream.WriteAsync(
```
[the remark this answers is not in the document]
WhitWaldo: This whole PR got a huge rewrite last night as it was still experiencing the reported issue of bombing out on large files, so review with a grain of salt.

While it could from a C# perspective be done in parallel with tasks, I don't think it would work with the Dapr runtime as it requires we submit the sequence number alongside the request payload and I believe it requires this to monotonically increase as it goes. It's not my recollection it will process chunks out of order. Also, I don't know that the response includes the sequence number, so re-assembly on the retrieval side could be problematic without some guarantee that the runtime is doing some reordering operation (and I don't know of any such guarantee).

WhitWaldo:
Closing this in favor of a PR that moves the Crypto client out to a separate package and has been tested to work with files up to 1 GB.

PR #1527, opened 2025-05-01
Add Dapr.Cryptography package + fix for large files

WhitWaldo:
# Description

This PR implements three things:
1. It splits out the cryptography building block to a separate `Dapr.Cryptography` package to be released with 1.16. All existing methods on Dapr.Client have an updated [Obsolete] attribute with a message telling developers that this method will be removed with the release of 1.17.
2. An issue was reported as a PR which identified a race condition wherein either the read or write buffer could run out of content and the other stream would stall as they were filling and draining in tandem. This PR fundamentally rewrites the implementation so this is no longer the case. A separate PR will be created to fix the implementation in the DaprClient so it can be patched in a 1.15 release.
3. I've updated the cryptography example to use the separate `Dapr.Cryptography` package and modernized to reflect a DI-first approach. It features three samples that demonstrate encryption and decryption of strings, streams and byte arrays of small (a dozen bytes), medium (~1 KB) and 1 GB files.

## Issue reference

Please reference the issue this PR will close: #1488 [https://github.com/dapr/dapr/issues/8244](https://github.com/dapr/dapr/issues/8244)

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


daprdocs/content/en/dotnet-sdk-docs/dotnet-cryptography/_index.md
```diff
@@ -0,0 +1,12 @@
+---
+type: docs
+title: "Dapr Cryptography .NET SDK"
+linkTitle: "Cryptography"
+weight: 51000
+description: Get up and running with the Dapr AI .NET SDK
```
watfordsuzy, line 6:
```suggestion
description: Get up and running with the Dapr Cryptography .NET SDK
```

daprdocs/content/en/dotnet-sdk-docs/dotnet-cryptography/dotnet-cryptography-usage.md
```diff
@@ -0,0 +1,131 @@
+Whit---
```
watfordsuzy, line 1:
```suggestion
---
```

examples/Cryptography/Examples/EncryptDecryptLargeFileExample.cs
```diff
@@ -0,0 +1,107 @@
...
+
+    private static async Task WriteLargeFileAsync(string fileName, long sizeLimit)
+    {
+        var buffer = new byte[5 * 1024 * 1024]; // 5 MB buffer
+        var random = new Random();
+
+        await using var fs = new FileStream(fileName, FileMode.Create, FileAccess.Write, FileShare.Read);
+        for (var written = 0; written < sizeLimit; written += buffer.Length)
+        {
+            random.NextBytes(buffer);
```
watfordsuzy, line 96:
```suggestion
        await using var fs = new FileStream(fileName, FileMode.Create, FileAccess.Write, FileShare.Read);
        for (var written = 0; written < sizeLimit; written += buffer.Length)
        {
            Random.Shared.NextBytes(buffer);
```

src/Dapr.Cryptography/Encryption/Models/DecryptionOptions.cs
```diff
@@ -0,0 +1,34 @@
...
+//  ------------------------------------------------------------------------
+
+namespace Dapr.Cryptography.Encryption.Models;
+
+/// <summary>
+/// A collection fo options used to configure how decryption cryptographic operations are performed.
+/// </summary>
+public sealed class DecryptionOptions
+{
+    private int streamingBlockSizeInBytes = 4 * 1024; // 4KB
```
watfordsuzy, line 21: 4KB seems a little small for a buffer, but not 100% sure what the .NET gRPC defaults are (16KB? 64KB?).

src/Dapr.Cryptography/Encryption/EncryptionStreamProcessor.cs
```diff
@@ -0,0 +1,120 @@
...
+
+    /// <summary>
+    /// Sends the provided bytes in chunks to the sidecar for the encryption operation.
+    /// </summary>
+    /// <param name="inputStream">The stream containing the bytes to encrypt.</param>
+    /// <param name="call">The call to make to the sidecar to process the encryption operation.</param>
+    /// <param name="options">The encryption options.</param>
+    /// <param name="streamingBlockSizeInBytes">The size, in bytes, of the streaming blocks.</param>
+    /// <param name="cancellationToken">Token used to cancel the ongoing request.</param>
+    public async Task ProcessStreamAsync(
```
watfordsuzy, line 36: I think I'd expect this to return `Task.WhenAll(writerTask, readerTask)`

src/Dapr.Cryptography/Encryption/DaprEncryptionGrpcClient.cs
```diff
@@ -0,0 +1,188 @@
...
+            ComponentName = vaultResourceName, 
+            KeyName = keyName
+        };
+
+        var grpcCallOptions = DaprClientUtilities.ConfigureGrpcCallOptions(typeof(DaprEncryptionClient).Assembly,
+            this.DaprApiToken, cancellationToken);
+        var duplexStream = Client.DecryptAlpha1(grpcCallOptions);
+
+        using var streamProcessor = new DecryptionStreamProcessor();
+        await streamProcessor.ProcessStreamAsync(ciphertextStream, duplexStream, options.StreamingBlockSizeInBytes,
```
watfordsuzy, line 179: This await doesn't really do anything given how the processor is setup.

I think a design like you see in the Azure.Data.Tables library would be amenable here: https://learn.microsoft.com/en-us/dotnet/api/azure.data.tables.tableclient.queryasync?view=azure-dotnet

Basically, instead of returning a `Task` the process stream returns a wrapper which includes the IAsyncEnumerable.

src/Dapr.Cryptography/Encryption/DaprEncryptionGrpcClient.cs
```diff
@@ -0,0 +1,188 @@
...
+                nameof(encryptionOptions.DecryptionKeyName));
+            encryptRequestOptions.DecryptionKeyName = encryptRequestOptions.DecryptionKeyName;
+        }
+
+        var grpcCallOptions = DaprClientUtilities.ConfigureGrpcCallOptions(typeof(DaprEncryptionClient).Assembly,
+            this.DaprApiToken, cancellationToken);
+        var duplexStream = Client.EncryptAlpha1(grpcCallOptions);
+
+        using var streamProcessor = new EncryptionStreamProcessor();
+        await streamProcessor.ProcessStreamAsync(plaintextStream, duplexStream, encryptRequestOptions,
```
watfordsuzy, line 106: One interesting thing I noticed in the dapr Go code is that it does encryption/decryption on the full buffer in memory (`byte[]`) rather than streaming.

examples/Cryptography/Examples/EncryptDecryptLargeFileExample.cs
```diff
@@ -0,0 +1,107 @@
...
+            await fs.WriteAsync(buffer);
+        }
+    }
+
+    private static async Task<string> GetFileHashAsync(string fileName)
+    {
+        using var md5 = MD5.Create();
+        await using var stream = File.OpenRead(fileName);
+        return Convert.ToBase64String(await md5.ComputeHashAsync(stream));
```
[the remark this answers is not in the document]
WhitWaldo: It's a demo demonstrating that the input matches the output. I don't think it really matters which hash we use.

src/Dapr.Cryptography/Encryption/Models/DecryptionOptions.cs
WhitWaldo: Anyone is welcome to change this to match their own situations, but this mirrors the default file stream buffer size in C# (e.g. via `FileStream`).

src/Dapr.Cryptography/Encryption/DaprEncryptionGrpcClient.cs
WhitWaldo: I can't speak to what happens in the runtime, but the specification is that this is implemented using a bidirectional streaming connection.

src/Dapr.Cryptography/Encryption/DaprEncryptionGrpcClient.cs
WhitWaldo: It's true that await doesn't really do anything, but between it and just discarding the value, C# requires that I put something there. I don't know that I want to return any additional public types just for the sake of it when I could just return the `IAsyncEnumerable` and leave it to consumers to decide what to do with it.

src/Dapr.Cryptography/Encryption/DecryptionStreamProcessor.cs
```diff
@@ -0,0 +1,120 @@
...
+    /// <param name="cancellationToken">Token used to cancel the ongoing request.</param>
+    public async Task ProcessStreamAsync(
+        Stream inputStream,
+        AsyncDuplexStreamingCall<Autogenerated.DecryptRequest, Autogenerated.DecryptResponse> call,
+        int streamingBlockSizeInBytes,
+        Autogenerated.DecryptRequestOptions options,
+        CancellationToken cancellationToken)
+    {
+        //Read from the input stream and write to the gRPC call
+        _ = Task.Run(async () =>
```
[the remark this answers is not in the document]
WhitWaldo: Excellent point - we're rather limited in what we can do here for fire-and-forget, so I've opted to set up an event that the gRPC client implementation can subscribe to that'll surface it to the caller.

src/Dapr.Cryptography/Encryption/DecryptionStreamProcessor.cs
```diff
@@ -0,0 +1,134 @@
...
+                        await call.RequestStream.WriteAsync(request, cancellationToken);
+
+                        //Increment the sequence number
+                        sequenceNumber++;
+                    }
+                }
+
+                await call.RequestStream.CompleteAsync();
+            }
+            catch(Exception ex)
```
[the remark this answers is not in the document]
WhitWaldo: Sure, that's fair.

src/Dapr.Cryptography/Encryption/DecryptionStreamProcessor.cs
```diff
@@ -0,0 +1,134 @@
...
+                        }
+
+                        await call.RequestStream.WriteAsync(request, cancellationToken);
+
+                        //Increment the sequence number
+                        sequenceNumber++;
+                    }
+                }
+
+                await call.RequestStream.CompleteAsync();
```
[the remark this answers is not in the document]
WhitWaldo: Sure, I've got no problem with that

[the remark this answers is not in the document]
watfordsuzy: Not 100% sure I can use Discord at work so I've sent an email.

Issue #1273, opened 2024-04-16
intermittent exceptions from WorkflowLoggingService


TomasEkeli:
## Expected Behavior

It should be possible to run outside-in tests on an aspnetcore application with dapr without intermittent failures.

## Actual Behavior

sometimes when i run my test suite (not always) one of many tests will fail with a stack-trace telling me that something was added to a hash-set in a way that is not allowed. 

I am using dapr.aspnetcore and dapr.workflow version 1.13.0, with CLI version 1.13.0 and runtime version 1.13.2

I am using dotnet8 in a devcontainer.

i have set up my aspnetcore app with outside-in tests that use a WebApplicationFactory.

note: the failing test is not the same one between runs, and the failing tests are not related to dapr.

test-run-log with stack-trace (with references to my code expunged):

```
Test run for /***/bin/Debug/net8.0/***.tests.dll (.NETCoreApp,Version=v8.0)
Starting test execution, please wait...
A total of 1 test files matched the specified pattern.
[xUnit.net 00:00:01.98]     *** [FAIL]
  Failed *** [1 ms]
  Error Message:
   System.InvalidOperationException : Operations that change non-concurrent collections must have exclusive access. A concurrent update was performed on this collection and corrupted its state. The collection's state is no longer correct.
  Stack Trace:
     at System.Collections.Generic.HashSet`1.AddIfNotPresent(T value, Int32& location)
   at System.Collections.Generic.HashSet`1.Add(T item)
   at Dapr.Workflow.WorkflowLoggingService.LogWorkflowName(String workflowName)
   at Dapr.Workflow.WorkflowRuntimeOptions.<>c__DisplayClass3_0`1.<RegisterWorkflow>b__0(DurableTaskRegistry registry)
   at Dapr.Workflow.WorkflowRuntimeOptions.AddWorkflowsAndActivitiesToRegistry(DurableTaskRegistry registry)
   at Dapr.Workflow.WorkflowServiceCollectionExtensions.<>c__DisplayClass0_1.<AddDaprWorkflow>b__1(DurableTaskRegistry registry)
   at Microsoft.Extensions.Options.OptionsFactory`1.Create(String name)
   at System.Lazy`1.ViaFactory(LazyThreadSafetyMode mode)
   at System.Lazy`1.ExecutionAndPublication(LazyHelper executionAndPublication, Boolean useDefaultConstructor)
   at System.Lazy`1.CreateValue()
[...]

PR #1556, opened 2025-06-19
Replace HashSet with ConcurrentDictionary in WorkflowLoggingService

TomasEkeli:
This should help with issue #1273 - intermittent exceptions from WorkflowLoggingService.

There is no concurrent HashSet in .NET, so we use ConcurrentDictionary with a byte as the value to simulate a set. For this use-case we do not care if the Key already exists, so we can use TryAdd to add items without considering the return value.

# Description

Replaces `HashSet<string>` as the collection that stores Activity and Workflow names for logging with `ConcurrentDictionary<string, byte>`

## Issue reference

Please reference the issue this PR will close: #1273

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/src/Dapr.Workflow/WorkflowLoggingService.cs
+++ b/src/Dapr.Workflow/WorkflowLoggingService.cs
@@ -13,12 +13,11 @@
 
 namespace Dapr.Workflow
 {
-    using System.Collections.Generic;
     using System.Threading;
     using System.Threading.Tasks;
     using Microsoft.Extensions.Hosting;
     using Microsoft.Extensions.Logging;
-    using Microsoft.Extensions.Configuration;
+    using System.Collections.Concurrent;
 
     /// <summary>
     /// Defines runtime options for workflows.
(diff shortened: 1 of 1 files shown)


WhitWaldo:
First of all, thank you for taking the time to spot this and submit a PR - I really appreciate it!

Could you please sign your commit? You can read more about DCO requirement [here](https://docs.dapr.io/contributing/contributing-overview/#developer-certificate-of-origin-signing-your-work), but we'll need it in place before the PR can be merged.

WhitWaldo (approved):
Looks fine to me - good to go once the DCO is in place.

TomasEkeli:
I have force-pushed with the requried string, but for some reason the DCO does not register it?

WhitWaldo:
Thank you very much for your contribution here!

marcduiker:
@holopin-bot @TomasEkeli Thank you Tomas! Here's a digital badge as a small token of appreciation.

PR #1591, opened 2025-08-08
Add A2A implementation for TaskStore

yaron2:
Adds a Dapr state API implementation for Dotnet A2A, allowing agents to persist tasks using Dapr.

### DIFF
--- a/src/Dapr.AI/A2A/DaprTaskStore.cs
+++ b/src/Dapr.AI/A2A/DaprTaskStore.cs
@@ -1,0 +1,162 @@
+using A2A;
+using Dapr.Client;
+
+/// <summary>
+/// Represents a Dapr-backed state store implementation for the A2A Dotnet ITaskStore interface, allowing agents to keep persistent state.
+/// </summary>
+public class DaprTaskStore : ITaskStore
+{
+    private readonly DaprClient _daprClient;
+    private readonly string _stateStoreName;
+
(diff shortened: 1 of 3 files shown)


src/Dapr.AI/A2A/DaprTaskStore.cs
```diff
@@ -0,0 +1,162 @@
+using A2A;
+using Dapr.Client;
+
+/// <summary>
+/// Represents a Dapr-backed state store implementation for the A2A Dotnet ITaskStore interface, allowing agents to keep persistent state.
+/// </summary>
+public class DaprTaskStore : ITaskStore
```
WhitWaldo, line 7: Given the unimplemented methods, I think this might be better to implement as an abstract base class instead in which the developer can implement what they'd like (or throw exceptions, but at least then they know that they aren't implemented). We can always revisit as the base framework continues to evolve (it looks like there's a lot of opportunity for improvement). 

Please mark this class as 'abstract' as well as each of the push notification methods and remove the method bodies.

WhitWaldo (changes requested):
I would like to see this moved out to a separate `Dapr.AI.A2A` project so we don't pollute the existing package with this and all the other extension projects we'll undoubtedly cover over time. I'll follow up with you offline to talk you through this!

Thank you for putting this together and bringing the A2A project to my attention!

PR #1593, opened 2025-08-11
Add A2A Dapr Task Store

yaron2:
Adds a Dapr-enabled Task Store implementation to A2A Dotnet.


WhitWaldo (approved):
Thank you for your contribution and for putting this together - I look forward to iterating on it as A2A increasingly grows into a more mature implementation!

PR #1638, opened 2025-09-29
- Added a new method `PublishBulkByteEventAsync` in DaprClientGrpc.cs

kapilthareja:
to enable publishing multiple messages in a single call via gRPC.
- Method accepts IEnumerable<byte[]> and publishes them to a specified pub/sub topic, improving performance over repeated PublishEventAsync calls.
- This change aligns the .NET SDK with the bulk publish pattern in the Go SDK.
- Ensured backward compatibility; existing PublishEventAsync remains unchanged.
- Updated internal gRPC serialization and invocation logic to handle batch messages.
- Tested locally by publishing bulk messages to Redis pub/sub through a producer project.

This commit is part of the feature branch `feature/PublishBulkByteEventAsync`.

# Description

1️⃣ New Method: PublishBulkByteEventAsync

Where added: DaprClient.cs (abstract method), DaprClientGrpc.cs (gRPC implementation), DaprClientHttp.cs (HTTP implementation), and IDaprClient.cs (interface).

Purpose: Allow publishing multiple raw byte[] messages in a single request to a Dapr pub/sub component.

Reason: The official .NET SDK did not have a built-in bulk byte publish method like the Go SDK’s PublishBulkEvent.

Code Example
public abstract Task PublishBulkByteEventAsync(
    string pubsubName,
    string topicName,
    IEnumerable<byte[]> events,
    Dictionary<string, string>? metadata = null,

2️⃣ gRPC Implementation (DaprClientGrpc.cs)

Converts each byte[] event into a BulkPublishRequestEntry using ByteString.CopyFrom.

Adds optional metadata.

Calls the gRPC PublishBulkEventAsync API of the Dapr sidecar.

Why

Mirrors the Go SDK’s PublishBulkEvent implementation.

Ensures high-performance bulk publish without serializing to JSON unnecessarily.

3️⃣ HTTP Implementation (DaprClientHttp.cs)

Uses the Dapr HTTP bulk publish API:

POST /v1.0-alpha1/publish/bulk/{pubsub}/{topic}

Each event is Base64-encoded for transport (Convert.ToBase64String(e)) and labeled with application/octet-stream.

Sends all events in a single HTTP request.

Why

Supports environments where gRPC is not available.

Maintains parity with the Go SDK bulk publish behavior.

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


src/Dapr.Client/DaprClient.cs
```diff
@@ -917,10 +917,10 @@ public abstract Task<ReadOnlyMemory<byte>> GetByteStateAsync(
     /// <param name="cancellationToken">A <see cref="CancellationToken" /> that can be used to cancel the operation.</param>
     /// <returns>A <see cref="Task{T}" /> that will return the value when the operation has completed.  This wraps the read value and an ETag.</returns>
     public abstract Task<(ReadOnlyMemory<byte>, string etag)> GetByteStateAndETagAsync(
-        string storeName, 
-        string key, 
-        ConsistencyMode? consistencyMode = null, 
-        IReadOnlyDictionary<string, string>? metadata = null, 
+        string storeName,
```
WhitWaldo, line 920: Generally, please don't submit PRs with formatting changes like this as everyone's IDE is going to add/remove stuff like this based on local preferences and it slows down actually reviewing the changes you've made.

src/Dapr.Client/DaprClient.cs
```diff
@@ -1423,4 +1423,22 @@ protected static ProductInfoHeaderValue UserAgent()
...
+    /// <param name="events">The list of raw byte events to be published.</param>
+    /// <param name="dataContentType">The content type of the given bytes, defaults to application/json.</param>
+    /// <param name="metadata">A collection of metadata key-value pairs that will be provided to the pubsub. The valid metadata keys and values are determined by the type of binding used.</param>
+    /// <param name="cancellationToken">A <see cref="CancellationToken" /> that can be used to cancel the operation.</param>
+    /// <returns>A <see cref="Task" /> that will complete when the operation has completed.</returns>
+    public abstract Task<BulkPublishResponse<byte[]>> PublishBulkByteEventAsync(
+        string pubsubName,
+        string topicName,
+        IReadOnlyList<ReadOnlyMemory<byte>> events,
+        string dataContentType = "application/json",
```
WhitWaldo: For consistency, could you please have this use the same referenced constant as the non-bulk equivalent, `Constants.ContentTypeApplicationJson`?

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -263,6 +263,71 @@ events[counter] is CloudEvent
         }
     }
 
+#nullable enable
+    public override async Task<BulkPublishResponse<byte[]>> PublishBulkByteEventAsync(
+        string pubsubName,
+        string topicName,
+        IReadOnlyList<ReadOnlyMemory<byte>> events,
+        string dataContentType = "application/json",
```
WhitWaldo: Again, please update the default value to use the pre-existing constant `Constants.ContentTypeApplicationJson`.

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -263,6 +263,71 @@ events[counter] is CloudEvent
...
+        Dictionary<string, string>? metadata = null,
+        CancellationToken cancellationToken = default)
+    {
+        var request = new Autogenerated.BulkPublishRequest
```
WhitWaldo: Is there a reason we cannot refactor and use the existing generic `MakeBulkPublishRequest` here instead of duplicating the method for this additional type?

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -574,7 +639,8 @@ public override async Task InvokeMethodGrpcAsync(string appId, string methodName
 
         var envelope = new Autogenerated.InvokeServiceRequest()
         {
-            Id = appId, Message = new Autogenerated.InvokeRequest() { Method = methodName, },
+            Id = appId,
```
WhitWaldo, line 583: To this and the remaining formatting changes - same as above, please don't submit these generally.

test/Dapr.Client.Test/PublishEventApiTest.cs
```diff
@@ -294,6 +295,37 @@ public async Task PublishEventAsync_CanPublishWithRawData()
         envelope.Metadata.Count.ShouldBe(0);
     }
 
+    [Fact]
+    public async Task PublishBulkByteEventAsync_PublishesEventsAndReturnsResponse()
+    {
+        // Arrange
+        var client = new DaprClientBuilder().Build();
```
WhitWaldo: Unit tests aren't run with a running instance of Dapr, so this isn't going to work correctly. I would advise copying the example above using the `TestClient` it produces so you can monitor the "captured" gRPC request and validate against that payload.

WhitWaldo (changes requested):
Good catch - yes, that's certainly an oversight. 

In the future, please avoid PRs with a bunch of minor formatting changes. As I explained in one of the comments, those tend to be styled based on your own personal IDE preferences and settings. It just distracts from the actual code you're submitting and will just get flagged in all other contributors' IDEs as opportunities to fix when I'd really rather not get into formatting specifications.

PR #1713, opened 2026-02-19
fix: handle graceful cancellation in CloudEventsMiddleware to prevent 499 errors

### LINKED ISSUE #1237: "499 client closed connection" on PubSub consumer after upgrading to .NET 8
## Expected Behavior

Dapr not to close connections when event processing is successful.

## Actual Behavior

Dapr appears to be passing a cancellation to my API controllers which is triggering a `499 - Client Closed Connection` exception to throw even after processing within our application completes successfully. 

What I am observing from my OTEL traces is an event will be picked up by the Dapr sidecar off a topic, passed into my ASP.NET controller where it is successfully processed. Our code uses Dapr to publish an event on to the next topic. It all then wraps up and returns a `200 - OK` to the sidecar. Dapr should then ack the message off the queue. But there are many times where it doesn't. It seems to throw some kind of processing error. The status code I see on the sidecar in the trace is a response code of `2` which then appears to be passed into the controller as a client closing the connection which results in a flurry of 499 response codes.

A bit more context is needed here. Our API controllers are all built to take in a `CancellationToken cancellationToken = default` as the final parameter of every controller. Obviously that means that the calling context which Dapr provides is wrapped up and through magical ASP.NET middleware makes it's way into our code. At every layer in our code, we check the status of the cancellation token in an effort to reduce the amount of work the system is doing at all times. 
From what I can tell, the failure in the Dapr sidecar is propagating a cancelled context into our middleware and that is resulting in the trace being brought up to Application Insights. As best we can see, the event does indeed get published on to the next service via PubSub so in terms of functionality, nothing is necessarily wrong. But the behaviour we are observing is extremely erratic and it floods our telemetry with useless errors and hides real errors due to the sampling rates we have picked for cost reasons. 

In a given 24 hour period, we're seeing on average 40k of these `499` errors. We also see around 5000 - 10,000 `500` errors for cancellations being triggered. My gut feel says both are similarly related. 

When I enabled debug logs in my sidecar, I am seeing a _lot_ of messages saying the following:

> skipping status check due to error parsing result from pub/sub event <nil>: EOF

I think this is where the error starts.

More context:

* .NET 8 / C# 12
* AKS
* Using Kafka for PubSub
* Running Dapr 1.12.5 with 1.12.0 NuGet packages

I am only able to reproduce this under heavy load - during peak periods where we are doing around 170k events per minute. I have never been able to reproduce it locally or in my dev/test environment.

## Release Note

RELEASE NOTE:

ali-Hamza817:
# Description
This PR addresses the "499 Client Closed Connection" error in Pub/Sub consumers by properly propagating the CancellationToken from the HttpContext to the JsonSerializer. It also implements a graceful catch for OperationCanceledException that swallows the error ONLY if the response has already started, preventing erroneous telemetry failures during high-load race conditions.

## Issue reference
Fixes #1237

## Checklist
* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/src/Dapr.AspNetCore/CloudEventsMiddleware.cs
+++ b/src/Dapr.AspNetCore/CloudEventsMiddleware.cs
@@ -48,28 +48,22 @@ internal class CloudEventsMiddleware
         this.options = options;
     }
 
-    public Task InvokeAsync(HttpContext httpContext)
+    public async Task InvokeAsync(HttpContext httpContext)
     {
-        // This middleware unwraps any requests with a cloud events (JSON) content type
-        // and replaces the request body + request content type so that it can be read by a
-        // non-cloud-events-aware piece of code.
-        //
-        // This corresponds to cloud events in the *structured* format:
-        // https://github.com/cloudevents/spec/blob/master/http-transport-binding.md#13-content-modes
-        //
-        // For *binary* format, we don't have to do anything
-        //
-        // We don't support batching.
-        //
-        // The philosophy here is that we don't report an error for things we don't support, because
-        // that would block someone from implementing their own support for it. We only report an error
-        // when something we do support isn't correct.
         if (!MatchesContentType(httpContext, out var charSet))
         {
-            return this.next(httpContext);
+            await this.next(httpContext);
+            return;
         }
 
-        return this.ProcessBodyAsync(httpContext, charSet);
+        try
+        {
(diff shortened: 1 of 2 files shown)


marcduiker:
@holopin-bot @ali-Hamza817 sdk-badge Thank you! Here's a digital badge as a small token of appreciation.

PR #1738, opened 2026-03-06
Implement ListInstanceIDs, GetInstanceHistory, and RerunWorkflowFromEvent workflow RPCs

JoshVanL:
Add the three missing workflow RPCs that are defined in the proto but were not yet exposed through the SDK client. This includes the full stack: abstract base class, gRPC implementation, public interface, public client, model types, logging, and tests.

### DIFF
--- a/src/Dapr.Workflow/Client/WorkflowClient.cs
+++ b/src/Dapr.Workflow/Client/WorkflowClient.cs
@@ -190,7 +190,7 @@ internal abstract class WorkflowClient : IAsyncDisposable
     /// <param name="pageSize">The maximum number of instance IDs to return, or <c>null</c> for no limit.</param>
     /// <param name="cancellationToken">A token that can be used to cancel the operation.</param>
     /// <returns>A page of instance IDs and an optional continuation token for the next page.</returns>
-    public abstract Task<WorkflowInstancePage> ListInstanceIDsAsync(
+    public abstract Task<WorkflowInstancePage> ListInstanceIdsAsync(
         string? continuationToken = null,
         int? pageSize = null,
         CancellationToken cancellationToken = default);
--- a/src/Dapr.Workflow/Client/WorkflowGrpcClient.cs
+++ b/src/Dapr.Workflow/Client/WorkflowGrpcClient.cs
@@ -227,7 +227,7 @@ internal sealed class WorkflowGrpcClient(
     }
 
     /// <inheritdoc />
-    public override async Task<WorkflowInstancePage> ListInstanceIDsAsync(
+    public override async Task<WorkflowInstancePage> ListInstanceIdsAsync(
         string? continuationToken = null,
         int? pageSize = null,
         CancellationToken cancellationToken = default)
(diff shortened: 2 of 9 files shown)


src/Dapr.Workflow/DaprWorkflowClient.cs
```diff
@@ -315,6 +316,67 @@ public async Task<bool> PurgeInstanceAsync(
...
+    /// The continuation token from a previous call, or <c>null</c> to retrieve the first page.
+    /// </param>
+    /// <param name="pageSize">
+    /// The maximum number of instance IDs to return per page, or <c>null</c> for no limit.
+    /// </param>
+    /// <param name="cancellation">Token to cancel the list operation.</param>
+    /// <returns>
+    /// A page containing the instance IDs and an optional continuation token for the next page.
+    /// </returns>
+    public async Task<WorkflowInstancePage> ListInstanceIDsAsync(
```
WhitWaldo: Just as a convention, .NET uses "Ids" instead of "IDs" in method names - could you change this?

src/Dapr.Workflow/IDaprWorkflowClient.cs
```diff
@@ -234,4 +235,49 @@ Task ResumeWorkflowAsync(
...
+    /// The continuation token from a previous call, or <c>null</c> to retrieve the first page.
+    /// </param>
+    /// <param name="pageSize">
+    /// The maximum number of instance IDs to return per page, or <c>null</c> for no limit.
+    /// </param>
+    /// <param name="cancellation">Token to cancel the list operation.</param>
+    /// <returns>
+    /// A page containing the instance IDs and an optional continuation token for the next page.
+    /// </returns>
+    Task<WorkflowInstancePage> ListInstanceIDsAsync(
```
WhitWaldo: Same thing - please change from "IDs" to "Ids"

src/Dapr.Workflow/Client/WorkflowClient.cs
```diff
@@ -182,6 +183,42 @@ public abstract Task ResumeWorkflowAsync(string instanceId, string? reason = nul
...
     public abstract Task<bool> PurgeInstanceAsync(string instanceId, CancellationToken cancellationToken = default);
 
+    /// <summary>
+    /// Lists workflow instance IDs with optional pagination.
+    /// </summary>
+    /// <param name="continuationToken">The continuation token from a previous call, or <c>null</c> for the first page.</param>
+    /// <param name="pageSize">The maximum number of instance IDs to return, or <c>null</c> for no limit.</param>
+    /// <param name="cancellationToken">A token that can be used to cancel the operation.</param>
+    /// <returns>A page of instance IDs and an optional continuation token for the next page.</returns>
+    public abstract Task<WorkflowInstancePage> ListInstanceIDsAsync(
```
WhitWaldo: "IDs" -> "Ids"

src/Dapr.Workflow/Client/WorkflowGrpcClient.cs
```diff
@@ -224,6 +226,94 @@ public override async Task<bool> PurgeInstanceAsync(string instanceId, Cancellat
...
+
+        if (options?.NewInstanceId is not null)
+        {
+            request.NewInstanceID = options.NewInstanceId;
+        }
+
+        if (options is { OverwriteInput: true })
+        {
+            request.Input = SerializeToJson(options.Input);
+        }
```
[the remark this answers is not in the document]
WhitWaldo: @JoshVanL Could you confirm that if `OverwriteInput` is false, that if `request.Input` isn't set by the SDK, the runtime will properly pick it up from the workflow it's re-running from itself?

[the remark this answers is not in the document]
JoshVanL: Yes - when OverwriteInput is false, the SDK intentionally does not set request.Input.

PR #1739, opened 2026-03-11
feat: add application-level keepalive to prevent ALB idle connection timeouts

JoshVanL:
AWS ALBs do not forward HTTP/2 PING frames, causing idle gRPC connections to be closed. This adds a background loop that periodically calls the existing Hello RPC as application-level traffic to keep the connection alive through L7 load balancers.


WhitWaldo (approved):
Looks good - thank you for putting this together!

PR #1744, opened 2026-03-13
Fix SubscribeAsync silently failing when Dapr sidecar is unavailable

### LINKED ISSUE #1663: [Messaging] SubscribeAsync silently fails with unobserved task exceptions when Dapr sidecar is unavailable
## Expected Behavior

  When the Dapr sidecar is unavailable (not started, crashed, or disconnected), `SubscribeAsync()` should throw a catchable exception (like `DaprException` used in other SDK methods) so applications can handle failures gracefully.

  ## Actual Behavior

  Exceptions are hidden and become **unobserved task exceptions** that are nearly impossible to track:

  1. **If sidecar is not started**: Calling `SubscribeAsync()` appears to succeed, but silently fails. No exception is thrown.
  2. **If sidecar crashes after subscription**: The failure triggers `TaskScheduler.UnobservedTaskException` only when the task is garbage collected (may never happen or happen much later).
  3. No error callback, event, or observable way to detect the failure.

  ### Root Cause

  In `PublishSubscribeReceiver.cs:113-134`, `SubscribeAsync()` starts background tasks with fire-and-forget:

  ```csharp
  internal async Task SubscribeAsync(CancellationToken cancellationToken = default)
  {
      // ...
      var stream = await GetStreamAsync(cancellationToken);  // Line 121

      // Fire-and-forget - exceptions become unobserved
      _ = FetchDataFromSidecarAsync(stream, ...)
          .ContinueWith(HandleTaskCompletion, ..., TaskContinuationOptions.OnlyOnFaulted, ...);

      _ = ProcessAcknowledgementChannelMessagesAsync(...)
          .ContinueWith(HandleTaskCompletion, ..., TaskContinuationOptions.OnlyOnFaulted, ...);

      // Method returns immediately - appears successful even if sidecar is down
  }

  When GetStreamAsync() or background tasks fail, exceptions go to HandleTaskCompletion which re-throws into the void:

  internal static void HandleTaskCompletion(Task task, object? state)
  {
      if (task.Exception != null)
      {
          throw task.Exception;  // ⚠️ Becomes UnobservedTaskException
      }
  }

  Steps to Reproduce the Problem

  // Don't start daprd at all
  var subscription = await pubSubClient.SubscribeAsync("pubsub", "topic", ...);
  // Returns successfully, no exception thrown
  // Silently fails - very hard to debug

  // Or: start daprd, subscribe, then kill daprd
  // Exception only appears in TaskScheduler.UnobservedTaskException (if subscribed)
  // Otherwise completely silent

  Proposed Solution

  1. Make SubscribeAsync() actually await the connection establishment and throw on failure
  2. Add observable error callback: event EventHandler<Exception>? OnError
  3. Store and expose background tasks for monitoring
  4. Wrap RpcException in DaprException for consistency with other SDK methods

  Release Note

  RELEASE NOTE: FIX SubscribeAsync now properly throws exceptions when Dapr sidecar is unavailable instead of silently failing.

GeorgeTsiokos:
## Summary

Fixes #1663

`PublishSubscribeReceiver.SubscribeAsync` used to silently fail when the Dapr sidecar was unavailable: the initial gRPC call could throw, and three background continuations could fault, with all of those errors disappearing into unobserved task exceptions. This PR closes both halves of that hole and makes the receiver safely re-subscribable.

- **Initial-subscribe failures throw**: `SubscribeAsync` catches `RpcException` from the gRPC stream setup and throws a `DaprException` with topic / pubsub context, then resets `hasInitialized` so the caller can retry once the sidecar is back.
- **Background-task faults are no longer silent**: `HandleTaskCompletion` no longer throws (which would have faulted the unobserved inner `Task` returned by `_ = ContinueWith(HandleTaskCompletion, ...)` — the exact bug class this PR set out to fix). Instead it caches the first fault per subscribe cycle in `pendingBackgroundFault`; the next call to `SubscribeAsync` drains and rethrows it, so callers that don't configure an `ErrorHandler` still observe the error.
- **Single sidecar outage produces exactly one report**: a new `hasFaulted` flag (CAS) dedupes across the three sibling continuations (`FetchDataFromSidecarAsync`, `ProcessAcknowledgementChannelMessagesAsync`, `ProcessTopicChannelMessagesAsync`), which previously could each fire `HandleTaskCompletion` for the same logical failure.
- **Stream reset is race-free and leak-free**: new `ResetStreamStateAsync` runs under the same `SemaphoreSlim` as `GetStreamAsync`, nulls `clientStream` *before* resetting `hasInitialized`, and disposes the prior `AsyncDuplexStreamingCall<,>` to release the gRPC call. This closes the race where a concurrent `SubscribeAsync` could observe a reset `hasInitialized` while still seeing a stale stream reference.
- **`SubscriptionErrorHandler`-thrown exceptions are surfaced**: if a user-supplied error handler itself throws, the original `DaprException` and the handler exception are combined into an `AggregateException` and cached for the next `SubscribeAsync` instead of being silently swallowed by an empty `catch`.
- **Cancellation is detected across all inner exceptions**: the `OperationCanceledException` short-circuit now inspects every entry in `task.Exception.InnerExceptions` rather than only the first, so a fault wrapping both an OCE and a real `RpcException` still surfaces the real error rather than being suppressed as a clean cancel.
- **`ErrorHandler` delegate is invoked safely**: `ErrorHandler?.Invoke()` is guarded by try/catch, with the catch now routing through the cache-and-rethrow path described above instead of being a silent drop.
- **Continuations use `CancellationToken.None`**: ensures error-handling continuations always execute even after the caller's token is cancelled, so faults aren't lost to `OnlyOnFaulted` not running.
- **Doc comments updated**: XML docs on `DaprSubscriptionOptions.ErrorHandler` and `SubscriptionErrorHandler` now describe the cache-and-rethrow contract accurately (the prior comment claimed errors "surface as unobserved task exceptions", which was both inaccurate and the original bug).

## Test plan

Original test plan (unchanged, still passing):

- [x] `HandleTaskCompletion` invokes `ErrorHandler` with `DaprException` wrapping the original exception
- [x] `HandleTaskCompletion` does not throw synchronously when no `ErrorHandler` is set (now: caches for next subscribe)
- [x] `HandleTaskCompletion` does not propagate when `ErrorHandler` itself throws (now: caches combined fault)
- [x] `HandleTaskCompletion` does not invoke `ErrorHandler` for successful tasks
- [x] `SubscribeAsync` throws `DaprException` wrapping `RpcException` when sidecar unavailable
- [x] `SubscribeAsync` allows retry after `RpcException` failure (`hasInitialized` reset)
- [x] `SubscribeAsync` resets `hasInitialized` for non-RPC exceptions (e.g. `ObjectDisposedException`)
- [x] Tests use xUnit v3 patterns (`TestContext.Current.CancellationToken`)

New end-to-end repro tests (drive the real `SubscribeAsync` → `ContinueWith` → `HandleTaskCompletion` chain via a mock `IAsyncStreamReader.MoveNext` that faults — not by awaiting `HandleTaskCompletion` directly):

- [x] `SubscribeAsync_WhenBackgroundFetchFaults_WithoutHandler_NextSubscribeRethrows` — the no-handler cache-and-rethrow path
- [x] `SubscribeAsync_WhenBackgroundFetchFaults_InvokesErrorHandlerExactlyOnce` — `hasFaulted` dedupe across the three continuations
- [x] `SubscribeAsync_AfterBackgroundFault_NextSubscribeRecreatesStream` — reset + retry obtains a fresh stream end-to-end (verified via `Mock.Verify(..., Times.Exactly(2))`)
- [x] `SubscribeAsync_WhenBackgroundFetchCancelled_DoesNotCachePendingFault` — cancellation short-circuit doesn't pollute the pending-fault slot
- [x] `SubscribeAsync_WhenErrorHandlerThrows_CachesCombinedFaultForNextSubscribe` — handler-faulted path produces an `AggregateException` with both the original `DaprException` and the handler's exception
- [x] Verified the new tests genuinely reproduce the bug: with the source fix temporarily reverted, 5 of 6 fail with the exact "background fault not cached" / synchronous-throw symptoms the fix addresses
- [x] All 68 tests pass on net8.0, net9.0, and net10.0 across multiple back-to-back local runs

🤖 Generated with [Claude Code](https://claude.com/claude-code)


GeorgeTsiokos:
**Correction to [previous comment](https://github.com/dapr/dotnet-sdk/pull/1744#issuecomment-4063869772):**

Item 3 (`SubscriptionErrorHandler` being synchronous-only) was incorrectly listed as a pre-existing issue — it was actually introduced in this PR. This has now been fixed: the delegate returns `Task` and `HandleTaskCompletion` is async to await it.

The truly pre-existing out-of-scope issues remain:

1. **`ProcessTopicChannelMessagesAsync` swallows `OperationCanceledException`** from the message handler — timeout-triggered cancellations are silently converted to the default response action with no logging.
2. **`FetchDataFromSidecarAsync` has a bare `catch (Exception)`** (line 351) that silently swallows channel-write failures, which could mask backpressure or disposal races.
3. **No `ILogger` anywhere in `PublishSubscribeReceiver`** — all the silent catches above exist because there's no logger to report to. Adding an `ILogger` parameter would let these be logged instead of swallowed.
4. **`ContinueWith` + `TaskScheduler.Default`** — the continuation pattern is fragile; a failed error handler's exception becomes unobserved. Modern `async/await` with a top-level try/catch would be more robust.

WhitWaldo:
@GeorgeTsiokos We're coming up on the 1.18 release. While this needn't absolutely be done by then as it's not runtime-dependent, if you have a chance to fix the merge conflicts, I'd be happy to circle back and get this reviewed and merged. Thank you!

WhitWaldo:
@GeorgeTsiokos Thank you very much for taking the time to put this PR together. As the idea to add support for this is excellent, but you've not responded to wrap this PR up in a while, I've gone ahead and rethought the approach a bit in a PR that supersedes this one.

Again, thank you for taking the time here and I hope you'll be willing to contribute again in the future!

PR #1745, opened 2026-03-13
Add optional parameter to configuration providers for non-blocking startup

### LINKED ISSUE #1748: Add optional parameter to configuration providers
## Describe the feature

Standard .NET configuration providers support an `optional` parameter that controls whether a missing or unavailable source is treated as an error or silently skipped. For example:

- `AddJsonFile("appsettings.json", optional: true)`
- `AddXmlFile("config.xml", optional: true)`
- `AddIniFile("settings.ini", optional: true)`

The Dapr configuration providers (`AddDaprConfigurationStore`, `AddStreamingDaprConfigurationStore`, and `AddDaprSecretStore`) do not support this pattern. When the Dapr sidecar is unavailable at startup, these providers throw instead of gracefully degrading.

### Proposed behavior

Add an `optional` parameter (default `false`) to all Dapr configuration provider extension methods:

- **`optional: false` (default)** — existing behavior; `Load()` blocks until the sidecar is available or throws on timeout
- **`optional: true`** — `Load()` returns immediately with an empty data set and retries in the background, signaling via `IChangeToken` once data is populated

This aligns the Dapr providers with the standard `IConfigurationProvider` contract and enables applications to start without requiring the sidecar to be available at boot time.

## Release Note

RELEASE NOTE: **ADD** `optional` parameter to `AddDaprConfigurationStore`, `AddStreamingDaprConfigurationStore`, and `AddDaprSecretStore` for non-blocking startup

GeorgeTsiokos:
## Summary

Resolves #1748

- Adds an `optional` parameter to `AddDaprConfigurationStore`, `AddStreamingDaprConfigurationStore`, and `AddDaprSecretStore` extension methods so applications can start without blocking on Dapr sidecar availability
- Hardens `LoadInBackgroundAsync` in both providers with narrowed exception handling (`OperationCanceledException` + `DaprException`) instead of a broad `catch (Exception)` that silently swallowed programming errors
- Fixes resource leak by disposing `CancellationTokenSource` in both providers
- Fixes `StringComparer` inconsistency (`InvariantCultureIgnoreCase` → `OrdinalIgnoreCase`) in `FetchSecretsAsync`
- Makes fields `readonly` in `DaprConfigurationStoreProvider` for consistency with `DaprSecretStoreConfigurationProvider`
- Restores UTF-8 BOM per `.editorconfig` and adds missing trailing newlines

## Test plan

- [x] All 52 existing + new unit tests pass across net8.0, net9.0, net10.0
- [x] New unit tests use `GetReloadToken()` callbacks for deterministic assertions (no fragile `Task.Delay` polling)
- [x] Integration test: `optional: true` with delayed sidecar — config populates once sidecar starts (`ShouldPopulateSecretsWhenSidecarStartsLate`)
- [x] Integration test: `optional: false` (default) behavior unchanged — `Build()` blocks until sidecar ready (`ShouldBlockAndPopulateSecretsWhenNotOptional`)
- [x] Integration test: disposal during background loading exits cleanly without hanging (`ShouldExitCleanlyWhenDisposedDuringBackgroundLoad`)
- [x] Integration tests updated to xUnit v3 (`TestContext.Current.CancellationToken` pattern)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


GeorgeTsiokos:
A few items I'm intentionally leaving as-is — noting them here for visibility:

**Pre-existing `catch (Exception)` in streaming subscription loop** (`DaprConfigurationStoreProvider.cs`) — The broad catch in the `SubscribeConfiguration` retry loop predates this PR. It's there to reconnect on any stream error (including grpc status errors that aren't `DaprException`). Left unchanged to avoid scope creep.

**`UnsubscribeConfiguration` throwing inside catch block** — Same pre-existing code, same reasoning.

**Exponential backoff** — The flat `sidecarWaitTimeout` delay between retries is intentional. The sidecar typically becomes available within its startup window; exponential backoff would add unnecessary complexity for this use case.

**Splitting the `try` block in `LoadInBackgroundAsync`** — The current structure is deliberately compact to keep the retry logic readable. Happy to revisit if the reviewer feels strongly.

**Non-optional blocking-failure test for config store** — Pre-existing gap (the non-optional path throws on WaitForSidecar timeout, but there's no unit test for it). Out of scope for this PR.

**`IsOptional` source/builder layer test** — The existing end-to-end integration tests in `SecretStoreOptionalTests.cs` cover this path. A pure unit test for the source layer would be redundant.

WhitWaldo:
Superseded by #1844

Issue #1728, opened 2026-03-03
[ACTORS]  [Feature Request] Unload state operation


olitomlinson:
## Describe the feature

it would be great to have an `UnloadStateAsync()` method on `StateManager`

This would only remove from key from the in-memory cache / tracking store, but not remove from the underlying store.

In memory-intensive applications, it would be great to be able unload keys from memory when they're not needed, and simply re-load them via the usual `GetStateAsync` / `TryGetStateAsync` methods.

**Risk**

Of course there is a risk that someone unloads a key, which, already had been modified but not yet persisted, so guards could be put in place to prevent that by default.

For example, assuming that Key `X` had been added/modified but not yet saved / flushed,

When calling `this.StateManager.UnloadStateAsync("X");`

I would expect this to throw an exception e.g. `"UnloadingModifiedState"` exception.

However, for use-cases where this is acceptable, an option could be provided which would suppress that exception

`this.StateManager.UnloadStateAsync("X", options = new UnloadStateOptions{ AllowUnloadingWhenStateModified = true} );`

---

**Example**

```c#

// add this big 1mb payload to the actor state
await this.StateManager.AddStateAsync("1mb-important-data", payload);

// flush to disk
await this.StateManager.SaveStateAsync();

// unload from memory, as we aren't going to need it until some point in the future
await this.StateManager.UnloadStateAsync("1mb-important-data");

// minutes / hour / days / weeks later - load the state 
var payload = await this.StateManager.GetStateAsync<object>("1mb-important-data");
```

WhitWaldo:
Closed as part of #1750

PR #1750, opened 2026-03-16
Implementation of UnloadStateAsync

olitomlinson:
# Description

Adds an `UnloadStateAsync()` method to the Actor StateManager, so that state can be removed from the cache, without being removed from disk

## Issue reference

https://github.com/dapr/dotnet-sdk/issues/1728

Please reference the issue this PR will close: #_[1728]_

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


WhitWaldo (commented):
At first pass, this looks great - I'll keep an eye out for when you pull it out of draft.

src/Dapr.Actors/Runtime/IActorStateManager.cs
```diff
@@ -24,6 +24,15 @@ namespace Dapr.Actors.Runtime;
...
 {
+    /// <summary>
+    /// Unloads the specified state from the in-memory cache/tracker, but does not remove it from the underlying store.
+    /// </summary>
+    /// <param name="stateName">Name of the actor state to unload.</param>
+    /// <param name="options">Options for unloading state (e.g., allow unloading modified state).</param>
+    /// <param name="cancellationToken">The token to monitor for cancellation requests.</param>
+    /// <returns>A task that represents the asynchronous unload operation.</returns>
+    /// <exception cref="InvalidOperationException">Thrown if the state is modified and not yet persisted, unless allowed by options.</exception>
+    Task UnloadStateAsync(string stateName, UnloadStateOptions options = null, CancellationToken cancellationToken = default);
```
GeorgeTsiokos, line 35: consider adding as a default interface method with a `NotImplementedException` / virtual behavior.

src/Dapr.Actors/Runtime/UnloadStateOptions.cs
```diff
@@ -0,0 +1,14 @@
+// Options for UnloadStateAsync operation
```
GeorgeTsiokos: missing license header

src/Dapr.Actors/Runtime/IActorStateManager.cs
olitomlinson: why?

WhitWaldo: In this scenario, adding a default interface method to throw `NotSupportedException` isn't a meaningful inclusion because we already provide the concrete type implementing this interface ourselves, and this will ship with same-day support. 

Especially as this project doesn't support dependency injection, I would not expect that developers have implemented their own types on this interface, modified the library to use them instead, update the package and see this new method on the interface and call it in their applications just to see a `TypeLoadException` because they didn't implement the method themselves - I just don't think that's likely. I wouldn't have expected them to wrap every actor invocation in a try/catch block either, so this would just shift throwing a `TypeLoadException` to an `NotSupportedException` - not a meaningful change here.

src/Dapr.Actors/Runtime/UnloadStateOptions.cs
WhitWaldo: @olitomlinson Could you please add the Dapr license header to this file?

WhitWaldo (changes requested):
One missing license header

src/Dapr.Actors/Runtime/UnloadStateOptions.cs
olitomlinson: done

WhitWaldo (approved):
Thank you!

marcduiker:
@holopin-bot @olitomlinson sdk-badge Thank you! Here's a digital badge as a small token of appreciation.

PR #1813, opened 2026-05-09
chore: Address flaky integration tests

WhitWaldo:
# Description

Especially now that we're adding integration tests to N-2 stable versions + RCs (if any), it means that several of the integration tests are now running against less-optimized versions of the runtime and missing performance test benchmarks or using features not yet supported. This simply adds a skip flag to those tests so they don't run on the older versions as they aren't supported.

## Issue reference

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


WhitWaldo:
@copilot I'm seeing that the Dapr.Workflow integration tests for 1.15 and 1.16 of the runtime appear to be getting stuck or timing out partway through and there's nothing in the log to indicate which test it might be. Can you please work out which workflow tests did not pass successfully then do an analysis as to which of those might be responsible for hanging so I might flag with them with a `[MinimumDaprRuntimeFact("1.17")]` attribute?

PR #1817, opened 2026-05-15
api: accept canonical gRPC URI form on the gRPC endpoint

lrascao:
# Description

The Dapr SDKs warn that http/https are deprecated as gRPC URI schemes — the prefix denotes a name resolver, not a transport.

- Add `DaprDefaults.NormalizeGrpcEndpoint`. Returns http/https unchanged; rewrites `dns://host:port?tls=true` to `https://host:port` and `?tls=false` to `http://host:port` before the URI reaches `GrpcChannel.ForAddress` (which is HttpClient-backed and does not understand the dns scheme). Throws on `dns://` without a parseable `tls=true|false` and on any other scheme.
- `DaprClientBuilder.Build` and `DaprGenericClientBuilder .BuildDaprClientDependencies` now normalize the gRPC endpoint up front and feed the normalized URI to both the http/https switch and `GrpcChannel.ForAddress`. The redundant scheme check on the gRPC path is gone; HTTP endpoint validation is unchanged.
- `GetDefaultGrpcEndpoint` normalizes the env-var value before handing it to `BuildEndpoint` (otherwise `UriBuilder` would strip the `?tls=` query while reassembling).
- `UseGrpcEndpoint` docstrings now mention the canonical form.
- Tests: unit coverage for `NormalizeGrpcEndpoint` (incl. case-insensitive `TLS=True`, missing tls, unparseable tls, unsupported scheme), env-var path via `DAPR_GRPC_ENDPOINT=dns://...`, and dns acceptance / dns- without-tls rejection on both `DaprClientBuilder` and `DaprJobsClientBuilder` (which exercises the shared generic builder used by Workflow/Messaging/AI/etc). HTTP endpoints and `InvocationHandler` are unchanged.

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


WhitWaldo:
@lrascao Could you please add some integration tests demonstrating that this doesn't hit a wall when it encounters the runtime? You're right that from the client side, there's more than can be done here (as you did), but for the most part, the SDK is generally just a wrapper for functionality exposed by the runtime. So while yes, the client using `HttpClient` doesn't care what the prefix is, I don't know that's going to be true of the runtime that is then connected to.

WhitWaldo:
@lrascao Also, the warning about HTTP is because the SDKs are transitioning to only connect to the runtime via gRPC, so the intent is that developers have some heads up that they need to specify a gRPC endpoint/port for their apps so callbacks work (if using those building blocks).

PR #1818, opened 2026-05-16
bugfix: Workflow overhaul, new feature bugfix

WhitWaldo:
# Description

Fix for the workflow propagation integration tests not working as expected.

## Issue reference

We strive to have all PR being opened based on an issue, where the problem or feature have been discussed prior to implementation.

Please reference the issue this PR will close: #_[issue number]_

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Extended the documentation

PR #1802, opened 2026-05-02
Feature: Workflow history propagation

WhitWaldo:
# Description

Added support for workflow history propagation + unit/integration tests (>80% test coverage)

## Issue reference

We strive to have all PR being opened based on an issue, where the problem or feature have been discussed prior to implementation.

Please reference the issue this PR will close: #1801 
## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [X] Code compiles correctly
* [X] Created/updated tests
* [ ] Extended the documentation

PR #1825, opened 2026-05-20
feat(workflow): align history propagation API with go-sdk

### LINKED ISSUE #1801: Add support for workflow history propagation
## Background

Dapr Workflows now supports propagating execution history from a parent workflow to its child workflows and activities. This lets downstream code inspect what happened upstream — useful for chain-of-custody verification, fraud detection, audit, and AI-agent context flow.

Two scopes are exposed:

- **`PropagateLineage`** — caller's own events plus the full ancestor chain
- **`PropagateOwnHistory`** — caller's events only (a trust boundary; ancestors are dropped)

The Go path is implemented end-to-end. To use this feature from a .NET app, this SDK needs the matching API.

## Proto surface (already merged in the proto repo)

From `dapr/durabletask-protobuf` ([PR #38](https://github.com/dapr/durabletask-protobuf/pull/38)):

- `history_events.proto`
  - `TaskScheduledEvent.historyPropagationScope` (optional, field 7)
  - `ChildWorkflowInstanceCreatedEvent.historyPropagationScope` (optional, field 7)
  - new `PropagatedHistory { repeated HistoryEvent events; HistoryPropagationScope scope; repeated PropagatedHistoryChunk chunks; }`
  - new `PropagatedHistoryChunk { string appId; int32 startEventIndex; int32 eventCount; string instanceId; string workflowName; }`
- `orchestration.proto`
  - new `HistoryPropagationScope` enum: `HISTORY_PROPAGATION_SCOPE_NONE = 0`, `HISTORY_PROPAGATION_SCOPE_OWN_HISTORY = 1`, `HISTORY_PROPAGATION_SCOPE_LINEAGE = 2`
- `orchestrator_actions.proto`
  - `ScheduleTaskAction.historyPropagationScope` (optional, field 6)
  - `CreateChildWorkflowAction.historyPropagationScope` (optional, field 6)

### 1. Proto regen + plumbing

- Pull the updated `.proto` files from `dapr/durabletask-protobuf` and regenerate the C# protobuf classes
- Plumb the new `historyPropagationScope` field through the schedule-task / create-child-workflow code paths
- Surface the received `PropagatedHistory` to the workflow runtime so it can be handed to user code

### 2. User-facing API to add 

**Schedule-side (parent workflow author):**
- `WithHistoryPropagation(HistoryPropagationScope scope)` extension/option on the activity-call and child-workflow-call APIs
- `HistoryPropagationScope.Lineage` and `HistoryPropagationScope.OwnHistory` static members

**Receive-side (child workflow / activity author):**
- `GetPropagatedHistory()` returning `PropagatedHistory?` on `IWorkflowContext` and the activity context
- `PropagatedHistory` record/class with:
  - `Events` — list of `HistoryEvent`
  - `Scope` — the scope enum value
  - `Workflows` — chunks (each with `AppId`, `WorkflowName`, `InstanceId`, `StartEventIndex`, `EventCount`)
  - `AppIDs` — deduplicated list of app IDs in the chain
  - convenience filters: `GetWorkflowByName(string)`, `GetEventsByAppID(string)`, `GetEventsByInstanceID(string)`, `GetEventsByWorkflowName(string)`
  - throws `PropagationNotFoundException` (or returns `null`) for missing names

### 3. Example to add

`examples/WorkflowHistoryPropagation/` mirroring the Go demo: a 3-tier payment workflow (parent → activity → child workflow with `Lineage` → activity with `OwnHistory`). Standalone `dapr run` mode at minimum; K8s + signing variant a stretch.

### 4. Tests

At least one test per scope:
- **Lineage:** parent → child receives parent's events
- **OwnHistory:** parent → child receives caller's events but NOT grandparent's

## References

- `dapr/dapr` runtime PR: https://github.com/dapr/dapr/pull/9810
- `dapr/durabletask-go` PR: https://github.com/dapr/durabletask-go/pull/85
- `dapr/durabletask-protobuf` PR: https://github.com/dapr/durabletask-protobuf/pull/38
- `dapr/go-sdk` reference implementation: https://github.com/dapr/go-sdk/pull/823
- Go SDK example: `examples/workflow-history-propagation/`

nelson-parente:
## Summary

Aligns the workflow history propagation surface added in #1802 / #1818 with the canonical go-sdk shape — the same one python-sdk just adopted in [python-sdk#1047](https://github.com/dapr/python-sdk/pull/1047). [@cicoyle](https://github.com/cicoyle) (durabletask-go author) flagged the divergence post-merge while building 1.18 quickstarts; this PR delivers what issue #1801 originally described.

This brings .NET in line for cross-SDK parity before 1.18 ships.

### Three concrete gaps closed

**1. Activity-level opt-in (was missing entirely)**

Before this PR, `PropagationScope` only lived on `ChildWorkflowTaskOptions` — there was no way to propagate history when scheduling an activity. The go-sdk reference example uses exactly this pattern (`SettlePayment` activity with `PropagateOwnHistory`), so the cross-SDK quickstart literally couldn't be ported.

- `PropagationScope` moved to base `WorkflowTaskOptions`.
- `WithHistoryPropagation()` extension method available on both `WorkflowTaskOptions` and `ChildWorkflowTaskOptions`.
- `scheduleTaskAction.HistoryPropagationScope` wired in `WorkflowOrchestrationContext.CallActivityInternalAsync`, matching the existing child-workflow path.

**2. Read API replaced (was lossy filters + minimal events)**

The old `PropagatedHistoryEvent` only kept `(EventId, Kind, Timestamp)` — input/output/failure payloads were thrown away by `ConvertChunk`, making the high-level audit/fraud-detection patterns from the Go/Python quickstarts impossible.

- Removed: the old minimal `PropagatedHistoryEvent` and `HistoryEventKind` (clean break — 1.18 unreleased).
- On `PropagatedHistory`: `GetEntries()`, `FilterByWorkflowName()`, `TryGetLastWorkflowByName()`, `GetAppIds()`, `FilterByAppId()`, `FilterByInstanceId()`. The `FilterBy*` names are kept for forward-compatibility (per review — propagated history may carry non-workflow entries in future runtimes), but now return the richer `PropagatedHistoryEntry` values rather than the old lossy event view.
- New `PropagatedHistoryEntry` class (one per ancestor workflow) with `TryGetLastActivityByName()`, `GetActivitiesByName()`, `TryGetLastChildWorkflowByName()`, `GetChildWorkflowsByName()` — mirrors the `GetLast*ByName` rename merged in [durabletask-go#105](https://github.com/dapr/durabletask-go/pull/105), adapted to the idiomatic .NET `TryGet*` pattern with `[NotNullWhen(true)] out` parameters.
- New `PropagatedHistoryActivityResult(Name, Started, Completed, Failed, Input, Output, FailureDetails)` record — matches the Go SDK's [`ActivityResult` struct](https://github.com/dapr/durabletask-go/blob/main/api/propagation.go) and Python's `ActivityResult` dataclass.
- New `PropagatedHistoryChildWorkflowResult` with the equivalent shape.
- Both result records also expose a computed `Status` (`PropagatedHistoryTaskStatus.Pending`/`Completed`/`Failed`) projected from the flags (per review), for callers that prefer to `switch` on a single lifecycle value; the `Started`/`Completed`/`Failed` flags are retained for go-sdk/python-sdk field parity.
- Lookups by name and app ID are case-insensitive, matching how `WorkflowsFactory` registers workflow/activity names and how the SDK matches app IDs elsewhere.

**3. Event payload preserved internally**

`ConvertChunk` now walks the raw `HistoryEvent`s, pre-indexes `TaskCompleted` / `TaskFailed` (and the child-workflow counterparts) by `TaskScheduledId` in a single pass — so each scheduled activity or child resolves in O(1) — and produces fully-populated `PropagatedHistoryActivityResult` / `PropagatedHistoryChildWorkflowResult` records. SDK retries reuse `TaskExecutionId`, so we match on the scheduling event ID (matching Go/Python semantics). The proto `HistoryEvent` type is never exposed publicly — resolution happens at construction time inside `Dapr.Workflow`.

### Usage example (matches the go-sdk reference)

\`\`\`csharp
// Parent workflow propagates lineage to a child workflow and own-history to an activity
var fraudResult = await context.CallChildWorkflowAsync<FraudCheckResult>(
    nameof(FraudDetection),
    input: req,
    options: new ChildWorkflowTaskOptions()
        .WithHistoryPropagation(HistoryPropagationScope.Lineage));

await context.CallActivityAsync<SettlementResult>(
    nameof(SettlePayment),
    input: req,
    options: new WorkflowTaskOptions()
        .WithHistoryPropagation(HistoryPropagationScope.OwnHistory));

// Receive side — query specific events by name with TryGet*
var history = context.GetPropagatedHistory();
if (history is not null &&
    history.TryGetLastWorkflowByName("MerchantCheckout", out var merchant) &&
    merchant.TryGetLastActivityByName("ValidateMerchant", out var validate))
{
    if (validate.Completed) { /* inspect validate.Output */ }
}
\`\`\`

## Non-goals

- **No example app in this PR.** The cross-SDK quickstart lives in [dapr/quickstarts#1310](https://github.com/dapr/quickstarts/pull/1310) and will be refreshed against this API once it merges.
- **No proto / runtime changes.** All proto fields already exist (`ScheduleTaskAction.historyPropagationScope` field 6, `CreateChildWorkflowAction.historyPropagationScope` field 6); we just had to wire the activity path.

## Breaking changes (1.18 unreleased)

The propagation API shipped in #1802/#1818 has not been released. This PR replaces it cleanly rather than carrying `[Obsolete]` shims for an API that has no users yet. Removed types: the old minimal `PropagatedHistoryEvent` and `HistoryEventKind`. (`PropagatedHistoryEntry` is reused as the new per-ancestor entry type with a richer shape; the `PropagatedHistory.FilterByAppId/InstanceId/WorkflowName` method names are retained but now return `PropagatedHistoryEntry` values.)

## Test plan

- [x] `WorkflowHistoryPropagationTests.cs` rewritten end-to-end: activity resolution (completed/failed/pending/retried), child-workflow resolution, `Status` projection (incl. Failed-over-Completed precedence), history helpers (GetAppIds, TryGetLastWorkflowByName, FilterByWorkflowName, case-insensitive name/AppId lookups), scheduling option records (both base + derived), activity-level + child-workflow outbound action scope wiring.
- [x] `HistoryPropagationWorkflowTests.cs` (integration) updated to `GetEntries().Count`.
- [ ] **`dotnet build` + `dotnet test` to be verified by CI** — couldn't build locally (no .NET 10 SDK on dev machine). Sweep confirmed no stale references to removed types in `src/` or `test/`.

## References

- Closes #1801
- go-sdk reference impl + example: [dapr/go-sdk#823](https://github.com/dapr/go-sdk/pull/823)
- go-sdk `GetLast*ByName` rename: [dapr/durabletask-go#105](https://github.com/dapr/durabletask-go/pull/105)
- python-sdk parallel: [dapr/python-sdk#1047](https://github.com/dapr/python-sdk/pull/1047)
- java-sdk parallel: [dapr/java-sdk#1739](https://github.com/dapr/java-sdk/pull/1739)

cc @cicoyle @WhitWaldo


cicoyle (dismissed):

WhitWaldo:
@nelson-parente I'm not inclined to accept the type name changes. For types that are exclusively used for workflow history propagation, it makes sense that they'd reflect that in the name. For example, there's no such thing as a `WorkflowResult` except in the context of history propagation - its more accurately described as a `PropagatedHistoryEvent` as originally named.

That said, completely agree about the activity propagation oversight. 

I'll review this later this afternoon for specific notes, but I wanted to call this out for posterity.

src/Dapr.Workflow.Abstractions/PropagatedHistoryActivityResult.cs
WhitWaldo, line 1: As I said in the comment, `ActivityResult` isn't helpful, I don't think, because users provide their own types for activity results. Rather, this would be more aptly named as a `PropagatedHistoryActivityResult`.

src/Dapr.Workflow.Abstractions/ChildWorkflowResult.cs
```diff
@@ -0,0 +1,34 @@
...
+/// <param name="Name">The scheduled name of the child workflow.</param>
+/// <param name="Started">Whether the child workflow was scheduled in the propagated chunk.</param>
+/// <param name="Completed">Whether the child workflow completed successfully.</param>
+/// <param name="Failed">Whether the child workflow failed.</param>
+/// <param name="Output">The JSON-encoded output payload, or <c>null</c> when the child workflow has not completed.</param>
+/// <param name="FailureDetails">The failure details when <paramref name="Failed"/> is true, otherwise <c>null</c>.</param>
+/// <remarks>
+/// Mirrors the <c>ChildWorkflowResult</c> type in the Go and Python SDKs.
+/// </remarks>
+public sealed record ChildWorkflowResult(
```
WhitWaldo: Again, workflows (child or not) return user-defined types, so this isn't helpfully named. As the original implementation didn't have activity support, I think there's a great opportunity here to opt for a rename from the original and instead call this `PropgatedHistoryChildWorkflowResult`. 

I'm not terribly enthused by including "Child" in the type name as a workflow is a workflow whether it has a parent or not - that's just a provenance question. In other words, if a child workflow doesn't do something distinctly different from any other workflow, I don't entirely follow why we would call it something else (and I extend this argument to the "Detached Workflows" - they're just workflows that don't have a parent).

But I digress - in the interest of uniformity to the gRPC prototype descriptions, I'm fine calling this `PropgatedHistoryChildWorkflowResult`.

src/Dapr.Workflow.Abstractions/WorkflowHistory.cs
```diff
@@ -0,0 +1,46 @@
...
+/// Scheduling-side helpers for workflow history propagation, mirroring the
+/// <c>workflow.PropagateLineage()</c> / <c>workflow.PropagateOwnHistory()</c>
+/// factories in the Go SDK.
+/// </summary>
+/// <remarks>
+/// Both forms are equivalent: <c>options.WithHistoryPropagation(WorkflowHistory.PropagateLineage())</c>
+/// and <c>options.WithHistoryPropagation(HistoryPropagationScope.Lineage)</c> produce the same scope.
+/// The factory helpers exist for cross-SDK call-site parity.
+/// </remarks>
+public static class WorkflowHistory
```
WhitWaldo: This entire type is unnecessary. Just use `HistoryPropagationScope.Lineage` and `HistoryPropagationScope.OwnHistory`. We don't need more types just for the sake of having them.

src/Dapr.Workflow.Abstractions/WorkflowResult.cs
```diff
@@ -0,0 +1,134 @@
...
+/// <summary>
+/// A scoped view of a single workflow's chunk in propagated history.
+/// </summary>
+/// <remarks>
+/// Mirrors the <c>WorkflowResult</c> type in the Go SDK. Use
+/// <see cref="GetLastActivityByName"/> / <see cref="GetLastChildWorkflowByName"/>
+/// to query specific items inside this chunk; the plural <c>Get*sByName</c>
+/// variants return every occurrence in execution order.
+/// </remarks>
+public sealed class WorkflowResult
```
WhitWaldo: I find the name of the type confusing, especially for something presented as a public type - workflow results are types defined by the user and this is not a result of a workflow - it's something specific to propagated histories and should be named accordingly. I would propose the original `PropogatedHistoryEntry` as it's one view of potentially many in a broader history.

src/Dapr.Workflow.Abstractions/WorkflowResult.cs
```diff
@@ -0,0 +1,134 @@
...
+
+    /// <summary>
+    /// Initializes a new <see cref="WorkflowResult"/>.
+    /// </summary>
+    /// <param name="instanceId">The instance ID of the ancestor workflow.</param>
+    /// <param name="appId">The Dapr App ID that ran the ancestor workflow.</param>
+    /// <param name="name">The name of the ancestor workflow.</param>
+    /// <param name="activities">Activities resolved from this chunk, in execution order.</param>
+    /// <param name="childWorkflows">Child workflows resolved from this chunk, in execution order.</param>
+    public WorkflowResult(
```
WhitWaldo: I'd rather see this implemented using a primary constructor for conciseness and SDK uniformity as nothing is happening in the constructor outside of validation.

src/Dapr.Workflow.Abstractions/WorkflowResult.cs
```diff
@@ -0,0 +1,134 @@
...
+            .ToList();
+    }
+
+    /// <summary>
+    /// Returns the most recent activity in this chunk whose name matches.
+    /// </summary>
+    /// <param name="name">The activity name to look up.</param>
+    /// <returns>The last matching activity.</returns>
+    /// <exception cref="PropagationNotFoundException">No activity with the given name is present in this chunk.</exception>
+    public ActivityResult GetLastActivityByName(string name)
```
WhitWaldo: Might we instead implement this as a TryGet pattern? `public bool TryGetLastActivityByName(string name, out PropagatedHistoryActivityResult? result)`? No point throwing exceptions if we don't have to and there's nothing asynchronous in this method.

src/Dapr.Workflow.Abstractions/WorkflowResult.cs
```diff
@@ -0,0 +1,134 @@
...
+            .ToList();
+    }
+
+    /// <summary>
+    /// Returns the most recent child workflow in this chunk whose name matches.
+    /// </summary>
+    /// <param name="name">The child workflow name to look up.</param>
+    /// <returns>The last matching child workflow.</returns>
+    /// <exception cref="PropagationNotFoundException">No child workflow with the given name is present in this chunk.</exception>
+    public ChildWorkflowResult GetLastChildWorkflowByName(string name)
```
WhitWaldo: Might we instead implement this as a TryGet pattern? `public bool TryGetLastChildWorkflowByName(string name, out PropogatedHistoryChildWorkflowResult? result)`? No point throwing exceptions if we don't have to and there's nothing asynchronous in this method.

src/Dapr.Workflow.Abstractions/WorkflowResult.cs
WhitWaldo, line 1: Back to the naming thing, many of the XML comments throughout this type refer to this as a "chunk". I'd rather call this an entry (keeping with HistoryPropagationEntry than a chunk as we don't use the word "chunk" anywhere else in public documentation. Consistent names here and in the documentation help future maintainers and developers understand related concepts. If this were a private type, that'd be one thing - but it's not, so we should strive for clarity over internal description.

src/Dapr.Workflow.Abstractions/WorkflowContext.cs
```diff
@@ -341,15 +341,16 @@ public virtual Task CallChildWorkflowAsync(
     /// specified a <see cref="HistoryPropagationScope"/> other than <see cref="HistoryPropagationScope.None"/>.
     /// </para>
     /// <para>
-    /// Use <see cref="PropagatedHistory.FilterByAppId"/>, <see cref="PropagatedHistory.FilterByInstanceId"/>,
-    /// or <see cref="PropagatedHistory.FilterByWorkflowName"/> to narrow down the returned entries.
+    /// Use <see cref="PropagatedHistory.GetLastWorkflowByName"/> and
+    /// <see cref="WorkflowResult.GetLastActivityByName"/> / <see cref="WorkflowResult.GetLastChildWorkflowByName"/>
+    /// to query specific items from the chain. The plural <c>Get*sByName</c> variants return every match.
```
WhitWaldo: What is a chain? There's absolutely no reason to introduce brand new terminology in public XML documentation.

src/Dapr.Workflow.Abstractions/WorkflowContext.cs
```diff
@@ -341,15 +341,16 @@ public virtual Task CallChildWorkflowAsync(
...
     /// </para>
     /// <para>
     /// This method always returns the same value regardless of whether the workflow is replaying.
     /// </para>
     /// </remarks>
     /// <returns>
-    /// A <see cref="PropagatedHistory"/> containing entries from ancestor workflows,
+    /// A <see cref="PropagatedHistory"/> containing the chain of ancestor workflows,
```
WhitWaldo: Again, what's a chain? We have a list of entries. Let's not introduce new names of things in public documentation unless there's a good reason for it.

src/Dapr.Workflow.Abstractions/PropagatedHistory.cs
```diff
@@ -18,68 +18,112 @@ namespace Dapr.Workflow;
...
+    {
+        ArgumentException.ThrowIfNullOrWhiteSpace(appId);
+        return _workflows
+            .Where(w => string.Equals(w.AppId, appId, StringComparison.Ordinal))
+            .ToList();
     }
 
     /// <summary>
-    /// Returns a new <see cref="PropagatedHistory"/> containing only entries for the specified workflow name.
+    /// Returns every workflow chunk produced by the given instance, in execution order.
```
WhitWaldo: Let's avoid chunk please and stick with "event"

src/Dapr.Workflow.Abstractions/PropagatedHistory.cs
```diff
@@ -18,68 +18,112 @@ namespace Dapr.Workflow;
...
+            }
+        }
+
+        return result;
     }
 
     /// <summary>
-    /// Returns a new <see cref="PropagatedHistory"/> containing only the entry with the specified instance ID.
+    /// Returns every workflow whose name matches, in execution order. Useful when the
+    /// chain contains the same name more than once (e.g. recursion or ContinueAsNew).
```
WhitWaldo: I mention this later - this is public XML documentation. We should not be introducing new concepts like "chain" randomly when "list of events" perfectly describes what we're doing already.

src/Dapr.Workflow.Abstractions/PropagatedHistory.cs
```diff
@@ -18,68 +18,112 @@ namespace Dapr.Workflow;
...
-    /// correspond to progressively older ancestors when <see cref="HistoryPropagationScope.Lineage"/> is used.
+    /// Returns every workflow chunk in the chain, in execution order
+    /// (ancestor first, immediate parent last).
     /// </summary>
-    public IReadOnlyList<PropagatedHistoryEntry> Entries => _entries;
+    public IReadOnlyList<WorkflowResult> GetWorkflows() => _workflows;
 
     /// <summary>
-    /// Returns a new <see cref="PropagatedHistory"/> containing only entries from the specified App ID.
+    /// Returns an ordered, deduplicated list of app IDs in the propagated chain.
```
WhitWaldo: "chain" -> "list of propagated history events"

src/Dapr.Workflow.Abstractions/PropagatedHistory.cs
WhitWaldo, line 1: As in other classes, let's not introduce vague new terminology like "chunks" or "chains" in public-facing XML documentation when we already have terms that accurately and more obviously describe what they actually refer to. Such descriptions might be perfectly fine in internal implementations as they reflect usage from the gRPC protos themselves, but this is a public type and there should be nothing in here that doesn't have an analog directly to the publicly accessible Dapr runtime documentation.

src/Dapr.Workflow.Abstractions/HistoryEventKind.cs
WhitWaldo, line 1: The [protos](https://github.com/dapr/durabletask-protobuf/blob/29ede9434bcf3136784268caefcdc165693a0d9f/protos/history_events.proto#L227) describe the type of history event returned. Is this not something meaningful to provide to the user?

src/Dapr.Workflow.Abstractions/PropagatedHistoryActivityResult.cs
```diff
@@ -0,0 +1,38 @@
...
+/// quickstarts and chain-of-custody patterns line up.
+/// </remarks>
+public sealed record ActivityResult(
+    string Name,
+    bool Started,
+    bool Completed,
+    bool Failed,
+    string? Input,
+    string? Output,
+    WorkflowTaskFailureDetails? FailureDetails);
```
WhitWaldo: No interest in reflecting what type of event was recorded here? It's in the protos.

src/Dapr.Workflow.Abstractions/PropagatedHistoryChildWorkflowResult.cs
```diff
@@ -0,0 +1,34 @@
...
+/// <remarks>
+/// Mirrors the <c>ChildWorkflowResult</c> type in the Go and Python SDKs.
+/// </remarks>
+public sealed record ChildWorkflowResult(
+    string Name,
+    bool Started,
+    bool Completed,
+    bool Failed,
+    string? Output,
+    WorkflowTaskFailureDetails? FailureDetails);
```
WhitWaldo: No interest in reflecting what type of event was recorded here? It's in the protos.

src/Dapr.Workflow/Worker/Internal/WorkflowOrchestrationContext.cs
```diff
@@ -989,72 +1002,141 @@ private static WorkflowTaskFailedException CreateTaskFailedException(TaskFailedE
...
-            HistoryEvent? historyEvent;
             try
             {
-                historyEvent = HistoryEvent.Parser.ParseFrom(rawEvent);
+                events.Add(HistoryEvent.Parser.ParseFrom(rawEvent));
             }
             catch (InvalidProtocolBufferException)
             {
-                continue;
+                // Skip malformed events; a single bad event cannot poison the chunk.
```
WhitWaldo: This seems like something the SDK should _absolutely_ throw on if the runtime is throwing bad events to the SDK, right? Not just swallow the exception and pretend nothing is wrong here.

WhitWaldo (changes requested):
Thank you for putting this PR together. 

As I said in the earlier comment and repeat throughout, there's a distinction to me in terminology we use internally within the implementation of the project (in the gRPC protos, in the runtime) and the terminology we use in public-facing documentation. The protos refer to chunks and I've got no problem extending that to private methods, but most of this PR exposes public types with public-facing XML documentation - the names should be clear as to purpose and not introduce _anything_ not present in publicly available Dapr documentation (like "chains" or "chunks"), especially when far more specific and particular names are already exposed in the implementation.

All of this work is scoped to history propagation - the names of these types should reflect this accordingly. It'd be one thing if you put them in a namespace like `Dapr.Workflow.Abstractions.ProgagatedHistory` and named them as such, but this isn't the long-standing Dapr .NET convention - most types are just left in the project root. So in this case, vague types like `ActivityResult` or `ChildWorkflowResult` just aren't clear. A developer shouldn't have to look at the XML documentation to understand that this isn't actually describing the returned result from a workflow or a child workflow - the name should be self-documenting, so I've proposed better alternatives like `PropagatedHistoryActivityResult` and `PropagatedHistoryChildWorkflowResult`.

The kind of event from the history was dropped in this implementation. Yes, it's not mentioned in #1801 but doesn't it provide some value to the user regarding what the event represents (e.g. activity started or completed)?

I certainly do appreciate you identifying the gap between supporting the activity propagation in addition to the existing workflow implementation.

Please do let me know if you have any follow-up questions.

src/Dapr.Workflow.Abstractions/PropagatedHistoryActivityResult.cs
nelson-parente: Done — renamed to `PropagatedHistoryActivityResult` in b6042200.

src/Dapr.Workflow.Abstractions/ChildWorkflowResult.cs
nelson-parente: Done — renamed to `PropagatedHistoryChildWorkflowResult` in b6042200.

src/Dapr.Workflow.Abstractions/WorkflowHistory.cs
nelson-parente: Done — `WorkflowHistory` removed; callers use `HistoryPropagationScope.Lineage` / `OwnHistory` directly.

src/Dapr.Workflow.Abstractions/WorkflowResult.cs
nelson-parente: Done — renamed to `PropagatedHistoryEntry` in b6042200.

src/Dapr.Workflow.Abstractions/WorkflowResult.cs
nelson-parente: Done — `PropagatedHistoryEntry` now uses a primary constructor.

src/Dapr.Workflow.Abstractions/WorkflowResult.cs
nelson-parente: Done — `TryGetLastActivityByName(string, out PropagatedHistoryActivityResult?)`.

src/Dapr.Workflow.Abstractions/WorkflowResult.cs
nelson-parente: Done — `TryGetLastChildWorkflowByName(string, out PropagatedHistoryChildWorkflowResult?)`.

src/Dapr.Workflow.Abstractions/WorkflowResult.cs
nelson-parente: Done — public XML docs use "entry" throughout.

src/Dapr.Workflow.Abstractions/WorkflowContext.cs
nelson-parente: Done — rewrote to reference the list of `PropagatedHistoryEntry` values directly. No more "chain".

src/Dapr.Workflow.Abstractions/WorkflowContext.cs
nelson-parente: Done — same rewrite; no "chain" in public docs.

src/Dapr.Workflow.Abstractions/PropagatedHistory.cs
nelson-parente: Done — XML uses "event" / "entry" now, never "chunk".

src/Dapr.Workflow.Abstractions/PropagatedHistory.cs
nelson-parente: Done — replaced with "list of propagated history entries".

src/Dapr.Workflow.Abstractions/PropagatedHistory.cs
nelson-parente: Done — applied verbatim.

src/Dapr.Workflow.Abstractions/PropagatedHistory.cs
nelson-parente: Done — public XML scrubbed of "chunks" and "chains" across the abstractions.

src/Dapr.Workflow.Abstractions/HistoryEventKind.cs
nelson-parente: Lifecycle stage is captured by `Started` / `Completed` / `Failed` booleans on each result record — covers pending (Started && !Completed && !Failed) and matches go-sdk / python-sdk. `HistoryEventKind` removed.

src/Dapr.Workflow.Abstractions/PropagatedHistoryActivityResult.cs
nelson-parente: Captured via `Started` / `Completed` / `Failed` booleans rather than a separate event-kind enum, matching go-sdk `ActivityResult` and python-sdk.

src/Dapr.Workflow.Abstractions/PropagatedHistoryChildWorkflowResult.cs
nelson-parente: Same as `PropagatedHistoryActivityResult` — `Started` / `Completed` / `Failed` cover the lifecycle without a separate enum, matching go-sdk / python-sdk.

src/Dapr.Workflow/Worker/Internal/WorkflowOrchestrationContext.cs
nelson-parente: Done — the try / catch around `HistoryEvent.Parser.ParseFrom` is gone; malformed event bytes now surface as exceptions. See b6042200.

src/Dapr.Workflow.Abstractions/PropagatedHistory.cs
```diff
@@ -15,71 +15,120 @@ namespace Dapr.Workflow;
...
+/// the last entry is the immediate parent.
+/// <para>
+/// Use the <c>Get*</c> / <c>TryGet*</c> methods to walk the list by app, instance, or workflow name.
+/// Mirrors the <c>PropagatedHistory</c> type in the Go and Python SDKs.
+/// </para>
 /// </remarks>
 public sealed class PropagatedHistory
 {
-    private readonly IReadOnlyList<PropagatedHistoryEntry> _entries;
+    private readonly IReadOnlyList<PropagatedHistoryEntry> _workflows;
```
WhitWaldo: From a maintainer perspective, this seems murky - if the type name is "PropgatatedHistoryEntry" and might reflect either a workflow or activity entry, why would you call them "_workflows" instead of "_entries"? This was an unnecessary change - please revert.

src/Dapr.Workflow.Abstractions/HistoryEventKind.cs
WhitWaldo: It's premature to think that this might be the final state of the lifecycle functionality. A wall of flags in which one is only ever true seems cognitively confusing. I can anticipate developers preemptively asking if there's ever a situation in which more than one flag could ever be true at once - as this reflects a specific named lifecycle, no, in which case, it feels like a simpler effect here is to simply reflect the lifecycle enums of the original implementation. They needn't all be provided as only a subset are ever exposed here, but the idea is to think about how the SDK will be consumed most easily by developers, not just what's one of many ways to expose the concept. 

Further, should additional lifecycle events be added, a wall of flag evaluations must be extended to properly rule out new lifecycle flags. But if the new enums aren't necessary, they can simply be ignored.

All told, let's use enums instead of flags.

src/Dapr.Workflow.Abstractions/PropagatedHistory.cs
WhitWaldo: Do we know for a fact that propagated history will _never_ include anything other than workflows? I don't know that and we have no roadmap indicating if this feature will evolve in the future. As this is a public method, I would rather retain the more generic "GetEntries()" name to keep our option open without introducing future breaking changes.

src/Dapr.Workflow.Abstractions/PropagatedHistory.cs
```diff
@@ -15,71 +15,120 @@ namespace Dapr.Workflow;
...
+    /// <returns>An empty list when no match is found.</returns>
+    public IReadOnlyList<PropagatedHistoryEntry> GetWorkflowsByName(string name)
     {
-        ArgumentException.ThrowIfNullOrWhiteSpace(instanceId);
-        return new PropagatedHistory(
-            _entries.Where(e => string.Equals(e.InstanceId, instanceId, StringComparison.Ordinal)).ToList());
+        ArgumentException.ThrowIfNullOrWhiteSpace(name);
+        return _workflows
+            .Where(w => string.Equals(w.Name, name, StringComparison.Ordinal))
+            .ToList();
```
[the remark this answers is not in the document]
WhitWaldo: We should definitely do this.

src/Dapr.Workflow.Abstractions/PropagatedHistory.cs
```diff
@@ -15,71 +15,120 @@ namespace Dapr.Workflow;
...
+    /// Returns every workflow entry produced by the given app, in execution order.
+    /// </summary>
+    /// <param name="appId">The Dapr App ID to filter by.</param>
+    /// <returns>An empty list when no match is found.</returns>
+    public IReadOnlyList<PropagatedHistoryEntry> GetWorkflowsByAppId(string appId)
+    {
+        ArgumentException.ThrowIfNullOrWhiteSpace(appId);
+        return _workflows
+            .Where(w => string.Equals(w.AppId, appId, StringComparison.Ordinal))
+            .ToList();
```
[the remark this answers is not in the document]
WhitWaldo: This too

src/Dapr.Workflow.Abstractions/PropagatedHistory.cs
```diff
@@ -15,71 +15,120 @@ namespace Dapr.Workflow;
...
-    /// Returns a new <see cref="PropagatedHistory"/> containing only entries for the specified workflow name.
+    /// Returns every workflow entry produced by the given instance, in execution order.
+    /// Usually a single entry, except when the same instance reappears via ContinueAsNew.
     /// </summary>
-    /// <param name="workflowName">The workflow name to filter by.</param>
-    /// <returns>A filtered <see cref="PropagatedHistory"/> instance.</returns>
-    public PropagatedHistory FilterByWorkflowName(string workflowName)
+    /// <param name="instanceId">The workflow instance ID to filter by.</param>
+    /// <returns>An empty list when no match is found.</returns>
+    public IReadOnlyList<PropagatedHistoryEntry> GetWorkflowsByInstanceId(string instanceId)
```
WhitWaldo: Same idea as before - I don't know that this feature will always return only workflows. Today it might, but I'd like to leave our options open so we can avoid breaking changes. Please revert to "FilterByWorkflowName".

WhitWaldo (changes requested):
A few more changes - not sure why you went and renamed several of the methods as the original names were clearer and more flexible to avoid future breaking changes. The new methods are more brittle and susceptible to needing to be modified should the scope of this implementation ever change (which it very well might in future runtime releases).

src/Dapr.Workflow.Abstractions/PropagatedHistory.cs
nelson-parente: Done — renamed `_workflows` → `_entries` in da2f380a. Public API unchanged.

src/Dapr.Workflow.Abstractions/HistoryEventKind.cs
nelson-parente: Pushing back here: `Started`/`Completed`/`Failed` mirrors the go-sdk `ActivityResult` struct and python-sdk `ActivityResult` dataclass, which is the whole point of this PR (cross-SDK parity Cassie flagged in #1801). Restoring a .NET-only enum would drop us back out of sync. The booleans are also not mutually exclusive over time — `Started=true, Completed=false, Failed=false` is the pending state — so an enum would actually lose information. Happy to switch if you want parity drift, but want to call that out explicitly.

src/Dapr.Workflow.Abstractions/PropagatedHistory.cs
nelson-parente: Pushing back here too: `GetWorkflows()` matches the go-sdk `PropagatedHistory.GetWorkflows()` and python-sdk equivalents. Renaming to `GetEntries()` puts .NET out of sync with the other SDKs and undoes the parity goal. If/when the runtime extends propagated history beyond workflows, all SDKs would rev together. Keeping the current name.

src/Dapr.Workflow.Abstractions/PropagatedHistory.cs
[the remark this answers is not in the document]
nelson-parente: Already done in 25fe1d75 — all name lookups switched to `OrdinalIgnoreCase`.

src/Dapr.Workflow.Abstractions/PropagatedHistory.cs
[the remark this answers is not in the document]
nelson-parente: Already done in 25fe1d75 — `GetWorkflowsByAppId` and `GetAppIds` dedup use `OrdinalIgnoreCase`.

src/Dapr.Workflow.Abstractions/HistoryEventKind.cs
nelson-parente: Added a `PropagatedHistoryTaskStatus` enum (`Pending`/`Completed`/`Failed`) with a computed `Status` property on both `PropagatedHistoryActivityResult` and `PropagatedHistoryChildWorkflowResult`, so callers can `switch` on a single value instead of reading the three flags — and the enum extends cleanly if the runtime adds lifecycle states later.

I kept the `Started`/`Completed`/`Failed` booleans alongside it: those are the field-level parity with the go-sdk `ActivityResult`/`ChildWorkflowResult` structs and the python-sdk dataclasses, and `Status` is just a projection of them (with `Failed` taking precedence over `Completed`). That keeps the .NET surface in sync with the other SDKs while giving developers the single enum to consume. 5987d9c4

src/Dapr.Workflow.Abstractions/PropagatedHistory.cs
nelson-parente: Renamed `GetWorkflows()` → `GetEntries()` so the name stays scope-neutral if propagated history ever carries non-workflow entries. Applied the same reasoning to the sibling filters (see the `FilterByWorkflowName` thread). 5987d9c4

src/Dapr.Workflow.Abstractions/PropagatedHistory.cs
nelson-parente: Done — `GetWorkflowsByName()` → `FilterByWorkflowName()`, and for consistency `GetWorkflowsByAppId()` → `FilterByAppId()` and `GetWorkflowsByInstanceId()` → `FilterByInstanceId()`. `TryGetLastWorkflowByName()` stays (the TryGet pattern from the earlier round; "ByWorkflowName" lines up with `FilterByWorkflowName`). 5987d9c4

src/Dapr.Workflow.Abstractions/HistoryEventKind.cs
WhitWaldo: The design decisions behind the other SDKs isn't of much concern to me - those maintainers should do what is most appropriate per their own language conventions and express the functionality in ways that make it easier for their developers to consume their SDKs.

I weigh the perceived value of parity with other SDKs with the cost of expanding the public API and having to maintain two equivalent things going forward. I don't think a wall of flags is a developer-friendly way to communicate the state of something, so I appreciate you restoring the enum here. 

Please modify the visibility of the flag fields to make them internal - should there ever be a need to make them publicly accessible, we can always introduce them later, but I'd rather this remain internal and not a part of the public API today.

src/Dapr.Workflow.Abstractions/PropagatedHistoryChildWorkflowResult.cs
WhitWaldo: Flags are more than fine to retain internally, but the point of the SDK Is to simplify development and lower the cognitive burden on developers.

It's a burden on developers to properly map which combination of flags represents what action, especially when we already produce [a list of all possible workflow states](https://github.com/dapr/dotnet-sdk/blob/7f1c553517f16f430d77e1048ed3d564e5c4f563/src/Dapr.Workflow.Abstractions/WorkflowRuntimeStatus.cs#L19) as enums. It doesn't strike me as consistent to have activities reported as a wall of flags and workflows reported as an easy-to-understand enum.

src/Dapr.Workflow.Abstractions/PropagatedHistoryActivityResult.cs
WhitWaldo: Again, don't care about making other SDKs. I care about consistency and ease of use in the .NET SDK itself. Workflow status is reflected by enum. Let's reflect activity status by enum as well.

src/Dapr.Workflow.Abstractions/PropagatedHistory.cs
```diff
@@ -15,71 +15,121 @@ namespace Dapr.Workflow;
...
+        result = null;
+        return false;
+    }
+
+    /// <summary>
+    /// Returns every entry produced by the given app, in execution order.
+    /// </summary>
+    /// <param name="appId">The Dapr App ID to filter by.</param>
+    /// <returns>An empty list when no match is found.</returns>
+    public IReadOnlyList<PropagatedHistoryEntry> FilterByAppId(string appId)
```
cicoyle, line 114: Im very adamant to have consistency across sdks. Can we ensure func names are consistent and not use `filter` in the func name?

WhitWaldo: @cicoyle I'll address this in a separate PR so I can avoid playing any more telephone here :)

marcduiker:
@holopin-bot @nelson-parente sdk-badge Thank you! Here's a digital badge as a small token of appreciation.

PR #1833, opened 2026-05-27
Propagated history changes

WhitWaldo:
# Description

Making a few tweaks + added unit tests to the implementation to support the propagated history implementation.

## Issue reference

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests

### DIFF
--- a/src/Dapr.Workflow.Abstractions/PropagatedHistory.cs
+++ b/src/Dapr.Workflow.Abstractions/PropagatedHistory.cs
@@ -36,20 +36,23 @@ using System.Linq;
 /// </remarks>
 public sealed class PropagatedHistory(IReadOnlyList<PropagatedHistoryEvent> events)
 {
+    private readonly IReadOnlyList<PropagatedHistoryEvent> _events =
+        events ?? throw new ArgumentNullException(nameof(events));
+
     /// <summary>
     /// Returns every event in the propagated history, in execution
     /// order (ancestor first, immediate parent last).
     /// </summary>
-    public IReadOnlyList<PropagatedHistoryEvent> Events => events ?? [];
-    
+    public IReadOnlyList<PropagatedHistoryEvent> Events => _events;
+
     /// <summary>
     /// Returns an ordered, deduplicated list of app IDs in this propagated history.
     /// </summary>
     public IReadOnlyList<string> GetAppIds()
     {
         var seen = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
-        var result = new List<string>(events.Count);
-        result.AddRange(from entry in events where seen.Add(entry.AppId) select entry.AppId);
+        var result = new List<string>(_events.Count);
+        result.AddRange(from entry in _events where seen.Add(entry.AppId) select entry.AppId);
 
         return result;
     }
@@ -63,7 +66,7 @@ public sealed class PropagatedHistory(IReadOnlyList<PropagatedHistoryEvent> even
     public IReadOnlyList<PropagatedHistoryEvent> GetEventsByWorkflowName(string name)
     {
         ArgumentException.ThrowIfNullOrWhiteSpace(name);
-        return events
+        return _events
             .Where(e => string.Equals(e.Name, name, StringComparison.OrdinalIgnoreCase))
             .ToList();
     }
@@ -77,11 +80,11 @@ public sealed class PropagatedHistory(IReadOnlyList<PropagatedHistoryEvent> even
     public bool TryGetLastWorkflowEventByName(string name, [NotNullWhen(true)] out PropagatedHistoryEvent? result)
     {
         ArgumentException.ThrowIfNullOrWhiteSpace(name);
-        for (var i = events.Count - 1; i >= 0; i--)
+        for (var i = _events.Count - 1; i >= 0; i--)
         {
-            if (string.Equals(events[i].Name, name, StringComparison.OrdinalIgnoreCase))
+            if (string.Equals(_events[i].Name, name, StringComparison.OrdinalIgnoreCase))
             {
-                result = events[i];
+                result = _events[i];
                 return true;
             }
         }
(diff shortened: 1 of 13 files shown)


src/Dapr.Workflow.Abstractions/PropagatedHistory.cs
```diff
@@ -90,14 +74,14 @@ public IReadOnlyList<PropagatedHistoryEntry> FilterByWorkflowName(string name)
     /// <param name="name">The workflow name to look up.</param>
     /// <param name="result">When this method returns <see langword="true"/>, the last matching workflow entry; otherwise <see langword="null"/>.</param>
     /// <returns><see langword="true"/> if a matching entry was found; otherwise <see langword="false"/>.</returns>
-    public bool TryGetLastWorkflowByName(string name, [NotNullWhen(true)] out PropagatedHistoryEntry? result)
+    public bool TryGetLastWorkflowEventByName(string name, [NotNullWhen(true)] out PropagatedHistoryEvent? result)
```
cicoyle, line 80: Why `Try`? All the other sdks have `GetLastWorkflowEventByName` without `Try`. Is this more of a dotnet consistency thing?

src/Dapr.Workflow.Abstractions/PropagatedHistoryActivityResult.cs
```diff
@@ -17,34 +17,19 @@ namespace Dapr.Workflow;
...
+/// reflects how far it progressed past scheduling: <see cref="PropagatedHistoryTaskStatus.Pending"/>
+/// means it was scheduled but has not yet completed or failed, <see cref="PropagatedHistoryTaskStatus.Completed"/>
+/// means it succeeded, and <see cref="PropagatedHistoryTaskStatus.Failed"/> means it failed.
 /// </remarks>
 public sealed record PropagatedHistoryActivityResult(
     string Name,
-    bool Started,
-    bool Completed,
-    bool Failed,
+    PropagatedHistoryTaskStatus Status,
```
cicoyle: Users don't need to know we internally call this a `Task`. Could we rename to something like `ActivityStatus`? They care about the status, not the propagated history internals.

src/Dapr.Workflow.Abstractions/PropagatedHistoryWorkflowResult.cs
```diff
@@ -0,0 +1,33 @@
...
+/// <param name="FailureDetails">The failure details when <paramref name="Status"/> is <see cref="PropagatedHistoryTaskStatus.Failed"/>, otherwise <c>null</c>.</param>
+/// <remarks>
+/// Every child workflow surfaced through propagated history was scheduled, so <see cref="Status"/>
+/// reflects how far it progressed past scheduling: <see cref="PropagatedHistoryTaskStatus.Pending"/>
+/// means it was scheduled but has not yet completed or failed, <see cref="PropagatedHistoryTaskStatus.Completed"/>
+/// means it succeeded, and <see cref="PropagatedHistoryTaskStatus.Failed"/> means it failed.
+/// </remarks>
+public sealed record PropagatedHistoryWorkflowResult(
+    string Name,
+    PropagatedHistoryTaskStatus Status,
```
cicoyle: Same as above - could we rename to something like `WorkflowStatus`? `PropagatedHistoryTaskStatus` on a workflow result is a bit confusing too, since `task` usually refers to activities, but end users don't even know that. Users just care about the status, not the propagated history internals.

src/Dapr.Workflow.Abstractions/PropagatedHistoryEvent.cs
```diff
@@ -28,65 +28,62 @@ namespace Dapr.Workflow;
 /// <param name="activities">Activities resolved from this entry, in execution order.</param>
 /// <param name="childWorkflows">Child workflows resolved from this entry, in execution order.</param>
 /// <remarks>
-/// One <see cref="PropagatedHistoryEntry"/> exists per ancestor workflow in a
+/// One <see cref="PropagatedHistoryEvent"/> exists per ancestor workflow in a
 /// <see cref="PropagatedHistory"/>. Use <see cref="TryGetLastActivityByName"/> and
-/// <see cref="TryGetLastChildWorkflowByName"/> to look up specific items in this entry;
```
cicoyle, line 33: Maybe I missed this in our DMs, we do support a top level `GetLastWorkflowByName/GetWorkflowsByName` in the go-sdk. This returns a `WorkflowResult`. In dotnet you have `TryGetLastWorkflowEventByName/GetEventsByWorkflowName` returning `PropagatedHistoryEvent`. 

In go, off of `WorkflowResult.GetLastChildWorkflowByName/GetChildWorkflowsByName` I return the `ChildWorkflowResult`. In dotnet child is dropped. Can we re-add that on `PropagatedHistoryEvent`?

src/Dapr.Workflow.Abstractions/PropagatedHistory.cs
WhitWaldo: A `null` in a response can mean so many things. Taking a recent example from the state store - if you query a value from state and you get a null back, does that mean that the value didn't exist or that it did and it is null? A `TryGet` answers this definitively.

Especially as these are not asynchronous operations, I'd rather return a clear boolean indicating if getting the value(s) worked or not and include a payload with the corresponding result if it did.

src/Dapr.Workflow.Abstractions/PropagatedHistoryActivityResult.cs
WhitWaldo: No objection to just calling it `PropagatedHistoryStatus`, but it doesn't leak internal bits except coincidentally and it's a generic status applicable to either activities or workflows.

src/Dapr.Workflow.Abstractions/PropagatedHistoryWorkflowResult.cs
WhitWaldo: They're the same type as above - just refactored to `PropagatedHistoryStatus` so it's generically used for either one.

src/Dapr.Workflow.Abstractions/PropagatedHistoryEvent.cs
WhitWaldo: I'm not super worried about having a bunch of variations of the filters. I expose all the events in a property, so if there's some special filter a developer wishes to apply, they can trivially add it themselves. 

Also, yes, .NET has filters on the list of all propagated history events to return a single `PropagatedHistoryEvent` as you noted. But this also has filters within it for the specific workflows and activities you cite, so I think the bases are probably already covered here.

PR #1847, opened 2026-06-10
GetWorkflowStateAsync should always return `WorkflowState`

WhitWaldo:
# Description

Ensuring that GetWorkflowStateAsync always returns a `WorkflowState` so the `WorkflowState.Exists` properly reflects the presence of the value (or not). This was erroneously returning null if the metadata wasn't available when this is, in fact, a supported value in the `WorkflowState` object, so I'm applying this fix. The signature doesn't change, just the behavior, so this shouldn't be a dramatic breaking change.

Thank you to @atrauzzi for the helpful discussion in which this was pointed out.

Also, after a very long discussion with @atrauzzi about the merits of logging at the SDK level, he's won me over and I'm also removing the error that was previously logged when using `GetWorkflowStateAsync` when the workflow was returned as not existing. He's right in his assertion that an error was too severe for an operation that completed successfully. The error remains if the runtime returns an unexpected value (null or some other response shape), but will no longer log if it's just a matter of a workflow instance not existing in the runtime.

## Issue reference

## Checklist

* [X] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/src/Dapr.Workflow/Client/WorkflowGrpcClient.cs
+++ b/src/Dapr.Workflow/Client/WorkflowGrpcClient.cs
@@ -73,9 +73,14 @@ internal sealed class WorkflowGrpcClient(
             var grpcCallOptions = CreateCallOptions(cancellationToken);
             var response = await grpcClient.GetInstanceAsync(request, grpcCallOptions);
 
-            if (!response.Exists)
+            if (response is null)
             {
                 logger.LogGetWorkflowMetadataInstanceNotFound(instanceId);
+                return null;
+            }
+
+            if (!response.Exists)
+            {
                 return null;
             }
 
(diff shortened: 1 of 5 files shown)


atrauzzi:
Love it! As always @WhitWaldo, thank you for your vigilance as the keeper of the Dapr dotnet scrolls! :fist_raised:

src/Dapr.Workflow/Client/WorkflowGrpcClient.cs
```diff
@@ -73,17 +73,26 @@ public override async Task<string> ScheduleNewWorkflowAsync(string workflowName,
...
+            {
+                return null;
+            }
+
             return ProtoConverters.ToWorkflowMetadata(response.WorkflowState, serializer);
         }
         catch (RpcException ex) when (ex.StatusCode == StatusCode.NotFound)
         {
-            logger.LogGetWorkflowMetadataInstanceNotFound(ex, instanceId);
+            return null;
```
atrauzzi, line 91: 😍

atrauzzi (approved):

src/Dapr.Workflow/Client/WorkflowGrpcClient.cs
WhitWaldo: @barakbbn If there are other issues like this you're experiencing, please don't hesitate to file an issue about it so I can address them too!

atrauzzi: @barakbbn -- Echoing what Whit said:  We would absolutely love to have you (or anyone reading this) open tickets about any pain points or inconsistencies you think exist in the SDK! :pray:

PR #1862, opened 2026-07-01
Add stateful-history delta work items to the workflow worker

JoshVanL:
The sidecar re-sends a workflow instance's entire committed history to the worker on every turn. This adds the worker half of the "stateful history" optimization so that, once a worker is warm for an instance on a work-item stream, the sidecar sends only the new committed events (the delta) and the worker reconstructs the full history from its own cache. It mirrors the Go (durabletask-go) and Python SDK implementations and is on by default.

Protocol (src/Dapr.Workflow.Grpc/orchestrator_service.proto, compiled by Grpc.Tools at build time):
- Add the CachedHistory message and WorkflowRequest.cachedHistory.
- Add GetWorkItemsRequest.capabilities and WORKER_CAPABILITY_STATEFUL_HISTORY (reserving the never-implemented WORKER_CAPABILITY_HISTORY_STREAMING).

Worker (src/Dapr.Workflow/Worker/Grpc):
- WorkflowHistoryCache: a per-stream cache of each instance's committed history, bounded by a sliding TTL, an instance-count cap, and a byte budget with LRU eviction. Injectable clock for deterministic tests.
- GrpcProtocolHandler: advertise the capability; before each turn resolve the full committed history (cached prefix + delta on a hit, or a GetInstanceHistory fetch on a miss) and normalize request.PastEvents so the handler stays oblivious to the delta protocol; after each turn cache the committed history, or drop it on a CompleteWorkflow action (completed, failed, terminated, or continued-as-new). Reset the cache on reconnect and reclaim idle entries with a PeriodicTimer janitor.

Correctness never depends on the cache: any miss (cold stream, eviction, desync) self-heals via the existing GetInstanceHistory fallback, so this only changes per-turn bandwidth, not results.


WhitWaldo:
@JoshVanL This is intended for 1.19, right? I'll defer merging this until we have an RC to run integration tests against.

JoshVanL:
@WhitWaldo yes exactly, we can keep in draft in the meantime

PR #1881, opened 2026-08-13
build: lift transitive SSH.NET above the GHSA-q939-rpr3-3284 vulnerable range

JoshVanL:
Testcontainers 4.9.0 depends on SSH.NET 2025.1.0, which is inside the vulnerable range of GHSA-q939-rpr3-3284 (high severity, ScpClient recursive download arbitrary file write). NuGet audit promotes that to error NU1903, which currently fails every job in CI, on master and on every open PR, since the advisory published.

Pin SSH.NET 2026.0.0 (the first patched version) centrally and reference it directly from Dapr.Testcontainers so the transitive edge resolves above the vulnerable range. The pin can be dropped once Testcontainers ships a release depending on the patched line.


WhitWaldo (approved):
Looks good - you beat me to publishing this by a day. Nice one!

PR #1900, opened 2026-08-21
Don't send binary grpc-trace-bin header on InvokeBinding calls

### LINKED ISSUE #1897: Publish to servicebusqueue binding throws error invoking output binding <binding name>: (connlost): not a valid UTF-8 string
**Note:**
Not entirely sure if this is a sdk or more a runtime issue.
Tbh. this is probably really specific to the azure servicebus behavior.

## Expected Behavior
With the sdk version 1.18.5 we are facing issues publishing the the **servicebusqueue binding**.
We can see the following error:

```
Error invoking output binding <binding name>: (connlost): not a valid UTF-8 string
```
## Actual Behavior
We are able to publish messages via the **servicebusqueue binding** to the Azure Service Bus again 

## Analysis
The problem is not the message body serialization.
The failure is caused by metadata propagation in the runtime gRPC binding path:

1. In [hgrpc.go#L513](https://github.com/dapr/dapr/blob/c5522cea1f973e47f0d0d4682945e2013585bea7/pkg/api/grpc/grpc.go#L513), InvokeBinding copies incoming gRPC metadata into binding metadata.
2. In [grpc.go#L548](https://github.com/dapr/dapr/blob/c5522cea1f973e47f0d0d4682945e2013585bea7/pkg/api/grpc/grpc.go#L548), binary gRPC headers can be forwarded as string metadata.
3. The key grpc-trace-bin is binary by definition and not guaranteed UTF-8.
4. The Service Bus binding path maps metadata to Service Bus application properties, which require valid UTF-8 strings.
5. Result: Service Bus rejects metadata with not a valid UTF-8 string.

The recent SDK tracing propagation change increased likelihood that grpc-trace-bin reaches the runtime binding path, making this runtime behavior visible in production.

## Release Note

RELEASE NOTE:

DSpirit:
Closes #1897

InvokeBindingAsync was sending the binary grpc-trace-bin header (added in #1858) on every call. The sidecar forwards a call's incoming metadata verbatim into the invoked binding component's metadata, and components like Azure Blob Storage and Service Bus require that metadata to be valid UTF-8 text. Since grpc-trace-bin's bytes come from the random trace/span ID, this caused intermittent failures - Blob request signing broke (403 AuthenticationFailed) and Service Bus rejected the message outright (
ot a valid UTF-8 string).

This strips any binary (-bin suffixed) gRPC metadata from the call options right before invoking InvokeBinding, so it can't be forwarded into binding component metadata regardless of source. 	raceparent/	racestate are still sent since they're plain text and don't cause this issue.

Root cause is really in the runtime (InvokeBinding shouldn't forward binary metadata into component metadata at all) - tracked in dapr/dapr#10387. This is a client-side mitigation so .NET SDK users aren't affected while that's being worked on.

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/src/Dapr.Client/DaprClientGrpc.cs
+++ b/src/Dapr.Client/DaprClientGrpc.cs
@@ -349,6 +349,11 @@ internal class DaprClientGrpc : DaprClient
         }
 
         var options = CreateCallOptions(headers: null, cancellationToken);
+
+        // The sidecar forwards this call's metadata verbatim into the output binding component's metadata
+        // (e.g. Azure Blob/Service Bus properties), which must be valid UTF-8 text, so strip any binary
+        // ("-bin" suffixed) gRPC metadata such as grpc-trace-bin before it can be misrouted there.
+        RemoveBinaryMetadata(options.Headers);
         try
         {
             return await client.InvokeBindingAsync(envelope, options);
@@ -2306,6 +2311,17 @@ internal class DaprClientGrpc : DaprClient
         }
 
         return options;
+    }
+
+    private static void RemoveBinaryMetadata(Metadata headers)
+    {
+        for (var i = headers.Count - 1; i >= 0; i--)
+        {
+            if (headers[i].Key.EndsWith("-bin", StringComparison.OrdinalIgnoreCase))
+            {
+                headers.RemoveAt(i);
+            }
+        }
     }
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


DSpirit:
For reference, the runtime-side fix is now up for review: dapr/dapr#10388 (fixes dapr/dapr#10387 by skipping \-bin\ suffixed metadata keys in InvokeBinding's forwarding loop). This client-side change remains useful as defense-in-depth once that merges - protects users on older runtimes and avoids sending metadata that serves no purpose on this call.

WhitWaldo:
This PR is astonishingly similar to #1901 but the other adds a null check on the headers, so I'm going to take it. Thank you for taking the time to contribute this though!

PR #1901, opened 2026-08-21
Do not send binary gRPC metadata on InvokeBinding calls

### LINKED ISSUE #1897: Publish to servicebusqueue binding throws error invoking output binding <binding name>: (connlost): not a valid UTF-8 string

JoshVanL:
Strip -bin suffixed gRPC metadata from InvokeBinding call options before sending. Trace context still propagates via the traceparent and tracestate text headers, and all other APIs keep sending grpc-trace-bin. This is a client-side mitigation for runtimes without the fix from dapr/dapr#10387.

Fixes #1897

### DIFF
--- a/src/Dapr.Client/DaprClientGrpc.cs
+++ b/src/Dapr.Client/DaprClientGrpc.cs
@@ -349,6 +349,7 @@ internal class DaprClientGrpc : DaprClient
         }
 
         var options = CreateCallOptions(headers: null, cancellationToken);
+        RemoveBinaryMetadata(options.Headers);
         try
         {
             return await client.InvokeBindingAsync(envelope, options);
@@ -2306,6 +2307,28 @@ internal class DaprClientGrpc : DaprClient
         }
 
         return options;
+    }
+
+    /// <summary>
+    /// Removes binary gRPC metadata entries (keys ending in "-bin", such as grpc-trace-bin) because
+    /// the runtime copies InvokeBinding call metadata into string binding component metadata, where
+    /// raw bytes break components that require text values. Trace context still propagates via the
+    /// traceparent and tracestate headers.
+    /// </summary>
+    private static void RemoveBinaryMetadata(Metadata headers)
+    {
+        if (headers is null)
+        {
+            return;
+        }
+
+        for (var i = headers.Count - 1; i >= 0; i--)
+        {
+            if (headers[i].Key.EndsWith("-bin", StringComparison.OrdinalIgnoreCase))
+            {
+                headers.RemoveAt(i);
+            }
+        }
     }
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


WhitWaldo (approved):
Looks good to me - thank you!

src/Dapr.Client/DaprClientGrpc.cs
```diff
@@ -2308,6 +2309,28 @@ private CallOptions CreateCallOptions(Metadata headers, CancellationToken cancel
...
+    private static void RemoveBinaryMetadata(Metadata headers)
+    {
+        if (headers is null)
+        {
+            return;
+        }
+
+        for (var i = headers.Count - 1; i >= 0; i--)
+        {
+            if (headers[i].Key.EndsWith("-bin", StringComparison.OrdinalIgnoreCase))
```
WhitWaldo, line 2327: @JoshVanL Should this look for endswith("bin") instead of "-bin" given the example [here](https://github.com/dapr/dotnet-sdk/issues/1897#issuecomment-5365737458) of "x-ms-meta-daprgrpctracebin"?

JoshVanL: @WhitWaldo no I think keeping the `-bin` is correct:

https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md

```
Note that HTTP2 does not allow arbitrary octet sequences for header values so binary header values must be encoded using Base64 as per https://tools.ietf.org/html/rfc4648#section-4. Implementations MUST accept padded and un-padded values and should emit un-padded values. Applications define binary headers by having their names end with "-bin". Runtime libraries use this suffix to detect binary headers and properly apply base64 encoding & decoding as headers are sent and received.
```

`x-ms-meta-daprgrpctracebin` is not a proper gRPC binary header AFAIK

PR #1903, opened 2026-08-27
Make Actors.Next runtime option defaults nullable and unset

JoshVanL:
The app-wide DaprActorsOptions forced SDK-side defaults for every runtime option and sent them unconditionally on the actor event stream, so the Dapr runtime defaults could never apply. Most notably DrainOngoingCallTimeout was hard-coded to 30 seconds, which sits exactly at the daprd drain clamp boundary (80% of the 30 second dissemination timeout) and above the placement service's 8 second dissemination deadline, causing configured-looking values the app never chose.

Make ActorIdleTimeout, DrainOngoingCallTimeout (and its DrainRebalancedActorsTimeout alias), DrainRebalancedActors, EnableReentrancy, and MaxReentrantDepth nullable with no default. Null means unset: the field is omitted from the wire and the Dapr runtime default applies, matching the classic Dapr.Actors behaviour and the existing per-type DaprActorTypeOptions contract. This also drops the private setter on DrainOngoingCallTimeout, which previously made the alias the only way to set it.

The transport now assigns only configured fields on ActorEntityConfig (all proto fields are optional with explicit presence), and emits the reentrancy config only when reentrancy options are set. The validator skips unset values using the same pattern as the per-type rules. SubscribeActorEventsInitialConfig is widened to nullable accordingly; Actors.Next is not packable so there is no shipped API surface break

### DIFF
--- a/src/Dapr.Actors.Next.Abstractions/Options/DaprActorsOptions.cs
+++ b/src/Dapr.Actors.Next.Abstractions/Options/DaprActorsOptions.cs
@@ -52,18 +52,21 @@ public sealed class DaprActorsOptions
 
     /// <summary>
     /// Gets or sets the idle timeout for actor activations.
+    /// <see langword="null"/> leaves the value unset so the Dapr runtime default applies.
     /// </summary>
-    public TimeSpan ActorIdleTimeout { get; set; } = TimeSpan.FromMinutes(60);
+    public TimeSpan? ActorIdleTimeout { get; set; }
 
     /// <summary>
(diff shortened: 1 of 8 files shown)


WhitWaldo:
@JoshVanL Does this PR take into account the per-actor options changes already in master (but not yet live)?

JoshVanL:
@WhitWaldo  Yep, this PR makes the app-wide side follow the same convention, so the resolution chain is now per-type -> app-wide -> daprd default, with unset fields omitted from the wire entirely.

